PR #18, opened 2017-05-30
added signout functionality

sraspin:


app/controllers/SignOutController.scala
```diff
@@ -0,0 +1,34 @@
...
+import scala.concurrent.Future
+import com.google.inject.{Inject, Singleton}
+
+@Singleton
+class SignOutController @Inject()(implicit val config: AppConfig,
+                                  val messagesApi: MessagesApi
+                                   ) extends FrontendController with I18nSupport {
+
+  val signOut:Action[AnyContent] = Action.async { implicit request =>
+    Future.successful(Ok(views.html.signout.signOut()).withNewSession)
```
MJCallahanPage: For this, we will need to re-direct to the Sign-Out route on the Company Auth Frontend to properly sign them out - rather than just creating a new session. As this would leave an active session in auth.

See the `income-tax-subscription-frontend` here: https://github.com/hmrc/income-tax-subscription-frontend/blob/master/app/controllers/SignOutController.scala

Ours will need to do something similar and add in some application.conf for the sign-out URL 

:smile:

conf/messages
```diff
@@ -7,4 +7,8 @@ base.phase                                                      = BETA
 ## Timeout page ##
 timeout.title                                                   = Your session has timed out
 timeout.heading                                                 = Your session has timed out
-timeout.signIn                                                  = To view your quarterly reporting details, you''ll have to <a id="sign-in-link" href="{0}" rel="external">sign in</a> using your Government Gateway ID.
\ No newline at end of file
+timeout.signIn                                                  = To view your quarterly reporting details, you''ll have to <a id="sign-in-link" href="{0}" rel="external">sign in</a> using your Government Gateway ID.
+
+## Sign Out ##
+signout.title                                                   = Sign Out
+signout.heading                                                 = THIS IS SIGN OUT
```
MJCallahanPage: Is this being used anywhere?

sraspin: It's being used in the signout page

PR #31, opened 2017-06-07
Btat 159

sraspin:


MJCallahanPage:
@sraspin tests look good.

However, I think for our model we should have `LocalDate` type rather than treating them as a string. As it is more useful for us to be able to manipulate it as a date than it is as a string.

See how I've setup and tested the obligations model for dates :smile:

PR #47, opened 2017-06-16
[BTAT-215] refactor service and controller to receive new Calc model

MarkAKelly:


app/services/EstimatedTaxLiabilityService.scala
```diff
@@ -18,25 +18,31 @@ package services
...
-      case error: EstimatedTaxLiabilityError =>
-        Logger.debug(s"[EstimatedTaxLiabilityService][getEstimateTaxLiability] - Error Response Status: ${error.status}, Message: ${error.message}")
+  def getLastEstimatedTaxCalculation(nino: String)(implicit headerCarrier: HeaderCarrier): Future[LastTaxCalculationResponseModel] = {
+    Logger.debug("[EstimatedTaxLiabilityService][getLastEstimatedTaxCalculation] - Requesting Last Tax from Backend via Connector")
+    lastTaxCalculationConnector.getLastEstimatedTax(nino).map {
+      case success: LastTaxCalculation =>
+        Logger.debug(s"[EstimatedTaxLiabilityService][getLastEstimatedTaxCalculation] - Retrieved Estimated Tax Liability: \n\n$success")
+        success match {
+          case LastTaxCalculation(_,_,Some(amount)) => success
+          case LastTaxCalculation(_,_,None) =>
```
MJCallahanPage: This is an interesting one. The call could legitimately be successful but return no last calculation. I.e. if the user has not submitted any data yet then we won't be able to show a calculation.

I think we need to raise this with Oli, Rohima and Laura to see what there views are? As maybe we should handle this as a valid scenario and display different content.

app/connectors/LastTaxCalculationConnector.scala
```diff
@@ -0,0 +1,64 @@
...
+  lazy val getEstimatedTaxLiabilityUrl: String => String = nino => s"$protectedMicroserviceUrl/income-tax-view-change/estimated-tax-liability/$nino"
+
+  def getLastEstimatedTax(nino: String)(implicit headerCarrier: HeaderCarrier): Future[LastTaxCalculationResponseModel] = {
+
+    val url = getEstimatedTaxLiabilityUrl(nino)
+    Logger.debug(s"[LastEstimatedTaxCalculationConnector][getLastEstimatedTax] - GET $url")
+
+    http.GET[HttpResponse](url) flatMap {
+      response =>
+        response.status match {
```
MJCallahanPage: Similar question, what response do we get if there is no latest calculation? As we need to handle that as a valid response. As no calculation might have been done yet.

Again, might be worth speaking to Oli, Rohima and Laura about this,

app/controllers/EstimatedTaxLiabilityController.scala
```diff
@@ -33,16 +33,15 @@ class EstimatedTaxLiabilityController @Inject()(implicit val config: AppConfig,
                                                 implicit val messagesApi: MessagesApi
                               ) extends BaseController {
 
-  val getEstimatedTaxLiability: Action[AnyContent] = authentication.async { implicit request => implicit user =>
+  val c: Action[AnyContent] = authentication.async { implicit request => implicit user =>
```
MJCallahanPage: Typo here, wht is `c` :)

app/connectors/LastTaxCalculationConnector.scala
MJCallahanPage: For now we have agree to render ISE - separate story to handle this scenario gracefully

PR #60, opened 2017-06-26
[BTAT-231] add business tax account url to sidebar

MarkAKelly:


conf/application.conf
```diff
@@ -142,5 +142,9 @@ contact-frontend {
   host = "http://localhost:9250"
 }
 
+business-tax-account {
```
MJCallahanPage: Rather than do this, can you add it as part of the services. Specifying the host and port. Like so:

```
      business-tax-account {
        host = localhost
        port = 9020
      }
```

app/config/frontendAppConfig.scala
```diff
@@ -74,4 +75,7 @@ class FrontendAppConfig @Inject()(configuration: Configuration) extends AppConfi
   //NINO Enrolment
   override lazy val ninoEnrolmentKey: String = loadConfig("enrolments.nino.key")
   override lazy val ninoIdentifierKey: String = loadConfig("enrolments.nino.identifier")
+
+  //Business Tax Account
+  override lazy val businessTaxAccount: String = configuration.getString("business-tax-account.url").getOrElse("")
```
MJCallahanPage: use `baseUrl("business-tax-account")` rather than `configuration.getString...`

MarkAKelly:
Should all be up to date now.  App config changes are coming in as well

PR #68, opened 2017-07-03
Btat 256

sraspin:


app/connectors/CalculationDataConnector.scala
```diff
@@ -0,0 +1,61 @@
...
+              },
+              valid => valid
+            ))
+          case _ =>
+            Logger.warn(s"[CalculationDataConnector][getCalculationData] - RESPONSE status: ${response.status}, body: ${response.body}")
+            Future.successful(CalculationDataErrorModel(response.status, response.body))
+        }
+    } recoverWith {
+      case _ =>
+        Logger.warn(s"[BusinessDetailsConnector][getBusinessList] - Unexpected future failed error when calling $url.")
```
MarkAKelly: This logger message will just need updating to the correct connector name

test/assets/TestConstants.scala
```diff
@@ -200,4 +201,60 @@ object TestConstants extends ImplicitDateFormatter {
     val obligationsDataErrorJson = Json.parse(obligationsDataErrorString)
 
   }
+
+  object CalcBreakdown {
+    val calculationDataSuccessModel = CalculationDataModel(incomeTaxYTD = Some(1),
```
MarkAKelly: Not that were focussing on validating data too much but the YTD figure is actually the final tax total estimate we supply back

PR #67, opened 2017-07-03
[BTAT-251] add IncomeSourceDetailsService and model

MarkAKelly:

PR #69, opened 2017-07-03
[BTAT-252] [BTAT-251] New IncomeSources object in async request predicates

MJCallahanPage:
This PR covers:

- **BTAT-251** : Refactor authentication predicates on controllers. Now chains predicates for:
   > Session Timeout
   > Authentication
   > Income Sources

- **BTAT-252** : Handle new Income Sources implicit from predicates.
   > This predicate calls a IncomeSourceDetails service which orchestrates calls to getBusinessDetails connector and getPropertyDetails connector. It then creates an IncomeSourcesModel which is passed through as an implicit to the controller action so that it is in scope of the controller logic. This will simplify our handling of multiple income sources as we have a clear picture of the user and their income sources straight away from the predicates.


MJCallahanPage:
This PR replaces #67

PR #77, opened 2017-07-05
Btat 297

sraspin:


app/views/estimatedTaxLiability.scala.html
```diff
@@ -17,8 +17,9 @@
 @import views.html.templates.main_template
 @import utils.ImplicitCurrencyFormatter._
 @import views.helpers.TaxYearRenderHelper._
+@import models.CalculationDataResponseModel
 
-@(lastTaxAmount: BigDecimal, taxYear: Int)(implicit request: Request[_], messages: Messages, appConfig: config.AppConfig, user: auth.MtdItUser)
+@(lastTaxAmount: BigDecimal, calcModel:CalculationDataModel, taxYear: Int)(implicit request: Request[_], messages: Messages, appConfig: config.AppConfig, user: auth.MtdItUser)
```
MarkAKelly: This is actually going to accept

`@(calcModel: CalcDisplayModel, taxYear: Int)(implicit request: Request[_], messages: Messages, appConfig: config.AppConfig, user: auth.MtdItUser, sources: models.IncomeSourcesModel)
`
as the last tax amount is in the calcModel, and were going to need the IncomeSources to work out if a user has businesses or property details.

PR #95, opened 2017-07-19
Btat 341

sraspin:


app/controllers/BTAPartialController.scala
```diff
@@ -0,0 +1,48 @@
...
+import services.BTAPartialService
+
+@Singleton
+class BTAPartialController @Inject()(implicit val config: AppConfig,
+                                     implicit val messagesApi: MessagesApi,
+                                     val actionPredicate: AsyncActionPredicate,
+                                     val btaPartialService: BTAPartialService
+                                      ) extends BaseController {
+
+  val setupPartial: Int => Action[AnyContent] = taxYear => actionPredicate.async {
```
MJCallahanPage: The partial needs to be TaxYear agnostic - BTA will not be able to determine this. So this needs to do some logic to work out which TaxYears are relevant based on their IncomeSources

app/views/helpers/BtaPartialHelper.scala
```diff
@@ -0,0 +1,62 @@
...
+  
+  def whichStatus(model: ObligationModel)(implicit messages: Messages) = model.getObligationStatus match {
+    case open: Open =>
+      Html(
+        s"""
+           |<p id="report-due">${messages("bta_partial.next_due", model.due.toLongDate)}</p>
+           |<a id="obligations-link" href=${controllers.routes.ObligationsController.getObligations().url}>${messages("bta_partial.deadlines_link")}</a>
+         """.stripMargin.trim
+      )
+    case overdue: Overdue.type =>
```
MJCallahanPage: Rather than this, you can just say: `case Overdue =>` rather than `case overdue: Overdue.type =>`

Just a little cleaner

app/views/helpers/BtaPartialHelper.scala
```diff
@@ -0,0 +1,62 @@
...
+      Html(
+        s"""
+           |<p id="report-due">${messages("bta_partial.next_due", model.due.toLongDate)}</p>
+           |<a id="obligations-link" href=${controllers.routes.ObligationsController.getObligations().url}>${messages("bta_partial.deadlines_link")}</a>
+         """.stripMargin.trim
+      )
+    case _ =>
```
MJCallahanPage: I assume a sealed trait is returned from getObligationStatus (if it's not then it should be). If it is, then you don't need a `case _ =>` as the compiler will know that the pattern match is exhaustive. So this can be removed :smile:

app/views/helpers/BtaPartialHelper.scala
```diff
@@ -0,0 +1,62 @@
...
+
+  def showLastEstimate(estimate: Option[BigDecimal])(implicit messages: Messages) = estimate match {
+    case Some(est) =>
+      Html(
+        s"""
+           |<p id="current-estimate">${messages("bta_partial.estimated_tax", est.toCurrency)}</p>
+           |<a id="estimates-link" href=${controllers.routes.FinancialDataController.redirectToEarliestEstimatedTaxLiability().url}>${messages("bta_partial.view_details_link")}</a>
+         """.stripMargin.trim
+      )
+    case None =>
```
MJCallahanPage: Can this be simplified to `case None => Html()`?

conf/app.routes
```diff
@@ -20,4 +20,4 @@ GET         /session-timeout                        controllers.timeout.SessionT
 GET         /sign-out                               controllers.SignOutController.signOut
 
 #Sign In Routes
-GET         /sign-in                                controllers.SignInController.signIn
\ No newline at end of file
+GET         /sign-in                                controllers.SignInController.signIn
```
MJCallahanPage: We need a route adding in here that will allow BTA to call and retrieve the partial. This route will need Component Tests.

app/views/helpers/BtaPartialHelper.scala
```diff
@@ -0,0 +1,62 @@
...
+  def whichStatus(model: ObligationModel)(implicit messages: Messages) = model.getObligationStatus match {
+    case open: Open =>
+      Html(
+        s"""
+           |<p id="report-due">${messages("bta_partial.next_due", model.due.toLongDate)}</p>
+           |<a id="obligations-link" href=${controllers.routes.ObligationsController.getObligations().url}>${messages("bta_partial.deadlines_link")}</a>
+         """.stripMargin.trim
+      )
+    case overdue: Overdue.type =>
+      Html(
```
MJCallahanPage: The Overdue output needs say that `Your next report is overdue` the prototype should be updated with this shortly.

app/views/helpers/BtaPartialHelper.scala
```diff
@@ -0,0 +1,62 @@
...
+        s"""
+           |<p id="report-due">${messages("bta_partial.next_due", model.due.toLongDate)}</p>
+           |<a id="obligations-link" href=${controllers.routes.ObligationsController.getObligations().url}>${messages("bta_partial.deadlines_link")}</a>
+         """.stripMargin.trim
+      )
+    case _ =>
+      //TODO something better than this
+  }
+
+  def showLastEstimate(estimate: Option[BigDecimal])(implicit messages: Messages) = estimate match {
```
MJCallahanPage: If there are multiple Tax Years then we need to output both Tax Year calculations

test/services/BTAPartialServiceSpec.scala
```diff
@@ -0,0 +1,170 @@
...
+        await(TestBTAPartialService.getEstimate(testNino, testYear)) shouldBe lastTaxCalcNotFound
+      }
+    }
+    "LastTaxCalculationError is returned from the FinancialDataService" should {
+      "return LastTaxCalcError" in {
+        setupMockGetLastEstimatedTaxCalculation(testNino, testYear)(lastTaxCalcError)
+        await(TestBTAPartialService.getEstimate(testNino, testYear)) shouldBe lastTaxCalcError
+      }
+    }
+    "" should {
```
MarkAKelly: This empty test just needs removing buddy

PR #118, opened 2017-08-11
Btat 430

sraspin:


app/models/IncomeSourcesModel.scala
```diff
@@ -40,16 +42,27 @@ case class IncomeSourcesModel(
   val earliestTaxYear: Option[Int] = orderedTaxYears.headOption
   val lastTaxYear: Option[Int] = orderedTaxYears.lastOption
 
-  val earliestAccountingPeriodStart: Option[LocalDate] = (hasBusinessIncome, hasPropertyIncome) match {
+  val earliestAccountingPeriodStart: Int => Option[LocalDate] = taxYear => (hasBusinessIncome, hasPropertyIncome) match {
     case (b,p) if b && p =>
```
MJCallahanPage: Could we refactor this so that it just uses booleans. E.g, instead of:

`case (b,p) if b && p =>`

Change to:

` case (true, true) =>`

etc/

PR #188, opened 2017-12-06
[BTAT-1197] - Added crystalisation page

sraspin:


app/models/CalcDisplayModel.scala
```diff
@@ -21,7 +21,8 @@ import play.api.libs.json.{Json, OFormat}
 sealed trait CalcDisplayResponseModel
 case class CalcDisplayModel(calcTimestamp: String,
                             calcAmount: BigDecimal,
-                            calcDataModel: Option[CalculationDataModel]) extends CalcDisplayResponseModel {
+                            calcDataModel: Option[CalculationDataModel],
+                            crystalisedFlag: String) extends CalcDisplayResponseModel {
```
MJCallahanPage: rather than `crystalisedFlag` can we call it `calcStatus` and instead of having its type as a `String` can we use either a SealedTrait or an Enumeration for the possible values. That way we don't have to use strings everywhere and we can get some intellisence for the possible values :smile: 

As for the possible values, can we have `Crystalised` and `Estimate`

app/services/CalculationService.scala
```diff
@@ -41,10 +41,10 @@ class FinancialDataService @Inject()(val lastTaxCalculationConnector: LastTaxCal
     } yield (lastCalc, calcBreakdown) match {
       case (calc: LastTaxCalculation, breakdown: CalculationDataModel) =>
         Logger.debug("[FinancialDataService] Retrieved all Financial Data")
-        CalcDisplayModel(calc.calcTimestamp, calc.calcAmount, Some(breakdown))
+        CalcDisplayModel(calc.calcTimestamp, calc.calcAmount, Some(breakdown), calc.crystalisedFlag.getOrElse("Uncrystalised"))
```
MJCallahanPage: Here we'd use the `Crystalised` caseobject of the sealedtrait or enumeration value instead of the string once the model is changed

app/services/CalculationService.scala
```diff
@@ -41,10 +41,10 @@ class FinancialDataService @Inject()(val lastTaxCalculationConnector: LastTaxCal
...
       case (calc: LastTaxCalculation, _) =>
         Logger.debug("[FinancialDataService] Could not retrieve Calculation Breakdown. Returning partial Calc Display Model")
-        CalcDisplayModel(calc.calcTimestamp, calc.calcAmount, None)
+        CalcDisplayModel(calc.calcTimestamp, calc.calcAmount, None, calc.crystalisedFlag.getOrElse("Uncrystalised"))
```
MJCallahanPage: Here we'd use the `Estimate` caseobject of the sealedtrait or enumeration value instead of the string once the model is changed

MJCallahanPage:
@sraspin - just a few small suggestions to use some sort of Type or Enumeration to represent the calcStatus rather than a String.

app/controllers/CalculationController.scala
```diff
@@ -60,8 +61,8 @@ class CalculationController @Inject()(implicit val config: FrontendAppConfig,
             calculationService.getFinancialData(user.nino, taxYear).map {
               case calcDisplayModel: CalcDisplayModel =>
                 submitData(user, sources, calcDisplayModel.calcAmount.toString)
-                if(calcDisplayModel.crystalisedFlag.equals("Crystalised")){
-                  Ok(views.html.crystalised(calcDisplayModel, taxYear))
+                if(calcDisplayModel.crystallisedFlag.equals(Crystallised)){
```
MJCallahanPage: Can this be called `calcStatus` rather than `crystallisedFlag` as this could cater for more than just crystalised calculations in the future. Also, can this be changed to a pattern match for the same reason. I.e.:

```
calcDisplayModel.calcStatus match {
   case calc: Crystallised => Ok(views.html.crystallised(calc, taxYear))
   case calc: Estimate => Ok(view.htmls.estimatedTaxLiability(calc, taxYear))
}
```

app/models/CalcDisplayModel.scala
```diff
@@ -16,13 +16,14 @@
...
 
+import enums.CalcStatus
 import play.api.libs.json.{Json, OFormat}
 
 sealed trait CalcDisplayResponseModel
 case class CalcDisplayModel(calcTimestamp: String,
                             calcAmount: BigDecimal,
                             calcDataModel: Option[CalculationDataModel],
-                            crystalisedFlag: String) extends CalcDisplayResponseModel {
+                            crystallisedFlag: CalcStatus) extends CalcDisplayResponseModel {
```
MJCallahanPage: same here, `calcStatus` instead of `crystallisedFlag`

app/models/LastEstimatedTaxCalculationResponseModel.scala
```diff
@@ -25,7 +26,7 @@ sealed trait LastTaxCalculationResponseModel
 case class LastTaxCalculation(calcID: String,
                               calcTimestamp: String,
                               calcAmount: BigDecimal,
-                              crystalisedFlag: Option[String] = None) extends LastTaxCalculationResponseModel
+                              crystallisedFlag: Option[CalcStatus] = None) extends LastTaxCalculationResponseModel
```
MJCallahanPage: `calcStatus` instead of `crystallisedFlag`

app/services/CalculationService.scala
```diff
@@ -87,4 +88,11 @@ class CalculationService @Inject()(val lastTaxCalculationConnector: LastTaxCalcu
         error
     }
   }
+
+  private[CalculationService] def getStuff(calcStatus: String): CalcStatus = {
```
MJCallahanPage: This doesn't appear to be used. Please can it be removed.

PR #198, opened 2017-12-14
[BTAT-1295/6] Refactored route and controller for home page

MJCallahanPage:


MarkAKelly:
Is this meant to raised against BTAT-828-Story rather than master?

MJCallahanPage:
Will re-raise against the story

PR #221, opened 2018-01-11
[BTAT-1448] edit for single estimate

LadonJackson:


app/controllers/EstimatesController.scala
```diff
@@ -53,6 +53,9 @@ class EstimatesController @Inject()(implicit val config: FrontendAppConfig,
       } yield {
         Logger.debug(s"[EstimatesController][viewEstimateCalculations] Retrieved Last Tax Calcs With Year response: $estimatesResponse")
         if (estimatesResponse.exists(_.isErrored)) itvcErrorHandler.showInternalServerError
+        else if(estimatesResponse.count(_.matchesStatus(Estimate)) == 1) {
```
MJCallahanPage: Small change needed here. Instead of counting all the estimates can you count all the ones that are not crystallised. Otherwise, the routing will be incorrect in certain circumstances where no calculation has been performed yet. An example of this is with test case AA888888A.

`_.matchesStatus(Estimate)) == 1` change to `!_.matchesStatus(Crystallised)) == 1`

This will probably impact the tests.

MJCallahanPage:
Waiting on Acceptance Test changes before merging.

PR #230, opened 2018-01-16
Btat 1495

LadonJackson:


MJCallahanPage:
@LadonJackson - could you add some unit tests in the EstimatesViewSpec to check the link and the url

PR #264, opened 2018-02-15
[BTAT-1793] updated calculationDataModel field name

LadonJackson:


MJCallahanPage:
As discussed, closing so that the PR can be re-raised against a BTAT-1646-Story branch instead of master.

PR #272, opened 2018-02-20
 [BTAT-1813] Added Payment controller, with paymentHandoff method

LadonJackson:


app/controllers/PaymentController.scala
```diff
@@ -0,0 +1,47 @@
...
+
+@Singleton
+class PaymentController @Inject()(implicit val config: FrontendAppConfig,
+                                  implicit val messagesApi: MessagesApi,
+                                  val checkSessionTimeout: SessionTimeoutPredicate,
+                                  val authenticate: AuthenticationPredicate) extends BaseController {
+
+  val action = checkSessionTimeout andThen authenticate
+
+  val paymentHandoff:Int => Action[AnyContent] = paymentAmount => action.async {
```
MJCallahanPage: Can you change `paymentAmount` to `paymentAmountInPence` so that it's clear that it's pennies

app/controllers/PaymentController.scala
```diff
@@ -0,0 +1,47 @@
...
+    implicit user =>
+      Future.successful(Redirect(config.paymentsUrl).addingToSession(
+        "payment-data" -> payment(user.mtditid, paymentAmount).toString())
+      )
+  }
+
+  private def payment(mtditid: String, paymentAmount: Int) = Json.toJson[PaymentDataModel](
```
MJCallahanPage: same here. `paymentAmount` to `paymentAmountInPence`

app/models/PaymentDataModel.scala
```diff
@@ -0,0 +1,32 @@
...
+
+import play.api.libs.json.{Json, OFormat}
+
+sealed trait PaymentData
+
+case class PaymentDataModel(taxType: String, taxReference: String, amountInPence: Int, returnUrl: String)
+case class PaymentErrorModel(code: Int, message: String) extends PaymentData
+
+object PaymentDataModel {
+  implicit val format: OFormat[PaymentDataModel] = Json.format[PaymentDataModel]
```
MJCallahanPage: Rather than `OFormat` can you change to `Format`

app/models/PaymentDataModel.scala
```diff
@@ -0,0 +1,32 @@
...
+}
+
+object PaymentErrorModel {
+  implicit val format: OFormat[PaymentErrorModel] =Json.format[PaymentErrorModel]
```
MJCallahanPage: Same here: `OFormat` can you change to `Format`

app/utils/ImplicitCurrencyFormatter.scala
```diff
@@ -26,6 +26,7 @@ trait ImplicitCurrencyFormatter {
     def toCurrency: Html = Html("&pound;" + f.format(x).replace(".00",""))
     def toCurrencyString: String = "£" + f.format(x).replace(".00","")
     def toStringNoDecimal: String = x.toString.replace(".0","")
+    def toPence: Int = x.toInt * 100
```
MJCallahanPage: This won't work as expected. You can only do the `.toInt` after the multiplication, otherwise you will lose the decimal places. See below:

```
scala> val x = BigDecimal(12.22)
x: scala.math.BigDecimal = 12.22

scala> x.toInt * 100
res0: Int = 1200

scala> (x*100).toInt
res1: Int = 1222
```

test/views/CrystallisedViewSpec.scala
```diff
@@ -307,6 +307,20 @@ class CrystallisedViewSpec extends TestSupport {
...
     }
 
+    "NOT show a button to go to payments, when the payment feature is disabled" in {
+      mockAppConfig.features.paymentEnabled(false)
+      val setup = pageSetup(busPropBRTCalcDataModel, testIncomeSources)
+      import setup._
+      document.getElementById("payment-button") should be(null)
+    }
+
+    "show a button to go to payments, when the payment feature is enabled" in {
```
MJCallahanPage: Can you add some tests to prove that:
- The wording on the button is correct
- The link on the button goes to the correct location

app/config/features/Keys.scala
```diff
@@ -18,4 +18,5 @@ package config.features
 
 object Keys {
   val homePageEnabled = "features.homePageEnabled"
+  val paymentEnabled = "features.paymentEnabled"
```
MJCallahanPage: This needs to be added to the `conf.application.conf` as well.

PR #273, opened 2018-02-20
[BTAT-1790] updated content for income tax estimates where there are …

LadonJackson:
…no relevent tax years

### DIFF
--- a/conf/messages
+++ b/conf/messages
@@ -47,7 +47,8 @@ estimates.title                                                 = Current estima
 estimates.heading                                               = View your estimates
 estimates.viewEstimates                                         = View your current estimates.
 estimates.estimatesLink                                         = Tax year: {0} to {1}
-estimates.noEstimates                                           = You''ve not got any current estimates.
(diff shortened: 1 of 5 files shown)


conf/messages
```diff
@@ -77,7 +77,7 @@ estimated_tax_liability.calc-breakdown.reliefs                  = Your tax relie
 
 estimated_tax_liability.calc-breakdown.total-estimate           = Your Income Tax and National Insurance estimate
 estimated_tax_liability.estimated-date                          = Estimate up to your <span id="calc-date">{0}</span> submission
-estimated_tax_liability.no-data                                 = Once you''ve submitted a report using your accounting software, you can view your tax estimate here.
+estimated_tax_liability.no-data                                 = You don''t have an estimate right now. We''ll show your next Income Tax estimate when you submit a report using software.
```
MJCallahanPage: @LadonJackson - I don't think this is the right message. It should be the one on the `estimates` page not the `estimated_tax_liability` page. We should probably refactor `estimated_tax_liability` to be `calculation` so that it's clearer.

PR #283, opened 2018-03-02
[BTAT-1976] Updated the Income Tax Partial Content for BTA

MJCallahanPage:
- Refactored the tests, view and controller
- Teared down unused classes
- Created a test-only route so that the partial can be easily demoed without BTA integration


app/views/btaPartial.scala.html
```diff
@@ -14,19 +14,18 @@
...
-
-@import views.helpers.BtaPartialHelper
 @import play.twirl.api.Html
 
-@(obMod: ReportDeadlinesResponseModel, estimates: List[LastTaxCalculationWithYear])(implicit request: Request[_], messages: Messages, appConfig: config.FrontendAppConfig, user: auth.MtdItUser[_])
+@()(implicit request: Request[_], messages: Messages, appConfig: config.FrontendAppConfig)
+
+<section id="section-it-qr" class="section divider--bottom">
+
+    <h2 id="it-quarterly-reporting-heading">@messages("bta_partial.heading")</h2>
```
sraspin: Do we need to have the 'New' symbol here?

MJCallahanPage: No we don't; there's nothing in assets frontend to support the styling. So the decision has been made to leave it out (Jordan confirmed with Craig yesterday. The prototype should be updated to reflect this).

PR #285, opened 2018-03-02
[BTAT-1949] Added Feature Switch for Statements view

LadonJackson:


app/controllers/StatementsController.scala
```diff
@@ -38,11 +38,15 @@ class StatementsController @Inject()(implicit val config: FrontendAppConfig,
     implicit user =>
       for {
         financialTransactionsResponse <- financialTransactionsService.getFinancialTransactions(user.mtditid)
-      } yield financialTransactionsResponse match {
-        case model: FinancialTransactionsModel =>
+        statementsEnabled = config.features.statementsEnabled()
```
MJCallahanPage: Instead of this, move the previous logic to a private method called `renderView`

Have another private method called `redirectToHome`

Then change the public method to:

`if (config.features.statementsEnabled()) renderView else redirectToHome`

This has two benefits:
  - better readability, maintainability and separation of concerns
  - means the financialTransactionsService is only called when it needs to be called

MJCallahanPage:
@LadonJackson couple of changes please, otherwise looks good.

MJCallahanPage:
@LadonJackson can you resolve the conflicts please

PR #288, opened 2018-03-05
[BTAT-1952] added Bills feature switch

LadonJackson:


app/controllers/BillsController.scala
```diff
@@ -41,15 +44,22 @@ class BillsController @Inject()(implicit val config: FrontendAppConfig,
...
-        lastTaxCalcs <- calculationService.getAllLatestCalculations(user.nino, sources.orderedTaxYears)
-      } yield {
-          Logger.debug(s"[BillsController][viewCrystallisedCalculations] Retrieved Last Tax Calcs With Year response: $lastTaxCalcs")
-          if (lastTaxCalcs.exists(_.isErrored)) itvcErrorHandler.showInternalServerError
-          else Ok(views.html.bills(lastTaxCalcs.filter(_.matchesStatus(Crystallised))))
-      }
+    implicit user => if(config.features.billsEnabled()) renderView else redirectToHome
+  }
+
+  private[BillsController] def redirectToHome: Future[Result] = {
```
MJCallahanPage: We could refactor this to a common method that is on the `BaseController` trait. As there will be a number of controllers which need to redirect to home. There's an example on one of my PRs.

app/controllers/BillsController.scala
```diff
@@ -41,15 +44,22 @@ class BillsController @Inject()(implicit val config: FrontendAppConfig,
...
+    Future.successful(Redirect(controllers.routes.HomeController.home().url))
+  }
+
+  private[BillsController] def renderView[A](implicit user: MtdItUser[A]): Future[Result] = {
+    implicit val sources: IncomeSourcesModel = user.incomeSources
+
+    for {
```
MJCallahanPage: As there is only one future within this `for` comprehension it feels like overkill to use this.

Instead, you could just use `.map` on the call to the service. I.e.:
```
    calculationService.getAllLatestCalculations(user.nino, sources.orderedTaxYears).map {
      case lastTaxCalcs if lastTaxCalcs.exists(_.isErrored) =>
        Logger.debug(s"[BillsController][viewCrystallisedCalculations] Retrieved at least one Errored Last Tax Calc. Response: $lastTaxCalcs")
        itvcErrorHandler.showInternalServerError
      case lastTaxCalcs =>
        Ok(views.html.bills(lastTaxCalcs.filter(_.matchesStatus(Crystallised))))
    }
```

MJCallahanPage:
@LadonJackson as expected, there are some conflicts now with the base branch. Can you rebase and resolve them.

app/controllers/BillsController.scala
```diff
@@ -44,22 +44,20 @@ class BillsController @Inject()(implicit val config: FrontendAppConfig,
...
 
   val viewCrystallisedCalculations: Action[AnyContent] = (checkSessionTimeout andThen authenticate andThen retrieveNino andThen retrieveIncomeSources).async {
-    implicit user => if(config.features.billsEnabled()) renderView else redirectToHome
-  }
-
-  private[BillsController] def redirectToHome: Future[Result] = {
-    Future.successful(Redirect(controllers.routes.HomeController.home().url))
+    implicit user => if(config.features.billsEnabled()) renderView else fRedirectToHome
   }
+//
```
MJCallahanPage: Can you tidy up these commented out lines

PR #307, opened 2018-03-16
[BTAT-2115] removed homepage feature switch

LadonJackson:
Assigned to @sraspin


app/controllers/HomeController.scala
```diff
@@ -35,9 +35,7 @@ class HomeController @Inject()(val checkSessionTimeout: SessionTimeoutPredicate,
                                implicit val config: FrontendAppConfig,
                                val messagesApi: MessagesApi) extends FrontendController with I18nSupport {
 
-  val home: Action[AnyContent] = (checkSessionTimeout andThen authenticate andThen retrieveNino).async { implicit user =>
-    if (config.features.homePageEnabled()) renderView else redirectToBTA
-  }
+  val home: Action[AnyContent] = (checkSessionTimeout andThen authenticate andThen retrieveNino).async { implicit user => renderView}
```
sraspin: We can move the 'renderView' function into 'home' val and get rid of 'redirectToBTA'

it/controllers/HomeControllerISpec.scala
```diff
@@ -30,54 +30,26 @@ class HomeControllerISpec extends ComponentSpecBase with GenericStubMethods with
...
-          res should have(
-
-            //Check Status OK (200) Result
-            httpStatus(OK),
-
-            //Check Redirect Location
-            pageTitle("Your Income Tax")
-          )
-        }
-      }
```
sraspin: Keep this test in - just without the appConfig.enabled bit

LadonJackson: I think it's just not being shown properly for some reason, if you go onto the file I think it should be okay.

PR #316, opened 2018-03-21
[BTAT-2120] - refactored IncomeSourcesModel to be 'WithDeadlines'

sraspin:


MJCallahanPage:
@sraspin as discussed, a few little tweaks to the BusinessDetailsController to use the `user.incomeSources` instead of calling the connector again to retrieve the details it already has access to.

If you're happy with these additional changes we'll merge it in

sraspin:
Changes all look good to me - do it!

PR #391, opened 2018-09-03
BTAT-3588: Added new bills view and model

ja-green:


test/assets/FinancialTransactionsTestConstants.scala
```diff
@@ -187,4 +188,15 @@ object FinancialTransactionsTestConstants extends ImplicitDateFormatter {
     financialTransactions = Some(Seq(transactionModel(taxYear)))
   )
 
+  val paidBillsViewModel: BillsViewModel = BillsViewModel(
```
lewishayqa: Could these view models be put somewhere else? This constants object is related to the Financial Transactions API.

test/views/api19a/BillViewSpec.scala
```diff
@@ -0,0 +1,104 @@
...
+import play.api.i18n.Messages.Implicits._
+import play.api.test.FakeRequest
+import play.api.test.Helpers._
+import play.twirl.api.HtmlFormat
+import utils.ImplicitCurrencyFormatter._
+import utils.TestSupport
+
+class BillViewSpec extends TestSupport {
+
+  val mockAppConfig: FrontendAppConfig = app.injector.instanceOf[FrontendAppConfig]
```
lewishayqa: There is already an instance of `FrontendAppConfig` made available from `TestSupport`. You could reuse that one instead of defining another one here. Could you also update `EstimatesViewSpec` to do the same please as I missed this in the earlier PR.

PR #441, opened 2019-04-03
Updated model so isError checks the status of the error model,

AlexRimmerHMRC:
Updated checks on estimates controller to redirect if only one calculation is an uncrystallised calc model,
Removed some duplicate logging and unreachable code,
Updated and added tests

### DIFF
--- a/app/controllers/EstimatesController.scala
+++ b/app/controllers/EstimatesController.scala
@@ -19,12 +19,9 @@ package controllers
 import javax.inject.Inject
 
 import audit.AuditingService
-import auth.{MtdItUser, MtdItUserOptionNino, MtdItUserWithNino}
+import auth.MtdItUser
 import config.{FrontendAppConfig, ItvcErrorHandler, ItvcHeaderCarrierForPartialsConverter}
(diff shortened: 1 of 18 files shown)


app/controllers/EstimatesController.scala
```diff
@@ -48,15 +48,17 @@ class EstimatesController @Inject()(implicit val config: FrontendAppConfig,
...
-      case estimatesResponse if estimatesResponse.exists(_.isError) =>
-        Logger.error(s"[EstimatesController][viewEstimateCalculations] Retrieved at least one Errored Last Tax Calc. Response: $estimatesResponse")
-        itvcErrorHandler.showInternalServerError
-      case estimatesResponse if estimatesResponse.count(!_.isCrystallised) == 1 =>
-        Redirect(controllers.routes.CalculationController.renderCalculationPage(estimatesResponse.filter(!_.isCrystallised).head.year))
-          .addingToSession("singleEstimate" -> "true")
-      case estimatesResponse => Ok(views.html.estimates(estimatesResponse.filter(!_.isCrystallised)))
-        .addingToSession("singleEstimate" -> "false")
+    calculationService.getAllLatestCalculations(user.nino, user.incomeSources.orderedTaxYears).map { response =>
+      println(response)
```
jameshforster: Please remove

PR #488, opened 2020-02-15
 MISUV-258:changed FS function to be consist to ITSA FE

scullythornton1:


app/config/featureswitch/FeatureSwitch.scala
```diff
@@ -0,0 +1,112 @@
...
+    switches find (_.name == str) match {
+      case Some(switch) => switch
+      case None => throw new IllegalArgumentException("Invalid feature switch: " + str)
+    }
+
+  def get(str: String): Option[FeatureSwitch] = switches find (_.name == str)
+
+}
+
+case object PropertyDetails extends FeatureSwitch {
```
AlexRimmerHMRC: As part of the story it mentions "When moving over the feature switch do not convert any dead code over."
This feature switch is not used anywhere in the code, so should be removed.

app/config/featureswitch/FeatureSwitch.scala
```diff
@@ -0,0 +1,112 @@
...
+  override val name = s"$prefix.enable-property-details"
+  override val displayText = "Enable Property Details Connector Response"
+}
+
+case object PropertyEops extends FeatureSwitch {
```
AlexRimmerHMRC: As part of the story it mentions "When moving over the feature switch do not convert any dead code over."
This feature switch is not used anywhere in the code, so should be removed.

app/config/featureswitch/FeatureSwitch.scala
```diff
@@ -0,0 +1,112 @@
...
+  override val name = s"$prefix.enable-property-eops"
+  override val displayText = "Enable Property EOPS Deadlines"
+}
+
+case object BusinessEops extends FeatureSwitch {
```
AlexRimmerHMRC: As part of the story it mentions "When moving over the feature switch do not convert any dead code over."
This feature switch is not used anywhere in the code, so should be removed.

PR #503, opened 2020-03-31
MISUV-377 - Created view for tax year overview, Updated controller to use the new view

JSimmons072019:


app/models/calculation/CalcOverview.scala
```diff
@@ -0,0 +1,41 @@
...
+object CalcOverview {
+
+  def apply(calculation: Calculation, transaction: Option[TransactionModel]): CalcOverview = CalcOverview(
+    timestamp = calculation.timestamp,
+    income = calculation.totalIncomeReceived.getOrElse(0.00),
+    deductions = calculation.allowancesAndDeductions.totalAllowancesDeductionsReliefs.getOrElse(0.00),
+    totalTaxableIncome = calculation.totalTaxableIncome.getOrElse(0.00),
+    taxDue = calculation.totalIncomeTaxAndNicsDue.getOrElse(0.00),
+    payment = transaction.flatMap(_.clearedAmount).getOrElse(0.00),
+    totalRemainingDue = transaction.flatMap(_.outstandingAmount).getOrElse(0.00)
```
jameshforster: This should not default to 0, if the transaction is None then it should be equal to tax due.

PR #505, opened 2020-04-02
Misuv 387: created income link connect tax year overview to income summary page

scullythornton1:


app/views/taxYearOverview.scala.html
```diff
@@ -64,7 +64,8 @@ <h1 class="heading-xlarge">
 
             <table id="income-deductions-table">
                 <tr>
-                    <td>@messages("tax-year-overview.income")</td>
+                    <td id="income-row-title"><a href=@controllers.routes.IncomeSummaryController.showIncomeSummary(taxYear).url>
```
jameshforster, line 67: The link needs to be conditional based on the feature switch, as it stands if the feature switch is disabled the link still appears and routes to a Not Found page which is a poor user journey.

jameshforster, line 67: Please do not add ids unless they are needed to for html/css functionality or on a dynamic table. As this table always has a set number of rows it is not required.

test/views/TaxYearOverviewViewSpec.scala
```diff
@@ -88,11 +91,18 @@ class TaxYearOverviewViewSpec extends TestSupport {
     }
 
     "have a table of income and deductions" which {
-      "has a row for income" in new Setup {
+        "has a row for income when FS IncomeBreakdown is disabled " in new Setup {
```
jameshforster, line 94: Test should be updated to include a check to determine that a link for income does not appear in this situation.

PR #506, opened 2020-04-06
MISUV-330 Fix missing ids and link accessibility.

jameshforster:


JSimmons072019:
There is one failing unit test in IncomeBreakdownViewSpec

[info]     - has a total line with the correct value *** FAILED ***
[info]       "£[]0" was not equal to "£[10,010.1]0" (IncomeBreakdownViewSpec.scala:132)

PR #519, opened 2020-04-23
FS changes for Tax due and adding missing config

chandraAndem:


app/config/featureswitch/FeatureSwitch.scala
```diff
@@ -80,3 +80,8 @@ case object DeductionBreakdown extends FeatureSwitch {
   override val name: String = s"$prefix.enable-calculation-deduction-breakdown"
   override val displayText: String = "Enable Deduction Calc Breakdown"
 }
+
+case object TaxDue extends FeatureSwitch {
```
scullythornton1, line 84: Please also add this FS into line 37 after the DeductionBreakdown, otherwise, the new FS will not show in the test-only feature-switches list

val switches: Set[FeatureSwitch] = Set(
    Payment,
    Estimates,
    Bills,
    ReportDeadlines,
    ObligationsPage,
    IncomeBreakdown,
    DeductionBreakdown
  )

conf/application.conf
```diff
@@ -133,6 +133,9 @@ feature-switch {
   enable-calc-breakdown = false
   enable-calc-dataApi = true
   enable-obligations-page = false
```
scullythornton1, line 135: can you please also delete the below FS which are not in use anymore but we have never updated.

  enable-account-details = true
  enable-calc-breakdown = false
  enable-calc-dataApi = true

app/config/featureswitch/FeatureSwitch.scala
chandraAndem: Done

conf/application.conf
chandraAndem: Done

PR #520, opened 2020-04-28
Adding TaxDeductedAtSource model

chandraAndem:


app/models/calculation/Calculation.scala
```diff
@@ -115,6 +115,11 @@ case class NicBand(income: BigDecimal,
 
 object NicBand {
   implicit val format: OFormat[NicBand] = Json.format[NicBand]
-
 }
 
+case class TaxDeductedAtSource(savings: Option[BigDecimal],
+                               ukLandAndProperty: Option[BigDecimal])
```
AlexRimmerHMRC, line 121: we won't be using ukLandAndProperty as it's not in the prototype as far as I'm aware.
The case class needs to be a part of the Calculation model too, with the correct reads

chandraAndem: Done

PR #523, opened 2020-04-30
Tax due controller and FS changes

chandraAndem:


app/views/taxYearOverview.scala.html
```diff
@@ -22,7 +22,7 @@
 @import views.html.helpers.{breadcrumbHelper, statusHelper}
 @import views.html.templates.main_template
 
-@(taxYear: Int, overview: CalcOverview, transaction: Option[TransactionModel] = None, incomeBreakdown: Boolean = false, deductionBreakdown: Boolean = false)(implicit request: Request[_], messages: Messages, appConfig: config.FrontendAppConfig)
+@(taxYear: Int, overview: CalcOverview, transaction: Option[TransactionModel] = None, incomeBreakdown: Boolean = false, deductionBreakdown: Boolean = false, taxDue: Boolean = false)(implicit request: Request[_], messages: Messages, appConfig: config.FrontendAppConfig)
```
AlexRimmerHMRC, line 25: taxDue needs adding to the if statement on line 64 as another or case

app/controllers/TaxDueSummaryController.scala
```diff
@@ -0,0 +1,75 @@
...
+                                     ) extends BaseController with ImplicitDateFormatter with FeatureSwitching {
+
+  val action: ActionBuilder[MtdItUser] = checkSessionTimeout andThen authenticate andThen retrieveNino andThen retrieveIncomeSources
+
+
+  def showTaxDueSummary(taxYear: Int): Action[AnyContent] = {
+
+    action.async {
+      implicit user =>
+          if (isEnabled(TaxDue)) {
```
AlexRimmerHMRC, line 57: Formatting looks off, can you reformat the file?

app/controllers/TaxDueSummaryController.scala
```diff
@@ -0,0 +1,75 @@
...
+                                        val checkSessionTimeout: SessionTimeoutPredicate,
+                                        val authenticate: AuthenticationPredicate,
+                                        val retrieveNino: NinoPredicate,
+                                        val retrieveIncomeSources: IncomeSourceDetailsPredicate,
+                                        val calculationService: CalculationService,
+                                        val itvcHeaderCarrierForPartialsConverter: ItvcHeaderCarrierForPartialsConverter,
+                                        val auditingService: AuditingService,
+                                        val financialTransactionsService: FinancialTransactionsService,
+                                        val itvcErrorHandler: ItvcErrorHandler
+                                     ) extends BaseController with ImplicitDateFormatter with FeatureSwitching {
```
AlexRimmerHMRC, line 48: The way all of these are injected means that they are all implicit. Reference CalculationController on solving this.
Some of these that are injected don't seem to be used also. Can you check whether some of these are required?
For example, financialTransactionsService is not being used.

app/views/taxYearOverview.scala.html
chandraAndem: Added

app/controllers/TaxDueSummaryController.scala
chandraAndem: Made suggested changes.

app/controllers/TaxDueSummaryController.scala
chandraAndem: Reformatted whole file now

PR #533, opened 2020-06-09
Connector and Model changes for PaymentAllocation API

chandraAndem:


test/models/PaymentAllocationsResponseModelSpec.scala
```diff
@@ -0,0 +1,129 @@
...
+      }
+
+      "have the amount as '10.10'" in {
+        allocation.amount.get shouldBe 10.90
+      }
+
+      "have the clearedAmount as '5.50'" in {
+        allocation.clearedAmount.get shouldBe 5.90
+      }
+    }
```
AlexRimmerHMRC, line 97: I don't think we should be testing this, at this point you are just testing the functionality of scala

chandraAndem: Yeah, even I was thinking same but scoverage plugin reports this as not covered in coverage if not tested.
I just noticed that we ignore models package from scoverage, so removed these tests now :)

PR #538, opened 2020-08-27
MISUV-792: fix tax year overview page H1 and title pattern accessibility issue…

scullythornton1:
…ity issue

### DIFF
--- a/conf/messages.cy
+++ b/conf/messages.cy
@@ -238,3 +238,6 @@ button.continue                                                 = Yn eich blaen
 ## Not found
 notFound.heading = Ni ellir dod o hyd i’r dudalen hon
 notFound.message = Gwiriwch eich bod wedi nodi’r cyfeiriad gwe cywir.
+
+## Title pattern
+titlePattern.serviceName.govUk = {0} - Cyfrif Treth Busnes - GOV.UK
--- a/app/views/templates/govuk_wrapper.scala.html
+++ b/app/views/templates/govuk_wrapper.scala.html
@@ -103,4 +103,4 @@
       getHelpForm = getHelpForm)
 }
 
-@hmrcGovUkTemplate(Some(title), bodyClasses)(head, bodyEnd, insideHeader, afterHeader, footerTop, Some(uiLayouts.footer_links()), nav=true)(content)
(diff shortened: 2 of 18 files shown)


app/views/templates/main_template.scala.html
```diff
@@ -40,7 +40,7 @@
                 timeout: 900,
                 countdown: 120,
                 keep_alive_url: "@controllers.timeout.routes.SessionTimeoutController.keepAlive().url",
-                title: "@messages("timeoutDialog.title")",
+                title: "@messages("timeoutDialog.title") + " " + @messages("titlePattern.serviceName.govUk")",
```
AlexRimmerHMRC, line 43: I don't think this is necessary, I think the "title" for the timeout is actually the heading of the popup that appears to users, in which case, we likely do not want the service name and gov.uk appended

app/views/templates/main_template.scala.html
```diff
@@ -71,7 +71,7 @@
     @mainContent
 }
 
-@govuk_wrapper(title = title,
+@govuk_wrapper(title = title + " " + messages("titlePattern.serviceName.govUk"),
```
AlexRimmerHMRC, line 74: Not too much of a fan of this, might be cleaner to update the message to be
`{0} - Business Tax account - GOV.UK` that way here you could just do 
`title = messages("titlePattern.serviceName.govUk", title)`
without needing to use `+ " " +`

conf/messages.cy
```diff
@@ -237,4 +237,7 @@ button.continue                                                 = Yn eich blaen
 
 ## Not found
 notFound.heading = Ni ellir dod o hyd i’r dudalen hon
-notFound.message = Gwiriwch eich bod wedi nodi’r cyfeiriad gwe cywir.
\ No newline at end of file
+notFound.message = Gwiriwch eich bod wedi nodi’r cyfeiriad gwe cywir.
+
+## Title pattern
+titlePattern.serviceName.govUk = - Cyfrif Treth Busnes - GOV.UK
```
AlexRimmerHMRC, line 243: can you add a new line underneath this

app/views/templates/main_template.scala.html
scullythornton1: done

app/views/templates/main_template.scala.html
scullythornton1: done

PR #549, opened 2020-10-05
MISUV-971: fix the accessibility issue with the toggle menu.

scullythornton1:


public/stylesheets/itvc.css
```diff
@@ -326,8 +326,11 @@ html[data-useragent*='MSIE 10.0'] .flex-container {
...
 
-
-
-
+#global-header .header-proposition a.menu {
+    display: none;
+}
+.js-enabled #global-header .header-proposition #proposition-links {
+    display: block;
+}
```
AlexRimmerHMRC, line 334: Could you add a comment into the CSS to explain what this is for?
Will help in the future if we can ever remove our custom css to know what it was for

scullythornton1: fixed.

PR #552, opened 2020-10-06
MISUV-970. Added contrast.css and implemented changes

WozCapGem:


public/stylesheets/itvc.css
```diff
@@ -334,4 +334,78 @@ html[data-useragent*='MSIE 10.0'] .flex-container {
...
 
+@media (min-width: 40.0625em)
+.govuk-details {
+margin-bottom: 30px;
+}
+
+@media (min-width: 40.0625em)
+.govuk-details {
+font-size: 19px;
+font-size: 1.1875rem;
```
AlexRimmerHMRC, line 347: Can you remove these two lines, will cause the issue we saw with the text size when on smaller screens

public/stylesheets/itvc.css
```diff
@@ -334,4 +334,78 @@ html[data-useragent*='MSIE 10.0'] .flex-container {
     display: block;
 }
 
+/* fix styling for details */
```
AlexRimmerHMRC, line 337: Could you format the css below that was added

PR #572, opened 2020-12-15
MISUV-1255: Viewer content updates for December 2020

AIbrahimQA:


app/views/taxCalcBreakdown.scala.html
```diff
@@ -68,12 +68,12 @@ <h1 class="heading-xlarge">
         @messages("taxCal_breakdown.heading")
     </h1>
 
-    <h2 class="heading-large">
+    <div id="explanation" style="font-size:24px">
```
chandraAndem, line 71: Could we remove style tag and use class eg: **govuk-heading-m** ?

chandraAndem (changes requested):
Minor change required.

chandraAndem (approved):

PR #585, opened 2021-01-26
MISUV-1365: Viewer - Agent Auth Predicate

AIbrahimQA:


chandraAndem (changes requested):
Missing Authentication controllers and few formatting comments added.

PR #591, opened 2021-02-01
MISUV-1438: create call to getBusinessDetails using NINO

scullythornton1:


app/connectors/IncomeTaxViewChangeConnector.scala
```diff
@@ -46,6 +46,10 @@ trait IncomeTaxViewChangeConnector extends RawResponseReads {
   val config: FrontendAppConfig
   implicit val ec: ExecutionContext
 
+  def getBusinessDetailsUrl(nino: String): String = {
+    s"${config.itvcProtectedService}/income-tax-view-change/income-sources/$nino"
```
AlexRimmerHMRC, line 50: This won't work, the call to the backend will assume this is an mtd id
A seperate backend route will be needed

scullythornton1: fixed

app/models/incomeSourceDetails/IncomeSourceDetailsResponse.scala
```diff
@@ -25,7 +25,8 @@ sealed trait IncomeSourceDetailsResponse {
   def toJson: JsValue
 }
 
-case class IncomeSourceDetailsModel(businesses: List[BusinessDetailsModel],
+case class IncomeSourceDetailsModel(mtdbsa:Option[String],
```
AlexRimmerHMRC, line 28: As mentioned on the backend, mtdbsa in the api spec is not optional

PR #598, opened 2021-02-08
MISUV-1410 - Create confirm client view & controller

Wright93:


it/helpers/ComponentSpecBase.scala
```diff
@@ -125,5 +125,9 @@ trait ComponentSpecBase extends TestSuite with CustomMatchers
       }
     }
   }
+
+  def unauthorisedAgentTest(uri: String): Unit = {
+
+  }
```
AlexRimmerHMRC, line 131: Unused and not containing anything

PR #615, opened 2021-03-02
MISUV-1591 - Create Agent UTR Error page

Wright93:


app/controllers/agent/EnterClientsUTRController.scala
```diff
@@ -77,8 +77,8 @@ class EnterClientsUTRController @Inject()(enterClientsUTR: EnterClientsUTR,
                 ) ++ firstName.map(SessionKeys.clientFirstName -> _) ++ lastName.map(SessionKeys.clientLastName -> _)
                 Redirect(routes.ConfirmClientUTRController.show()).addingToSession(sessionValues: _*)
               case Left(CitizenDetailsNotFound | BusinessDetailsNotFound) =>
-                //todo: error page(s) will be implemented for these in future story
-                throw new InternalServerException("[EnterClientsUTRController][submit] - unable to verify client relationship")
+                val sessionValue: Seq[(String, String)] = Seq(SessionKeys.clientUTR -> validUTR)
+                Redirect(routes.UTRErrorController.show()).addingToSession(sessionValue: _*)
```
scullythornton1, line 81: I do not think this info needs to add to the session as it is not a valid value.

app/controllers/agent/UTRErrorController.scala
```diff
@@ -0,0 +1,72 @@
...
+                                   val appConfig: FrontendAppConfig,
+                                   val itvcErrorHandler: ItvcErrorHandler,
+                                   val ec: ExecutionContext)
+  extends SelectClientController with FeatureSwitching with I18nSupport {
+
+  def show: Action[AnyContent] = Authenticated.async { implicit request =>
+    implicit user =>
+    if (isEnabled(AgentViewer)) {
+      val clientUTR: Option[String] = request.session.get(SessionKeys.clientUTR)
+      clientUTR match {
```
scullythornton1, line 46: I do not think you will need to do this UTR check again as there is already a check in the entreClientUTR page. Should this just a matter of show this page and also when click the button the submit function can redirect the user back to entreClientUTR page.. And there is no session need to be removed as from the beginning we should not save this error UTR in the header. You can confirm with Alex.

PR #633, opened 2021-03-22
MISUV-1622: ITSA Submission link

AIbrahimQA:


app/config/FrontendAppConfig.scala
```diff
@@ -118,8 +118,10 @@ class FrontendAppConfig @Inject()(val servicesConfig: ServicesConfig, val config
   lazy val calcPollSchedulerTimeout: Int = servicesConfig.getInt("calculation-polling.timeout")
 
   // Submission service
-  lazy val submissionUrl: String = servicesConfig.baseUrl("income-tax-submission-frontend")
-  lazy val fullSubmissionUrl: String = s"$submissionUrl/income-through-software/return/2022/start"
+  lazy val submissionFrontendUrl: String = {
+    s"${servicesConfig.getString("income-tax-submission-frontend.url")}/income-through-software/return/2022/start"
```
chandraAndem, line 122: we do not need the string appending at the end i.e., everything needs to be in config.

conf/application.conf
```diff
@@ -266,6 +266,11 @@ payment-history {
   number-of-years = 7
 }
 
+income-tax-submission-frontend {
+  url = "http://localhost:9302"
```
chandraAndem, line 270: As mentioned above, URL here should be http://localhost:9302/income-through-software/return/2022/start

chandraAndem (changes requested):

chandraAndem (approved):

PR #661, opened 2021-04-20
MISUV- 1656: Charge Summary Audit Page

scullythornton1:


app/models/financialDetails/Charge.scala
```diff
@@ -16,57 +16,67 @@
...
+
+  val getChargeType: String = mainType match {
+    case Some("SA Payment on Account 1") => "Payment on account 1 of 2"
+    case Some("SA Payment on Account 2") => "Payment on account 2 of 2"
+    case Some("SA Balancing Charge") => "Remaining balance"
+    case error => {
+      Logger.error(s"[Charge][getChargeTypeKey] Missing or non-matching charge type: $error found")
+      "unknownCharge"
+    }
+  }
```
AlexRimmerHMRC, line 75: As this is a funciton which is solely ever used in the audit you have created, would you be able to move this functionality into that audit model as a function instead of here?

PR #671, opened 2021-04-26
Match Audit Json against an expected detail contents

MihailsC-hmrc:


it/helpers/servicemocks/WiremockMethods.scala
```diff
@@ -90,8 +90,7 @@ trait WiremockMethods {
   }
 
   def verifyContains[T](method: HTTPMethod, uri: String, body: T)(implicit writes: Writes[T]): Unit = {
-    val stringBody = writes.writes(body).toString()
-    verifyInternalContains(method, uri, Some(stringBody))
+    verifyInternalContainsJson(method, uri, Some(Json.toJson(body)))
```
AlexRimmerHMRC, line 93: Could this be reverted and left as is, with a new function which uses the new functionality
Reason, integration tests fail on the branch currently, keeping it as a separate function will mean people will only need to use this for their own tests.

PR #676, opened 2021-04-28
MISUV-1758 - Added the payment history request and response audits

AlexRimmerHMRC:


test/audit/models/PaymentHistoryRequestAuditModelSpec.scala
```diff
@@ -0,0 +1,98 @@
...
+    "Have the correct details for the audit event" when {
+      "the audit is full" when {
+        "the user is an individual" in {
+          paymentHistoryAuditFull(userType = Some("Individual")).detail shouldBe Json.obj(
+            "mtditid" -> testMtditid,
+            "nationalInsuranceNumber" -> testNino,
+            "saUtr" -> testSaUtr,
+            "credId" -> testCredId,
+            "userType" -> "Individual"
+          )
```
MihailsC-hmrc, line 76: Just a note.
I believe it's more readable when comparing with a json snippet rather than `Json.obj(...)`:
```Scala
          paymentHistoryAuditFull(userType = Some("Individual")).detail shouldBe Json.parse(
            s"""
               |{
               |  "mtditid": "$testMtditid",
               |  "nationalInsuranceNumber": "$testNino",
               |  "saUtr": "$testSaUtr"
               |  "credId": "$testCredId",
               |  "userType": "Individual"
               |}
               |""".stripMargin)
```

test/controllers/agent/PaymentHistoryControllerSpec.scala
```diff
@@ -119,17 +121,16 @@ class PaymentHistoryControllerSpec  extends TestSupport
...
-        status(result) shouldBe Status.SEE_OTHER
+      status(result) shouldBe Status.SEE_OTHER
 
-      }
     }
-
+  }
 
 
 }
```
MihailsC-hmrc, line 136: Optional suggestion: please add a newline at the end of the file.

test/controllers/PaymentHistoryControllerSpec.scala
```diff
@@ -55,6 +55,7 @@ class PaymentHistoryControllerSpec extends MockAuthenticationPredicate
       MockAuthenticationPredicate,
       app.injector.instanceOf[NinoPredicate],
       MockIncomeSourceDetailsPredicate,
+      mockAuditingService,
```
MihailsC-hmrc, line 58: Why don't we verify that `AuditingService` was called with expected values in controllers tests?

MihailsC-hmrc (commented):
Please read through the notes and questions. We'll merge after your answers. Thanks.

test/audit/models/PaymentHistoryRequestAuditModelSpec.scala
AlexRimmerHMRC: Personally I find it more readable being in Json.obj(...) etc.
But I also prefer creating any json through Json functions as there is less chance of making a mistake which results in a Json parse exception. If you take the snippet above as an example, the Json parse would result in a Json parse exception because there is a missing comma after the "saUtr" line, due to it entirely being in a string there is nothing to notify of things like that and is much more difficult to detect.

test/controllers/PaymentHistoryControllerSpec.scala
AlexRimmerHMRC: It's down to the effort required at this level I believe. To verify the audit in the controller, we have to verify the entire MtdItUser matches what we expect which contains lots of premade test data for most scenarios already. This only verifies that the auditing service gets the information we expect it to get, not that the resulting call was made or whether it contained the correct information.

In integration tests, in theory it becomes much simpler to match up the few fields you need and it also checks calls made + content.

MihailsC-hmrc (approved):

PR #677, opened 2021-04-28
[MISUV-1487] Individual What you owe page: Content when no charges are due

MihailsC-hmrc:
- Showing more notes to a user
- Link to view their L&P in a new tab
- Also when there are charges - adjusted link to view their L&P to open in a new tab, as per prototype


MihailsC-hmrc:
Assembled with Agent part of the story in the single PR #668

PR #695, opened 2021-05-10
MISUV-1863 Removal of CA_FRONTEND

stephenkward:


app/config/FrontendAppConfig.scala
```diff
@@ -67,7 +67,7 @@ class FrontendAppConfig @Inject()(val servicesConfig: ServicesConfig, val config
 
   //Sign out
   lazy val ggUrl: String = servicesConfig.getString(s"government-gateway.url")
-  lazy val ggSignOutUrl = s"$ggUrl/gg/sign-out?continue=$exitSurveyUrl"
+  lazy val ggSignOutUrl = s"$ggUrl/bas-gateway/sign-out-without-state?continue=$exitSurveyUrl"
```
chandraAndem, line 70: Could you change comment on line 60 appropiately? Also remove line 62.

app/controllers/SignInController.scala
```diff
@@ -32,6 +32,6 @@ class SignInController @Inject()(val appConfig: FrontendAppConfig,
                                 (implicit mcc: MessagesControllerComponents) extends FrontendController(mcc) with AuthRedirects {
 
   val signIn: Action[AnyContent] = Action.async { implicit request =>
-    Future.successful(toGGLogin(appConfig.ggSignInContinueUrl))
+    Future.successful(Redirect(appConfig.ggSignInContinueUrl))
```
chandraAndem, line 35: Sorry could we change it to something like this:
Redirect(config.ggLoginUrl, Map("continue_url" -> Seq(config.HomePage), "origin" -> Seq(config.appName)))

Then then change application.conf accordingly

chandraAndem (changes requested):

app/config/FrontendAppConfig.scala
stephenkward: resolved

chandraAndem (approved):

app/controllers/SignOutController.scala
```diff
@@ -41,7 +41,7 @@ class SignOutController @Inject()(config: FrontendAppConfig,
     }
       .map(contactFormIdentifier => Redirect(config.ggSignOutUrl(contactFormIdentifier)))
       .recover {
-        case _: AuthorisationException => Redirect(config.signInUrl)
+        case _: AuthorisationException => Redirect(config.ggSignInUrl)
```
chandraAndem, line 44: We need to keep signInUrl val in frontend config which is deleted and point to that rather than ggsigninurl as it doesn't have redirect params in it.

lazy val signInUrl: String = servicesConfig.getString("base.sign-in")

test/controllers/SignOutControllerSpec.scala
```diff
@@ -86,7 +86,7 @@ class SignOutControllerSpec extends TestSupport {
       }
 
       "redirect to the unauthorised sign out URL" in {
-        redirectLocation(result) shouldBe Some(appConfig.signInUrl)
+        redirectLocation(result) shouldBe Some(appConfig.ggSignInUrl)
```
chandraAndem, line 89: same as previous comment, this should not be changed.

chandraAndem (changes requested):
Signin url needs to be changed

chandraAndem (approved):

PR #709, opened 2021-05-21
MISUV-1921: Hook into API 1554 to retrieve data for Charge History

AIbrahimQA:


app/models/chargeHistory/ChargeHistoryResponseModel.scala
```diff
@@ -0,0 +1,38 @@
...
+
+package models.chargeHistory
+
+import play.api.libs.json.{Format, Json}
+
+sealed trait ChargeHistoryResponseModel
+
+
+case class ChargesHistoryModel(idType: String,
+															 idValue: String,
```
RoheelArshad, line 25: Can you move these inline

app/models/chargeHistory/ChargeHistoryModel.scala
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.chargeHistory
+
+import play.api.libs.json.{Format, Json}
+
+
+case class ChargeHistoryModel(documentId: String,
```
RoheelArshad, line 22: Comparing to the backend changes, does "tax-year" need to also be here?

PR #712, opened 2021-05-24
MISUV-1937: Remove enable obligations FS

AIbrahimQA:


RoheelArshad:
Changes look good, but should remove feature switch from application.conf

PR #727, opened 2021-06-07
MISUV-1885: Agent - Charge Summary page with charge history table

MihailsC-hmrc:


chandraAndem (changes requested):
All changes looks good but integration test for success scenario is missing when ChargeHistory FS is enabled. Please add that.

chandraAndem (approved):

PR #733, opened 2021-06-08
MISUV-2001: Connector parses errors in Get Charge History

MihailsC-hmrc:


MihailsC-hmrc:
As discussed before, we don’t need to recognise any non-OK response from the API. Even a "missing history data" case is covered within a successful API response.
Means that the changes are not needed

PR #744, opened 2021-06-23
MISUV-2030: Refactoring financialDetails and paymentAllocationCharges models

AIbrahimQA:


app/models/financialDetails/FinancialDetail.scala
```diff
@@ -17,13 +17,39 @@
...
-import java.time.LocalDate
-
-import play.api.Logger
 
 case class FinancialDetail(taxYear: String,
                            mainType: Option[String] = None,
-                           items: Option[Seq[SubItem]] = None)
+                           transactionId: Option[String] = None,
+                           transactionDate: Option[String] = None,
+                           `type`: Option[String] = None,
```
RoheelArshad, line 25: Is there any reason why type has single quotes?

PR #754, opened 2021-07-01
MISUV-2091 updated payment allocation charge model

david-walters-HMRC:


jameshforster (changes requested):
Does not include any changes which allows the model to combine the clearing date with the data from a payment allocation data set or the original charge. The new model needed the original charge from the first #1553 call, the payment allocations from #1413 and the clearing date from the subsequent #1553 calls as well.

PR #757, opened 2021-07-02
MISUV-2032: What you owe - Late payment interest

AIbrahimQA:


jameshforster (changes requested):
I am going to request two changes:
1. The charges need to be organised by due date and the late payment charges are not being sorted in this way. 
2. I saw a late payment being displayed which still had accruing interest, this should not be possible. Either the code needs to be changed to remove accruing interest on the late payment charges or the test data needs to be changed.

PR #762, opened 2021-07-12
MISUV-2089 added service layer for payment allocation page

david-walters-HMRC:


app/services/PaymentAllocationsService.scala
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+
+  def createPaymentAllocationWithClearingDate(nino: String, paymentCharge: PaymentAllocations)
+                                             (implicit hc: HeaderCarrier): Future[Seq[(PaymentAllocations, Option[String])]] = {
+    Future.sequence(paymentCharge.allocations map {
+      allocations =>
+        incomeTaxViewChangeConnector.getFinancialDataWithDocumentDetails(nino, allocations.transactionId.get) map {
+          case paymentAllocationChargesModel: FinancialDetailsWithDocumentDetailsModel =>
+            (paymentCharge, paymentAllocationChargesModel.paymentDetails.head.items.get.head.clearingDate)
```
jameshforster, line 67: This is not correct, for each allocation it cannot just get the paymentDetails.head, it must find the paymentDetails with a matching chargeType from the allocation. It also cannot just get items.get as it must also find the item which has a matching paymentLot and paymentLotItem as the first subitem will always be the charge data and not a payment allocation.

jameshforster (changes requested):

app/services/PaymentAllocationsService.scala
jameshforster: Please refer to the diagram attached to the story for more detail

PR #767, opened 2021-07-16
MISUV-2094 Display the payment allocations in payment history on the charge summary page

JSimmons072019:


app/controllers/ChargeSummaryController.scala
```diff
@@ -45,78 +47,88 @@ class ChargeSummaryController @Inject()(authenticate: AuthenticationPredicate,
...
-          }
+        financialDetailsService.getFinancialDetails(taxYear, user.nino).flatMap {
+          case success: FinancialDetailsModel if success.documentDetails.exists(_.transactionId == id) =>
+            val backLocation = user.session.get(SessionKeys.chargeSummaryBackPage)
+            val documentDetail = success.documentDetails.find(_.transactionId == id).get
+
+            val paymentAllocationEnabled: Boolean = isEnabled(PaymentAllocation)
+            val paymentAllocations: List[PaymentsWithChargeType] =
+              if (paymentAllocationEnabled) {
+                success.financialDetails.filter(_.transactionId.contains(id)).flatMap(_.allocation)
```
chandraAndem, line 73: Can you change this to '==' rather than 'contains' as there could be 'id' which partially match other id's which we don't know?

app/views/chargeSummary.scala.html
```diff
@@ -130,6 +133,17 @@ <h2 class="heading-large">@messages("chargeSummary.chargeHistory.heading")</h2>
...
 					</tr>
 				}
+			}
+
+				@if(paymentAllocationEnabled) {
+					@for(allocation <- paymentAllocations ; payment <- allocation.payments) {
+					<tr>
+						<td style="white-space: nowrap;">@LocalDate.parse(payment.date.get).toLongDateShort</td>
+						<td>@allocation.getPaymentAllocationTextInChargeSummary.map(messages(_))</td>
+						<td class="numeric">@payment.amount.map(_.abs.toCurrency)</td>
```
chandraAndem, line 143: Could you change it to payment.amount.get? By doing that we error out in this page rather than displaying blank value just like line 141 (or) it might be worth checking for nonEmpty values all the row values before displaying.

chandraAndem (changes requested):

app/controllers/predicates/AuthenticationPredicate.scala
```diff
@@ -85,7 +85,7 @@ class AuthenticationPredicate @Inject()(implicit val ec: ExecutionContext,
 				Logger.info("[AuthenticationPredicate][async] Unauthorised request. Redirect to Sign In.")
 				Redirect(controllers.routes.SignInController.signIn())
 			case s =>
-				Logger.error(s"[AuthenticationPredicate][async] Unexpected Error Caught. Show ISE.\n${s.getMessage}\n\n$s\n")
+				Logger.error(s"[AuthenticationPredicate][async] Unexpected Error Caught. Show ISE.\n$s\n", s)
```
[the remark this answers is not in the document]
chandraAndem: Resolving as it prints stack trace which is helpful

app/controllers/ChargeSummaryController.scala
chandraAndem: My bad, as I thought transactionId is a string.

chandraAndem (approved):

PR #795, opened 2021-07-27
Migrating testonly views to be Injectable

chandraAndem:


app/testOnly/views/helpers/FieldErrorHelper.scala.html
```diff
@@ -18,6 +18,8 @@
 @import play.api.data.Field
 @import testOnly.forms.validation.ErrorMessageHelper._
 
+@this()
```
JSimmons072019, line 21: Why is this here

chandraAndem: @this() is how you make a template injectable. With or without arguments.

JSimmons072019 (approved):

PR #826, opened 2021-08-18
MISUV-2271 - Allowing the EnterClient UTR page to accept spaces

Wright93:


app/forms/agent/ClientsUTRForm.scala
```diff
@@ -57,4 +57,10 @@ object ClientsUTRForm {
     )
   )
 
+  def removeWhiteSpaces(data: Map[String, Seq[String]]): Map[String, Seq[String]] = {
```
MihailsC-hmrc, line 60: I reckon this is quite an intrusive way to solve the task. I would suggest modification of the form validation only, targeted to sanitising the specific field rather than modifying the whole request.
Something like the following in the form:
```Scala
      utr -> default(text, "")
        .transform[String](_.replaceAll("\\s", ""), identity)
        .verifying(...)
```
The tests in the PR are still fine, but will need to add couple of test cases to the form spec.

MihailsC-hmrc (changes requested):
Please consider form validation changes only without changing controller.

MihailsC-hmrc (approved):

PR #827, opened 2021-08-19
MISUV-2211 corrected caption positioning

david-walters-HMRC:


jack-malik (commented):
The table pattern specified here  https://design-system.service.gov.uk/components/table/ uses no bold attribute .. Instead, it relies on <thead>..</thead> to achieve bold font .. If we are strictly to adhere to the table pattern perhaps we should rely on the <thead> instead of bold attribute

jack-malik (approved):
Required changes needed as per the prototype look given.

jack-malik:
Rebasing 1 commit and merging.

PR #847, opened 2021-08-31
Added caption to Payments tab on TaxYearOverview screen.

jack-malik:


jack-malik:
Verified with auto signature

PR #869, opened 2021-09-15
Adding balance details to WhatYouOwe page changes

chandraAndem:


public/stylesheets/itvc.css
```diff
@@ -1665,3 +1665,23 @@ margin-bottom: 0 !important;
...
+
+.what-you-owe-total-values {
+    font-size: 36px !important;
+    line-height: 1.11 !important;
+    font-weight: bold;
+}
+
+.what-you-owe-total-values-bold {
+    font-weight: 700 !important;
+}
```
MihailsC-hmrc, line 1687: Just a note: these styles probably won't be required after migration to the new frontend system. Or probably they'll be replicated in new application.css during migration of the page.

app/models/financialDetails/FinancialDetailsResponseModel.scala
```diff
@@ -91,7 +93,7 @@ case class FinancialDetailsModel(documentDetails: List[DocumentDetail],
   }
 
   def merge(financialDetailsModel: FinancialDetailsModel): FinancialDetailsModel = {
-    FinancialDetailsModel(documentDetails ++ financialDetailsModel.documentDetails,
+    FinancialDetailsModel(balanceDetails, documentDetails ++ financialDetailsModel.documentDetails,
```
MihailsC-hmrc, line 96: Debatable: this function does not merge `balanceDetails` anyhow. Would it be more feasible to rename it to something like `def mergeLists`?

MihailsC-hmrc (commented):
Need to clarify a question before approving the PR.

public/stylesheets/itvc.css
chandraAndem: Yes agree, will let Jack who is working on migration know about it.

app/models/financialDetails/FinancialDetailsResponseModel.scala
chandraAndem: Agreed and made those changes.

MihailsC-hmrc (approved):

PR #871, opened 2021-09-16
Misuv 2398

RoheelArshad:


app/views/agent/UTRError.scala.html
```diff
@@ -17,26 +17,32 @@
...
+
+
 
 @this(
-    mainTemplate: MainTemplate,
+    mainTemplate: layout,
     continueButton: ContinueButton,
-    form: FormWithCSRF
+    form: FormWithCSRF,
+h1:h1,
```
david-walters-HMRC, line 29: adjust the indentation to match all the parameters

PR #960, opened 2021-10-29
MiISUV-2626 fixed personal allowance retrieving wrong value in the deduction table

david-walters-HMRC:


app/models/calculation/Calculation.scala
```diff
@@ -171,11 +173,17 @@ case class AllowancesAndDeductions(personalAllowance: Option[BigDecimal] = None,
 
   val totalAllowancesDeductionsReliefs: Option[BigDecimal] = (totalAllowancesAndDeductions ++ totalReliefs).reduceOption(_ + _)
 
+  val personalAllowanceDisplayValue: Option[BigDecimal] =
```
jameshforster, line 176: If you are using this fold, you don't need the if statement around it. This logic would fail in the scenario where marriage allowance is not defined but the user has a reducedPersonalAllowance. This scenario would happen in any case where a reduced personal allowance is present because marriage allowance transfer cannot exist in this scenario.

jameshforster (changes requested):

jameshforster (approved):

PR #1011, opened 2021-11-30
Fixing issue where class 2 nics is shown as Remaining balance when FS…

chandraAndem:
… is disabled and coding out details exists

### DIFF
--- a/app/services/FinancialDetailsService.scala
+++ b/app/services/FinancialDetailsService.scala
@@ -87,13 +87,17 @@ class FinancialDetailsService @Inject()(val incomeTaxViewChangeConnector: Income
       chargesWithYears.flatMap {
         case (_, errorModel: FinancialDetailsErrorModel) => Some(errorModel)
         case (_, financialDetails: FinancialDetailsModel) =>
-          val unpaidDocumentDetails: List[DocumentDetail] = financialDetails.documentDetails.collect {
-            case documentDetail: DocumentDetail if isEnabled(CodingOut) && documentDetail.isCodingOut => documentDetail
-            case documentDetail: DocumentDetail if documentDetail.latePaymentInterestAmount.isDefined && !documentDetail.interestIsPaid => documentDetail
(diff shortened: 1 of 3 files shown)


app/services/FinancialDetailsService.scala
```diff
@@ -87,13 +87,17 @@ class FinancialDetailsService @Inject()(val incomeTaxViewChangeConnector: Income
...
+          if (unpaidDocDetails.nonEmpty) Some(financialDetails.copy(documentDetails = unpaidDocDetails)) else None
       }
     }
   }
+
+  private def unpaidDocumentDetails(financialDetailsModel: FinancialDetailsModel): List[DocumentDetail] = {
+    financialDetailsModel.documentDetails.collect {
+      case documentDetail: DocumentDetail if documentDetail.isCodingOutDocumentDetail(isEnabled(CodingOut)) => documentDetail
+      case documentDetail: DocumentDetail if documentDetail.latePaymentInterestAmount.isDefined && !documentDetail.interestIsPaid => documentDetail
+      case documentDetail: DocumentDetail if !documentDetail.isClass2Nic && !documentDetail.isCodingOut && !documentDetail.isPaid => documentDetail
```
joesheridan, line 100: Could a document be class2Nic and NOT be coding out?
Is this not more what we want? : 
if !(documentDetail.isClass2Nic && documentDetail.isCodingOut) && !documentDetail.isPaid...

joesheridan (changes requested):

app/services/FinancialDetailsService.scala
chandraAndem: Ideally there should not be any case in happy path where class2Nic exists but coding out(PAYE) doesn't exists but it could exists in unhappy and complex paths.
Also coming to which condition to use, for condition 3 on line 100 it doesn't matter whether coding out details co-exists or not, we should only collect charges which are not coding out and not class 2 nics. Could discuss more in call if needed.

joesheridan (approved):

PR #1025, opened 2021-12-09
Misuv 2805

chandraAndem:


chandraAndem:
test this please

app/views/ChargeSummary.scala.html
```diff
@@ -103,20 +127,31 @@ <h1 class="govuk-heading-xl">
                 @messages("chargeSummary.codingOutInset-3")
             </p>
         </div>
-        @p(id=Some("coding-out-message")){
-            @messages("chargeSummary.codingOutMessage", taxYearFrom, taxYearTo)
-        }
+        @codingOutMessage
+    }
+
+    @if(codingOutEnabled && documentDetail.isCancelledPayeSelfAssessment){
```
JSimmons072019, line 133: why aren't components being used, like "@link" or "@p"

chandraAndem: This is because for P tag any ways we are using custom class, so to me helper doesn't really matter here as either we need to use it in html tags or braces, so I personally preferred using html tags. But happy to have a discussion on that with team.
Also @link tag to me is of less use as we have to pass in all arguments. But again happy to go with whatever team wants/decides. But that should be part of another ticket because its inconsistent across all templates.

JSimmons072019: I'll merge for now and we'll discuss further later

JSimmons072019 (approved):

PR #1041, opened 2021-12-24
MISUV-2872 Fix the bug with overdue warning message is not appearing

andriikondiuk:
Bug fix for not showing Due Payments warning on Home page when logged as an Agent

### DIFF
--- a/test/controllers/agent/HomeControllerSpec.scala
+++ b/test/controllers/agent/HomeControllerSpec.scala
@@ -64,6 +64,8 @@ class HomeControllerSpec extends TestSupport
       app.injector.instanceOf[FrontendAppConfig],
       mockItvcErrorHandler,
       app.injector.instanceOf[ExecutionContext])
+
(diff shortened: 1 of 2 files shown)


test/controllers/agent/HomeControllerSpec.scala
```diff
@@ -64,6 +64,8 @@ class HomeControllerSpec extends TestSupport
       app.injector.instanceOf[FrontendAppConfig],
       mockItvcErrorHandler,
       app.injector.instanceOf[ExecutionContext])
+
+		val overdueWarningMessage: String = "! You have overdue payments and one or more of your tax decisions are being reviewed. You may be charged interest on these until they are paid in full."
```
joesheridan, line 68: we could get this from the injected messages service no?

andriikondiuk: Good one, thank you. Done

AIbrahimQA (approved):

PR #1043, opened 2022-01-06
MISUV-fix-2022 - fixed tests failing after the new year

Wright93:


chandraAndem (changes requested):
Just one change needed

chandraAndem (approved):

PR #1044, opened 2022-01-06
MISUV 2839 unify homepage and combine logic of Agent and Individual home controllers

andriikondiuk:


app/views/errorPages/templates/ErrorTemplate.scala.html
```diff
@@ -14,11 +14,11 @@
  * limitations under the License.
  *@
 
-@import views.html.layouts.layout
+@import views.html.layouts.unifiedLayout
 @import views.html.components._
 
 @this(
-      mainTemplate: layout,
+      mainTemplate: unifiedLayout,
```
chandraAndem, line 21: We need to pass isAgent to this template as well as we are using it in both Individual and Agent.

chandraAndem (changes requested):
Just one small change where isAgent needs to be passed.

app/views/errorPages/templates/ErrorTemplate.scala.html
andriikondiuk: Thanks it's been done

chandraAndem (approved):

PR #1061, opened 2022-01-25
MISUV-2924 Fix tiles width on mobile devices

andriikondiuk:


chandraAndem (approved):

chandraAndem (changes requested):
Changes looks good but needs a small change i.e., when clicked on View your current 2021 to 2022 return link from Home page and then click Back link its taking us to Tax years page rather than Home page.

chandraAndem (approved):

PR #1069, opened 2022-01-31
Changes for TaxYearOverviewpage to integrate with new tax calc proxy

chandraAndem:


app/audit/models/AllowanceAndDeductionsResponseAuditModel.scala
```diff
@@ -19,9 +19,9 @@ package audit.models
 import audit.Utilities.userAuditDetails
 import auth.MtdItUserBase
 import models.calculation.AllowancesAndDeductions
+import models.liabilitycalculation.viewModels.AllowancesAndDeductionsViewModel
```
joesheridan, line 22: package names lower case? -> viewmodels

app/controllers/TaxYearOverviewController.scala
```diff
@@ -19,25 +19,29 @@ package controllers
 import audit.AuditingService
 import audit.models.TaxYearOverviewResponseAuditModel
 import auth.MtdItUser
-import config.featureswitch.{CodingOut, FeatureSwitching, TxmEventsApproved}
```
joesheridan, line 22: are we removing txmeventsapproved feature switch?

app/audit/models/AllowanceAndDeductionsResponseAuditModel.scala
chandraAndem: Corrected now

app/controllers/TaxYearOverviewController.scala
chandraAndem: Yes, these are actually added at the time of Release 6. So whereever possible we could remove it.

PR #1089, opened 2022-02-08
MISUV-2947: Adding integration tests for checking the Agent/Ind specific text in heading

AIbrahimQA:


jack-malik:
Went through all file changes .. All clear .. Recap below: 

1/. In most cases replacement of pageTitle with pageTitleIndividual and pageTitleAgent in the code. 
2/. Introduction of new messages files under it/testConstants/messages. 
3/. Modifications to imports where needed. 
4/. All correct programmatically. 
5/. Introduction of pageTitleIndividual() and pageTitleAgent() functions not necessary. 
Extending **pageTitle()** function (now removed) to take a boolean flag indicating if agent 
context or not, would be sufficient and result in less changes and code written.

PR #1094, opened 2022-02-10
Modified all instances of opening page in new tab to contain rel attr…

jack-malik:
…ibute and added the '(opens in new tab)' text.

### DIFF
--- a/app/views/TaxYears.scala.html
+++ b/app/views/TaxYears.scala.html
@@ -87,7 +87,9 @@
         @utr.map { utrValue =>
             @p(id = Some("oldSa-para")) {
                 @messages("taxYears.oldSa.content.text")
-                @link(link = appConfig.saViewLandPService(utrValue), messageKey = "taxYears.oldSa.content.link", id = Some("self-assessment-online-account-link"), outerMessage = ".", target = Some("_blank"))
+                @link(link = appConfig.saViewLandPService(utrValue), messageKey = "taxYears.oldSa.content.link",
+                    messageKeyExt = "pagehelp.opensInNewTabText",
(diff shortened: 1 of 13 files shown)


app/views/TaxYears.scala.html
```diff
@@ -87,7 +87,9 @@
         @utr.map { utrValue =>
             @p(id = Some("oldSa-para")) {
                 @messages("taxYears.oldSa.content.text")
-                @link(link = appConfig.saViewLandPService(utrValue), messageKey = "taxYears.oldSa.content.link", id = Some("self-assessment-online-account-link"), outerMessage = ".", target = Some("_blank"))
+                @link(link = appConfig.saViewLandPService(utrValue), messageKey = "taxYears.oldSa.content.link",
+                    messageKeyExt = messages("pagehelp.opensInNewTabText", " ", ""),
+                    id = Some("self-assessment-online-account-link"), outerMessage = ".", rel=Some("noreferrer noopener"), target = Some("_blank"))
```
david-walters-HMRC, line 92: since the (open in new tab) is mandatory for for a link opening in a new tab, we should rework it into here rather then keep parsing it into every instance

jack-malik: David .. **rework it into here** - not sure what you mean. The reason why it is passed the way I invented is because (opens in new tab) appears in different context. Sometimes as end of the sentence, sometimes in the middle of a text, sometimes requires  dot, sometimes it doesn't. Moreover, existing tests sometimes have expected text to contain dot, sometimes they don't. Withouth rewriting all this I don't know what your solution is. 
Also: as part of your story we needed to identify other pages requiring (opens in new tab) .. I identified one such page and I'm adding (opens in new tab) to it.

david-walters-HMRC: that's okay if your unsure to do that, we can pair on it tommorrow

PR #1120, opened 2022-02-25
MISUV-3065 made ariaLabel compulsory and updated links

RoheelArshad:


AIbrahimQA (approved):

app/views/components/link.scala.html
```diff
@@ -22,7 +22,7 @@
     messageKey: String,
     id: Option[String] = None,
     target: Option[String] = None,
-    ariaLabel: Option[String] = None,
```
joesheridan, line 25: I think we agreed we'd just default the value to the link text

app/views/agent/confirmClient.scala.html
```diff
@@ -56,7 +56,8 @@
         </div>
     </dl>
 	@p(){
-		@link(link = backUrl, messageKey = "agent.confirmClient.changeClient")
```
joesheridan, line 59: I think we also agreed that this ticket was only for the home page

PR #1130, opened 2022-03-01
MISUV-3065-1 reverted mandatory changes and added default value to be…

RoheelArshad:
… link text

### DIFF
--- a/app/views/components/link.scala.html
+++ b/app/views/components/link.scala.html
@@ -34,7 +34,7 @@
     @if(ariaLabel == None) {
         aria-label="@messages(messageKey).replaceAll(" ", "-")@if(target == Some("_blank")){@messages("pagehelp.opensInNewTabText").replaceAll(" ", "-")}">
     } else {
-    aria-label="@ariaLabel.map(c=>s"$c")">
+            @ariaLabel.map(lbl => s"aria-label=$lbl")">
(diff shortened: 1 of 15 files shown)


app/views/components/link.scala.html
```diff
@@ -31,5 +31,10 @@
 
 <a href="@link" class="@classes" @id.map(id => s"id=$id") rel="@rel.map(c=>s"$c")" @target.map(target => s"target=$target")
     @role.map(role => s"role=$role")
+    @if(ariaLabel == None) {
```
joesheridan, line 34: advisory only - we can swap the if blocks around using: if(ariaLabel.isDefined)

app/views/components/link.scala.html
```diff
@@ -31,5 +31,10 @@
...
+        aria-label="@messages(messageKey).replaceAll(" ", "-")@if(target == Some("_blank")){@messages("pagehelp.opensInNewTabText").replaceAll(" ", "-")}">
+    } else {
     aria-label="@ariaLabel.map(c=>s"$c")">
```
joesheridan: advisory only - param name "c" makes no sense here, perhaps just:
.map(label=>s"$label")
instead

joesheridan (approved):

PR #1154, opened 2022-03-15
MISUV-3121 Unify DeductionsSummaryController

andriikondiuk:


app/controllers/DeductionsSummaryController.scala
```diff
@@ -20,59 +20,82 @@ import audit.AuditingService
...
+      implicit agent =>
+        handleRequest(
+          backUrl = controllers.agent.routes.TaxYearOverviewController.show(taxYear).url,
+          itcvErrorHandler = itvcErrorHandlerAgent,
+          taxYear = taxYear,
+          isAgent = true
+        )(getMtdItUserWithNino()(agent, request, implicitly), implicitly, implicitly, implicitly)
+    }
+  }
+}
```
joesheridan, line 101: newline pls :)

joesheridan (approved):

app/controllers/DeductionsSummaryController.scala
andriikondiuk: Done, thanks. But intelij Idea Default Scala style formatting is not adding extra line here. And I thought we've agreed to use Ctrl + Alt + L for the default format

PR #1179, opened 2022-04-01
MISUV-3222 make integration tests handle any future current date

joesheridan:


it/controllers/TaxYearsControllerISpec.scala
```diff
@@ -61,10 +62,11 @@ class TaxYearsControllerISpec extends ComponentSpecBase with FeatureSwitching {
           verifyIncomeSourceDetailsCall(testMtditid)
 
           Then("The view should have the correct headings and all tax years display")
+          val expectedYearsShown = getCurrentTaxYearEnd.getYear - propertyAccountingStartLocalDate.getYear + 1
```
chandraAndem, line 65: I think changing propertyAccountingStartLocalDate value to **currentTaxYear - 5** will be better. Also other fields in the PropertyDetailsIntegrationTestConstants file needs to be changed accordingly. But if its taking more time then will leave it.

chandraAndem (changes requested):

it/controllers/TaxYearsControllerISpec.scala
joesheridan: updated

chandraAndem (approved):

PR #1213, opened 2022-04-27
Filtering cut over credits from Tax year summary payments tab

chandraAndem:


test/models/financialDetails/DocumentDetailSpec.scala
```diff
@@ -58,6 +58,22 @@ class DocumentDetailSpec extends UnitSpec {
...
+        }
+        "original amount is negative" in {
+          fullDocumentDetailModel.copy(originalAmount = Some(-20)).originalAmountIsNotZeroOrNegative shouldBe false
+        }
+      }
+      "return true" when {
+        "original amount is positive" in {
+          fullDocumentDetailModel.copy(originalAmount = Some(20)).originalAmountIsNotZeroOrNegative shouldBe true
+        }
+      }
```
joesheridan, line 74: we are missing the case where originalAmount is None - should it really be true here (and therefore will be included in the filtered results)?

chandraAndem: Functionally originalAmount should never be None and if backend is not passing that data then we should be throwing an exception, which is what we are doing in the view. Let me know if you think otherwise and happy to discuss.

joesheridan (approved):

PR #1278, opened 2022-06-07
MISUV-3554: Tax year summary spacing

AIbrahimQA:


jack-malik:
Ahmed .. I think we agreed to tabs being 4 spaces .. Please have a look at lines 31-40, 44-46  as well as 136-140 if they are consistent with 4 .. I think these are 8 and even though it's minor let me know your thoughts. I'm editing WhatYouOwe and want those to be consistent.

PR #1288, opened 2022-06-13
Misuv 3572: create testOnly endpoint to add calcId to user session

avelx:


avelx:
Every commit must be Gpg signed, so dropping this PR and raising another one with commits signed etc;

PR #1314, opened 2022-07-06
MISUV-3665 Update WhatYouOweService to handle MFA Debits

edward-dennis-ey:


app/services/WhatYouOweService.scala
```diff
@@ -40,7 +40,11 @@ class WhatYouOweService @Inject()(val financialDetailsService: FinancialDetailsS
     documentDescription == "ITSA- POA 1" ||
       documentDescription == "ITSA - POA 2" ||
       documentDescription == "TRM New Charge" ||
-      documentDescription == "TRM Amend Charge"
+      documentDescription == "TRM Amend Charge" ||
```
joesheridan, line 43: this code needs to depend on the feature switch - also it needs to be unit tested and integration tested
I'd suggest passing in a MFADebitsEnabled boolean from the controller, so that its easier to unit test the service

PR #1319, opened 2022-07-11
MISUV-3719 refactor and tests for payment history refunds

joesheridan:


test/models/repaymentHistory/RepaymentHistoryUtilsSpec.scala
```diff
@@ -0,0 +1,148 @@
...
+import models.financialDetails.Payment
+import models.repaymentHistory.RepaymentHistoryUtils._
+import org.scalatest.Matchers
+import testUtils.{TestSupport}
+
+import java.time.LocalDate
+
+class RepaymentHistoryUtilsSpec extends TestSupport with Matchers {
+
+  val paymentsWithMFA: List[Payment] = List(
```
avelx, line 29: really optional, but worth to consider moving data into ~Helper trait, in order to be able re-use data in the future etc;

app/models/repaymentHistory/RepaymentHistoryUtils.scala
```diff
@@ -43,9 +43,41 @@ object RepaymentHistoryUtils {
...
+        .map { case (_, payments) => payments }
+    }
+
+    payments.groupBy[Int] { payment => {
+      LocalDate.parse(payment.date).getYear
+    }
+    }.toList.sortBy(_._1).reverse
+      .map { case (year, payments) => (year, sortPayments(payments)) }
+  }
+
```
avelx, line 68: I guess this grouping would make more sense on the model level or as a helper function / utils i.e?

avelx (approved):
looks good otherwise

PR #1359, opened 2022-08-05
[MISUV-3804]: sealed trait implemented for AuditType

ayubMercator:


app/enums/AuditType.scala
```diff
@@ -0,0 +1,93 @@
...
+
+case object LowConfidenceLevelIvOutcomeSuccess extends AuditType {
+  val name = "LowConfidenceLevelIvOutcomeSuccess"
+}
+
+case object LowConfidenceLevelIvHandoff extends AuditType {
+  val name = "LowConfidenceLevelIvHandoff"
+}
+
+case object obligationsPageView extends AuditType {
```
avelx, line 55: should be fine to use upper casing in this instance for type name

app/enums/AuditType.scala
```diff
@@ -0,0 +1,93 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package enums
+
+sealed trait AuditType {
```
avelx, line 19: pls define implicit conversion function from AuditType to String and avoid using .name while calling required string value of the etc

avelx (changes requested):
* minor changes requested

PR #1378, opened 2022-08-19
MISUV-3863 Show cutOver credit status

andriikondiuk:


app/models/financialDetails/DocumentDetail.scala
```diff
@@ -96,7 +96,7 @@ case class DocumentDetail(taxYear: String,
     else isPaid
   }
 
-  val isPartPaid: Boolean = outstandingAmount.getOrElse[BigDecimal](0) != originalAmount.getOrElse[BigDecimal](0)
+  val isPartPaid: Boolean = outstandingAmount.getOrElse[BigDecimal](0).abs != originalAmount.getOrElse[BigDecimal](0).abs
```
avelx, line 99: what would happen if one value negative and one positive?
I would rather add unit test to see various scenarios

avelx (changes requested):

app/models/financialDetails/DocumentDetail.scala
andriikondiuk: Agree it's a valid question. thanks

andriikondiuk: I've removed the .abs and added a unit tests, thanks

test/testConstants/FinancialDetailsTestConstants.scala
```diff
@@ -938,15 +938,15 @@ object FinancialDetailsTestConstants {
       outstandingAmount = Some(BigDecimal(-500.00)),
       paymentLotItem = None,
       paymentLot = None,
-      originalAmount = Some(BigDecimal(800.00)),
+      originalAmount = Some(BigDecimal(-800.00)),
```
avelx, line 941: changes in this file seems to be not relevant ( if you rollback them back test would not fail)

avelx (approved):
looks good

PR #1386, opened 2022-08-24
MISUV 3824: Update credits views

ayubMercator:


app/models/CreditDetailModel.scala
```diff
@@ -17,14 +17,21 @@
...
 
 case class CreditDetailModel(date: LocalDate , documentDetail: DocumentDetail, creditType: CreditType)
 
-sealed trait CreditType
+sealed trait CreditType {
+  val creditTypeAsString: String
+}
+
+case object MfaCreditType extends CreditType {
+  override val creditTypeAsString = "Credit from HMRC adjustment"
```
avelx, line 31: we should use language resources where required by the context

app/controllers/CreditsSummaryController.scala
```diff
@@ -109,20 +85,26 @@ class CreditsSummaryController @Inject()(creditsView: CreditsSummary,
...
+            calendarYear = calendarYear,
+            backUrl = if (isAgent) getAgentBackURL(user.headers.get(REFERER), calendarYear) else getBackURL(user.headers.get(REFERER), origin, calendarYear),
+            isAgent = isAgent,
+            utr = user.saUtr,
+            btaNavPartial = user.btaNavPartial,
+            enableMfaCreditsAndDebits = isEnabled(MFACreditsAndDebits),
+            charges = charges,
+            origin = origin)))
+        case Left(_) => {
+          Logger("application").error(s"Could not retrieve financial details for year: $calendarYear")
```
avelx, line 102: we might need to include Nino as well in the log message

app/models/repaymentHistory/RepaymentHistoryUtils.scala
```diff
@@ -34,8 +34,8 @@ object RepaymentHistoryUtils {
     }
   }
 
-  private def getMFACreditsLink(date: LocalDate, isAgent: Boolean) = {
-    val year = date.getYear
+  private def getMFAandCutoverCreditsLink(date: String, isAgent: Boolean) = {
+    val year = LocalDate.parse(date).getYear
```
avelx, line 38: pls check up what the caller stack, we converting LocalDate to String and the parsing again to LocalDate and extracting date, its looks a bit not logical; pls review this logic

app/models/repaymentHistory/RepaymentHistoryUtils.scala
```diff
@@ -87,17 +87,18 @@ object RepaymentHistoryUtils {
               date = payment.documentDate,
               description = "paymentHistory.mfaCredit",
               amount = payment.amount,
-              linkUrl = getMFACreditsLink(payment.documentDate, isAgent),
+              linkUrl = getMFAandCutoverCreditsLink(payment.documentDate.toString, isAgent),
```
avelx, line 90: as per comment above/below: no need to convert to string at all

avelx (changes requested):
as per comments provided

app/models/CreditDetailModel.scala
ayubMercator: not sure how to implement this? We don't have access to messages in CreditDetailModel.scala

ayubMercator: I'll replace the text with the messages key

avelx (approved):
looks good

PR #1399, opened 2022-09-02
MISUV-3921 Improve coverage for NextUpdatesResponseModel

andriikondiuk:


avelx:
@andriikondiuk =>  you should be able to drop NextUpdateModel.currentTime as it redundant and not in use anywhere else, so this would raise test coverage to 100% for all 3 positions

avelx (approved):
minor comment, otherwise looks good

PR #1400, opened 2022-09-02
MISUV-3896 : Tax Year Summary Date Displaying Incorrectly

AshokGitAccount:


app/models/liabilitycalculation/viewmodels/asd.sc
```diff
@@ -0,0 +1,38 @@
...
+  val fourthQuarterEnd = LocalDate.parse(s"${taxYear}-04-05")
+  //2019-02-15T09:35:15.094Z
+   val df = java.time.format.DateTimeFormatter.ofPattern("yyyy-mm")
+  val formatter = java.time.Instant.parse(calculationTimestamp)
+  val date = LocalDateTime.ofInstant(formatter,ZoneId.of(ZoneOffset.UTC.getId)).toLocalDate
+  println(thirdQuarterEnd.getMonthValue+"!!!" +date.getMonthValue)
+  println(SecondQuarterEnd.getMonthValue+"111" +date)
+  println(firstQuarterEnd.getMonthValue+"22" +date)
+  println(fourthQuarterEnd.getMonthValue+"33" +date)
+  //println( "11111" + date.isBefore(thirdQuarterEnd))
```
avelx, line 21: remove commented code pls

app/models/liabilitycalculation/viewmodels/asd.sc
```diff
@@ -0,0 +1,38 @@
...
+  val firstQuarterstart = LocalDate.parse(s"${taxYear - 1}-04-05")
+  val firstQuarterEnd = LocalDate.parse(s"${taxYear - 1}-07-05")
+  val SecondQuarterEnd = LocalDate.parse(s"${taxYear - 1}-10-05")
+  val thirdQuarterEnd = LocalDate.parse(s"${taxYear}-01-05")
+  val fourthQuarterEnd = LocalDate.parse(s"${taxYear}-04-05")
+  //2019-02-15T09:35:15.094Z
+   val df = java.time.format.DateTimeFormatter.ofPattern("yyyy-mm")
+  val formatter = java.time.Instant.parse(calculationTimestamp)
+  val date = LocalDateTime.ofInstant(formatter,ZoneId.of(ZoneOffset.UTC.getId)).toLocalDate
+  println(thirdQuarterEnd.getMonthValue+"!!!" +date.getMonthValue)
```
avelx, line 17: I don't think we need this test file at all, pls remove; in general these need to be wrapped in the relevant unit test, so looks like something is missing

Joshtest1570 (approved):

app/views/TaxYearSummary.scala.html
```diff
@@ -282,7 +282,11 @@ <h1 class="govuk-heading-xl" id="heading">
                     @messages("tax-year-summary.total-due")
                 } else {
                     @{
-                        messages("tax-year-summary.tax-calculation.date", s"${taxYear - 1}", s"$taxYear")
+                        if(model.inYearCalculationRunDate.isDefined) {
```
avelx, line 285: are you sure this would work for Welsh ??

app/models/liabilitycalculation/viewmodels/TaxYearSummaryViewModel.scala
```diff
@@ -44,9 +43,35 @@ object TaxYearSummaryViewModel {
...
     )
   }
+
+  def isUnattendedCalc(calculationReason: Option[String]): Boolean = calculationReason match {
+    case Some("unattendedCalculation") => true
+    case _ => false
+  }
+
+  def getInYearCalculationRunDate(calculationTimestamp: String, taxYear: Int): Option[String] = {
+    val firstQuarterstart = LocalDate.parse(s"${taxYear - 1}-04-05")
```
avelx, line 57: you might scratch surface for the next ticket => https://jira.tools.tax.service.gov.uk/browse/MISUV-3215
I would suggest to create singleton DateService and create lambdas for: firstQuarterEnd .. to fourthQuarterEnd, so these could be re-used + method to convert calculationTimestamp into Either[Error,LocalDate];

Also I would think that calculationTimestamp method could be move under DateService singleton as well and you can call it were required; 
Maybe worth to model new data type. ie. enum for year quarters? not sure if we can trust/rely on strings

avelx (changes requested):
changes requested

app/models/liabilitycalculation/viewmodels/TaxYearSummaryViewModel.scala
```diff
@@ -44,9 +43,35 @@ object TaxYearSummaryViewModel {
...
+    val firstQuarterEnd = LocalDate.parse(s"${taxYear - 1}-07-05")
+    val SecondQuarterEnd = LocalDate.parse(s"${taxYear - 1}-10-05")
+    val thirdQuarterEnd = LocalDate.parse(s"${taxYear}-01-05")
+    val fourthQuarterEnd = LocalDate.parse(s"${taxYear}-04-05")
+
+    val instant = java.time.Instant.parse(calculationTimestamp)
+    val date = LocalDateTime.ofInstant(instant,ZoneId.of(ZoneOffset.UTC.getId)).toLocalDate
+
+    date match {
```
avelx, line 66: I would guess this part of the logic need to be unit tests, I don't see unit tests for these; pls add

avelx (commented):
changes requested

PR #1407, opened 2022-09-05
MISUV-3945 Implement tests and improve coverage

andriikondiuk:


test/implicits/ImplicitJsonValidationFormatterSpec.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  "ImplicitJsonValidationFormatter.CurrencyFormatter" when {
+
+    "calling asString" should {
+
+      "return a properly formatted string" in {
+        testValidationErrors.asString shouldBe s", path: /someNode and errors: ValidationError(List($someValidationError1, $someValidationError2),WrappedArray(List($someArg1, $someArg2)))"
+      }
+
+      "return a properly formatted a strings when ValidationErrors are empty" in {
```
avelx, line 46: this should go under separate when statement, but it optional I would think

test/implicits/ImplicitJsonValidationFormatterSpec.scala
```diff
@@ -0,0 +1,55 @@
...
+        testEmptyValidationErrors.asString shouldBe ", path: /someNode and errors: "
+      }
+
+      "return an empty string when passing an empty sequence" in {
+        testEmptySequence.asString shouldBe ""
```
avelx, line 51: worth to reply on method to check is empty

avelx (approved):
minor comments, looks otherwise

PR #1408, opened 2022-09-05
MISUV-3929 : Improve test coverage for CitizenDetailsConnector.scala

AshokGitAccount:


avelx:
At this moment test coverage is 74.16 %, can we raise it to at least 85% ??

avelx (approved):
looks good otherwise

AshokGitAccount:
The invalid json parsing code block will never be covered because we are returning the empty object in case of invalid json as a valid response. 74.16 % is the maximum coverage we get

PR #1420, opened 2022-09-07
MISUV-3931 improve scoverage for CurrencyFormatter

sarinajsal:


andriikondiuk (approved):
Looks good to me

test/testUtils/CurrencyFormatterSpec.scala
```diff
@@ -36,6 +36,23 @@ class CurrencyFormatterSpec extends TestSupport with ImplicitCurrencyFormatter {
       }
     }
 
+    "CurrencyFormatterInt.toNegativeCurrency" should {
+      "return a value as negative" when {
+        "the number is more than 0" in {
+          val amount: Int = 1
```
avelx, line 42: * not sure why we use Int to model amount. maybe BigDecimal?, minor thing;
* to be 100% sure I would use property base testing to check if this conversion work for all variation of inputs(this is optional);

avelx (approved):
minor comment, looks good otherwise

test/testUtils/CurrencyFormatterSpec.scala
sarinajsal: We had two implicits for bigDecimal and Int, we've added tests for the Int and now have just pushed for the BigDecimal as well. You're right, property based tests are the best but for this case its should be sufficient (Andrii) Thank you!

sarinajsal: Pushed big decimal tests, is it ok if Andrew merges?

avelx: yeh, sure

avelx (approved):

PR #1431, opened 2022-09-13
MISUV-3974 : Redirect 'Credit from an earlier tax year' hyperlink in Claim a refund page

AshokGitAccount:


avelx:
you might need to get latest from the main before starting work / creating a new branch, as there some commit dated back like 15 days ago; otherwise good (not sure if there are must be any IT test changed ? )

avelx (approved):
looks good

PR #1438, opened 2022-09-16
MISUV-4034 Bug fix for add 'Money in your account' section to 'Credits summary' page

andriikondiuk:
Fix the issue that the Agent should gave this text Money in your client's account
Fix the issue when availableCredit is £0.00 then Money in your account block should not be rendered.

### DIFF
--- a/app/views/CreditsSummary.scala.html
+++ b/app/views/CreditsSummary.scala.html
@@ -86,6 +86,10 @@
     }
 }
 
+@getMoneyInYourAccountSectioName(isAgent: Boolean) = @{
+    if(isAgent) messages("credits.money-in-your-account-section.agent.name") else messages("credits.money-in-your-account-section.name")
+}
+
 @dropDownDetails(isAgent: Boolean) = {
     @h2(msg = "credits.drop-down-list.credit-from-hmrc-adjustment", classes = "govuk-heading-s", optId = Some("h2-credit-from-hmrc-adjustment"))
     @p(id = Some("p-credit-from-hmrc-adjustment")){@messages("credits.drop-down-list.credit-from-hmrc-adjustment.value")}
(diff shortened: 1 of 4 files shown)


app/views/CreditsSummary.scala.html
```diff
@@ -111,8 +115,8 @@
 
 @moneyInYourAccount(isAgent: Boolean) = {
     @maybeBalanceDetails.map { balanceDetails =>
-        @balanceDetails.availableCredit.map { availableCredit =>
-            @h2(msg = "credits.money-in-your-account-section.name", classes = "govuk-heading-s", optId = Some("h2-money-in-your-account"))
+        @balanceDetails.availableCredit.filter(_ > 0.00).map { availableCredit =>
```
avelx, line 118: we have a full control of what type we are passing to this view, why not to do this transformation on the controller level at least and pass appropriate type in the view and make it "lighter" ?

avelx (changes requested):
minore change requested

app/views/CreditsSummary.scala.html
andriikondiuk: To be honest I can add it to the CreditService.maybeBalanceDetails() ?

andriikondiuk: Hi, I've moved logic to the controller which simplified a view a bit. thanks

app/views/CreditsSummary.scala.html
```diff
@@ -86,6 +86,10 @@
     }
 }
 
+@getMoneyInYourAccountSectioName(isAgent: Boolean) = @{
```
avelx, line 89: isAgent is global variable, so it use is redundant in scope of this help function;

andriikondiuk: done thanks

avelx (approved):
looks good

PR #1441, opened 2022-09-21
MISUV-3978 Missing scenarios in Claim a refund page

andriikondiuk:


app/views/CreditAndRefunds.scala.html
```diff
@@ -82,162 +64,199 @@
     }
 }
 
+@creditAndRefundRefundProgressHtml(amount: BigDecimal) = {
+    @if(messages.lang.language.toLowerCase().equals("cy")) {
+        <li>
+            <p class="govuk-heading-s">
```
avelx, line 70: * not sure if this is really required to handle language specific scenarios separately;
* have a look at the example how next resource is handled, 'inYear-tax-overview.caption', you should be able to pass parameter in the resource etc;

andriikondiuk: I'm not touch this code at all

app/views/CreditAndRefunds.scala.html
```diff
@@ -82,162 +64,199 @@
...
+}
+
+@maybeBalancesDetails = @{
+    balance.getOrElse(throw MissingFieldException("Balance"))
+}
+
+@maybeAvailableCredit = @{
+    maybeBalancesDetails.availableCredit.getOrElse(throw MissingFieldException("Available Credit"))
+}
+
```
avelx, line 96: this maybe worth to move all these conditions/checks on the controller level?
I think adding some sort of mode variable of custom type/enum [None / UnallocatedCreditFromOnePayment/ isUnallocatedCreditFromSingleCreditItem] would work better;

avelx (changes requested):
changes requested

app/views/CreditAndRefunds.scala.html
avelx: ok, this can be covered by existing story raised: https://jira.tools.tax.service.gov.uk/browse/MISUV-4059

app/views/CreditAndRefunds.scala.html
andriikondiuk: Good idea thank you. I've just pushed the refactored code

app/views/CreditAndRefunds.scala.html
```diff
@@ -82,162 +65,191 @@
...
     isAgent = isAgent,
     useFallbackBackLink = true
 ) {
 
-
     @h1(msg = "credit-and-refund.heading")
 
     @if(balance.isDefined){
-        @if(balance.getOrElse(throw MissingFieldException("Balance")).availableCredit.getOrElse(throw MissingFieldException("Available Credit")) != 0){
+        @if(maybeUnallocatedCreditType(creditCharges, balance).contains(UnallocatedCreditFromOnePayment)) {
```
avelx, line 108: actual call to maybeUnallocatedCreditType can be done of the controller level;
I was also expecting simple pattern matching instead of using if/else statement, as we have nesting statements etc, this would be a hard to support/read;

test/utils/CreditAndRefundUtilsSpec.scala
```diff
@@ -0,0 +1,106 @@
...
+
+import org.scalatest.prop.TableDrivenPropertyChecks._
+import org.scalatest.prop.TableFor1
+import testConstants.CreditAndRefundConstants.{balanceDetailsModel, documentDetailWithDueDateFinancialDetailListModel}
+import testUtils.TestSupport
+import utils.CreditAndRefundUtils.UnallocatedCreditType.{UnallocatedCreditFromOnePayment, UnallocatedCreditFromSingleCreditItem, maybeUnallocatedCreditType}
+
+class CreditAndRefundUtilsSpec extends TestSupport {
+
+  val validMfaCreditDescriptions: TableFor1[String] = Table(
```
avelx, line 27: it is used in the context of mainType, confusing naming convension

app/views/CreditAndRefunds.scala.html
andriikondiuk: Thanks it make sense. Hi I've just refactored code and pushed the fix. thanks

app/controllers/CreditAndRefundController.scala
```diff
@@ -125,22 +131,21 @@ class CreditAndRefundController @Inject()(val authorisedFunctions: FrontendAutho
...
     }
     creditsGroupedPaymentTypes
   }
 
   def getCreditTypeGroupKey(credits: (DocumentDetailWithDueDate, FinancialDetail)): String = {
-    val isMFA : Boolean = credits._2.validMFACreditType()
-    val isCutOverCredit : Boolean = credits._2.mainType.get == "ITSA Cutover Credits"
-    val isPayment : Boolean = credits._1.documentDetail.paymentLot.isDefined
+    val isMFA: Boolean = credits._2.validMFACreditType()
+    val isCutOverCredit: Boolean = credits._2.mainType.get == "ITSA Cutover Credits"
```
avelx, line 142: I understand this is not your code, but we should use next method moving forward =>
mainType.validCutoverCreditType()

test/utils/CreditAndRefundUtilsSpec.scala
andriikondiuk: I propose this one: `val validMfaCreditMainTypeDescriptions: TableFor1[String] = Table(`

andriikondiuk: Done, thanks

app/controllers/CreditAndRefundController.scala
avelx: will be a part of another story

avelx (approved):

PR #1442, opened 2022-09-22
MISUV-3951: Update agent content in Tax year summary page

sarinajsal:


app/views/TaxYearSummary.scala.html
```diff
@@ -260,7 +260,10 @@ <h1 class="govuk-heading-xl" id="heading">
     @modelOpt.map { model =>
         @if(!model.crystallised.contains(true)) {
             @govukInsetText(InsetText(content = HtmlContent(
-                p(id=Some("calc-date-info"))(content = Html(messages("tax-year-summary.calc-from-last-time", model.timestamp.map(_.toZonedDateTime.toLocalDate.toLongDate).getOrElse(throw MissingFieldException("Calculation timestamp")))))
+                p(id=Some("calc-date-info"))(content = Html(
```
avelx, line 263: this change is correct according with the way we are approaching similar tasks; however, I thinking to changes timestamp field in the model to Option[LocalDate], so to avoid these transformations in the views

avelx (approved):
looks good

app/views/TaxYearSummary.scala.html
avelx: next story created => MISUV-4060

PR #1444, opened 2022-09-23
MISUV-3946: Update agent content in Home page

ayubMercator:


conf/messages
```diff
@@ -123,9 +124,10 @@ home.self-assessment.description                                = Use this servi
 home.credAndRefund.view                                         = Claim a refund
 home.credAndRefund_howToClaim.view                              = How to claim a refund
 
+home.agent.overdue.message.dunningLock.true                     = Your client has overdue payments and one or more of their tax decisions are being reviewed. They may be charged interest on these until they are paid in full.
```
avelx, line 127: might be better use "on" and "off" in this context;

ayubMercator: This would require mapping true -> on, and false -> off in the view. It will be simpler to stick with true/false

avelx (approved):
otherwise looks good

avelx (approved):
looks good[2]

PR #1448, opened 2022-09-26
Misuv-4054 Update content in What you owe page

andriikondiuk:


app/views/WhatYouOwe.scala.html
```diff
@@ -265,16 +265,32 @@
...
+
+@getPaymentsMadeBulletLinkText = @{
+    if(isAgent) messages("whatYouOwe.payments-made-bullet-agent-1.2") else messages("whatYouOwe.payments-made-bullet-1.2")
+}
+
+@getPaymentsMadeBulletText = @{
+    if(isAgent) messages("whatYouOwe.payments-made-bullet-agent-2") else messages("whatYouOwe.payments-made-bullet-2")
+}
+
+@getNoPaymentsDueText = @{
```
avelx, line 280: worth to consider to create an extension method something like: messagesIfElse(Boolean, Key1, Key2) as this pattern appears very often across project

avelx: story created => 
https://jira.tools.tax.service.gov.uk/browse/MISUV-4079

avelx (approved):
comments added, looks good otherwise

app/views/WhatYouOwe.scala.html
andriikondiuk: sounds good, thanks

PR #1453, opened 2022-09-28
MISUV-3946: Update agent content in Home page

ayubMercator:


avelx (approved):
looks good, but there are some ATs failing, need fixing before merging these changes

PR #1456, opened 2022-09-29
MISUV-4082 : Fix CreditAndRefunds view

AshokGitAccount:


app/views/CreditAndRefunds.scala.html
```diff
@@ -66,33 +67,25 @@
...
 }
 
 @maybeBalancesDetails = @{
-    balance.getOrElse(throw MissingFieldException("Balance"))
+balance.getOrElse(throw MissingFieldException("Balance"))
 }
 
 @maybeAvailableCredit = @{
-    maybeBalancesDetails.availableCredit.getOrElse(throw MissingFieldException("Available Credit"))
+    maybeBalancesDetails.availableCredit.getOrElse(BigDecimal(0))
```
avelx, line 88: assuming this is a part of requirement, just checking

app/views/CreditAndRefunds.scala.html
```diff
@@ -202,41 +195,32 @@ <h2 class="govuk-heading-s">@{
...
-                                        </p>
-                                        </li>
-                                }
+                                    </p>
+                                </li>
                             }
                         }
                         else {
-                            maybeBalancesDetails.secondPendingAmountRequested map { amount =>
+                            balance.get.secondPendingAmountRequested map { amount =>
```
avelx, line 210: pls use headOption instead, as this is not safe to call get directly / use MissingFieldException is required

avelx (changes requested):
minor changes requested

app/views/CreditAndRefunds.scala.html
AshokGitAccount: Yes, we want to trat None as 0

avelx (approved):
approved

PR #1459, opened 2022-09-30
Misuv 3983

sarinajsal:


app/views/CreditAndRefunds.scala.html
```diff
@@ -253,25 +253,37 @@ <h2 class="govuk-heading-m">
...
-                    @p(){
-                        @messages("credit-and-refund.bullet-one-prt-1")
-                        @link(link = getWhatYouOwe(isAgent),
-                            messageKey = "credit-and-refund.bullet-one-link",
-                            target = None)
-                        @messages("credit-and-refund.bullet-one-prt-2")
-                    }
-                </li>
+                @if(isAgent) {
+                    <li>
```
avelx, line 257: lines 256-276: the only difference is line 259, can we refactor this code and apply if/else statement in the line 259 instead ?

avelx (changes requested):
minor change requested, pls refactor to reduce number of lines in the view;

test/views/CreditAndRefundsViewSpec.scala
```diff
@@ -51,6 +51,12 @@ class CreditAndRefundsViewSpec extends TestSupport with FeatureSwitching with Im
   val creditAndRefundFromHMRCTitlePart1: String = messages("credit-and-refund.credit-from-hmrc-title-prt-1")
   val creditAndRefundFromHMRCTitlePart2: String = messages("credit-and-refund.credit-from-hmrc-title-prt-2")
   val creditAndRefundPaymentFromEarlierYearLinkText: String = messages("paymentHistory.paymentFromEarlierYear")
+  val creditAndRefundAgentHasCredits2: String = messages("credit-and-refund.agent.subHeading.has-credits-2")
```
avelx, line 54: if variable use only once, maybe we don't need it all ? you can call messages(..) directly where required etc

avelx (approved):
looks good

avelx:
ATs to be fixed before merging this PR

PR #1467, opened 2022-10-06
MISUV-4184: add optional taxYear parameter when calling tax calcs in BE

avelx:
As we are integrating 1885 within income-tax-calculation component, we need to have mandatory taxYear parameter provided every time we are calling corresponding API ;
Changes required in app/services/CalculationPollingService.scala service;


test/mocks/services/MockCalculationPollingService.scala
```diff
@@ -44,7 +44,9 @@ trait MockCalculationPollingService extends UnitSpec with MockitoSugar with Befo
       .initiateCalculationPollingSchedulerWithMongoLock(
         ArgumentMatchers.eq(calcId),
         ArgumentMatchers.eq(nino),
-        ArgumentMatchers.eq(mtditid)
+        //ArgumentMatchers.eq(taxYear),
```
andriikondiuk, line 47: The same as above, shall we put a TODO comment here?

andriikondiuk (approved):
Looks good to me

PR #1476, opened 2022-10-14
MISUV-3916_CL218 - End of Year forecast calculation to include allowances/deductions data item

SiddharthAtos:


app/views/ForecastTaxCalcSummary.scala.html
```diff
@@ -43,6 +43,7 @@ <h1 class="govuk-heading-xl">
     </h1>
 
     @dataItem("totalEstimatedIncome", endOfYearEstimateModel.totalEstimatedIncome.map(tei => BigDecimal(tei)))
+    @dataItem("totalAllowancesAndDeductions", Some(endOfYearEstimateModel.totalAllowancesAndDeductions.getOrElse(BigDecimal(0)) * -1))
```
edward-dennis-ey, line 46: Could this getOrElse be handled in the model? Also is Some needed?

edward-dennis-ey (approved):

PR #1487, opened 2022-10-26
MISUV-4289 : TimeMachine and Updated Tests

AshokGitAccount:


test/services/DateServiceSpec.scala
```diff
@@ -50,25 +51,10 @@ class DateServiceSpec extends TestSupport with FeatureSwitching {
     }
   }
 
-  "The getCurrentTaxYearEnd method when TimeMachineAddYear FS is on" should {
-    "return the next year if the current date is before april 6 and the timeMachineAddYears 1" in {
-      enable(TimeMachineAddYear)
-      TestDateService.getCurrentTaxYearEnd(LocalDate.parse("2018-03-29")) shouldBe 2019
+  "The getCurrentTaxYearEnd" should {
```
avelx, line 54: * I would need to rewire these unit test to use ~.getCurrentDate method;
* the best way test I would think to test getCurrentDate would be that if LocalDate.getYear return currentYear + 1 when timeMachine is on;

test/services/DateServiceSpec.scala
```diff
@@ -50,25 +51,15 @@ class DateServiceSpec extends TestSupport with FeatureSwitching {
...
+    "return the current tax year" in{
+      disable(TimeMachineAddYear)
+      TestDateService.getCurrentTaxYearEnd shouldBe 2023
     }
-    "return the following year if current date is on or after april 6 and the timeMachineAddYears 1" in {
+    "return next tax year when time machine is enabled" in {
       enable(TimeMachineAddYear)
-      TestDateService.getCurrentTaxYearEnd(LocalDate.parse("2018-04-06")) shouldBe 2020
+      TestDateService.getCurrentTaxYearEnd shouldBe 2024
```
avelx, line 61: this would not work next year, as mentioned in the prev comment, you would need to calculate year

PR #1512, opened 2022-11-15
MISUV 4394: WIP => Update API path for forecast field in Forecast tax calculation page

sarinajsal:


app/views/ForecastTaxCalcSummary.scala.html
```diff
@@ -34,6 +36,27 @@
     }
 }
 
+@dataItem2(message: String, value: Option[BigDecimal], value2: Option[BigDecimal], bold: Boolean = false) = {
+    @if(value2.isDefined && value.isDefined){
```
avelx, line 40: * looks like a good case to apply pattern matching on 'value' and 'value2', 
* might be good idea to rename value2 to something more easy to understand/read, as you have assume code would be maintained in the future etc;

avelx (approved):
looks ok, approved with comments

PR #1513, opened 2022-11-15
MISUV-3215:  Use DateService in preference to LocalDate.now

ayubAz:


app/views/WhatYouOwe.scala.html
```diff
@@ -17,16 +17,17 @@
...
 @import models.financialDetails._
 @import auth.MtdItUser
 @import views.html.layouts.unifiedLayout
 @import views.html.components._
 @import exceptions.MissingFieldException
 @import uk.gov.hmrc.govukfrontend.views.html.components._
+@import services.DateService
 
 @this(
+    dateService: DateService,
```
avelx, line 30: I would think we don't have to expose services in the views; so on this occasion, this would be more than enough to pass current date which you will get by calling relevant dateService method in the controller

app/views/ChargeSummary.scala.html
```diff
@@ -57,7 +59,7 @@
 
 @dueDate = @{documentDetailWithDueDate.dueDate}
 
-@currentDate = @{LocalDate.now}
+@currentDate = @{dateService.getCurrentDate}
```
avelx, line 62: again, no need to pass service in the view, pass the actual current date from the controller level

app/services/WhatYouOweService.scala
```diff
@@ -76,7 +77,7 @@ class WhatYouOweService @Inject()(val financialDetailsService: FinancialDetailsS
 
         val whatYouOweChargesList = WhatYouOweChargesList(
           balanceDetails = balanceDetails,
-          chargesList = getFilteredChargesList(financialDetailsModelList),
+          chargesList = getFilteredChargesList(financialDetailsModelList)(dateService),
```
avelx, line 80: no need to pass implicit dateService on this occasion, as dateService in a global value within this class as you can specify this it directly  on line 106 when you calling getAllDocumentDetailsWithDueDates

app/services/WhatYouOweService.scala
```diff
@@ -100,7 +101,7 @@ class WhatYouOweService @Inject()(val financialDetailsService: FinancialDetailsS
     }
   }
 
-  private def getFilteredChargesList(financialDetailsList: List[FinancialDetailsModel]): List[DocumentDetailWithDueDate] = {
+  private def getFilteredChargesList(financialDetailsList: List[FinancialDetailsModel])(implicit dateService: DateService): List[DocumentDetailWithDueDate] = {
```
avelx, line 104: no need to pass implicit dateService ~

app/models/nextUpdates/NextUpdatesResponseModel.scala
```diff
@@ -72,9 +72,8 @@ case class NextUpdateModel(start: LocalDate,
                            dateReceived: Option[LocalDate],
                            periodKey: String) extends NextUpdatesResponseModel {
 
-  def currentTime(): LocalDate = LocalDate.now()
-
-  def getNextUpdateStatus: NextUpdateStatus = if (!currentTime().isAfter(due)) Open(due) else Overdue(due)
+  def currentTime(implicit dateService: DateService): LocalDate = dateService.getCurrentDate
```
avelx, line 75: I would think that currentTime is redundant, so you can remove it

app/models/nextUpdates/NextUpdatesResponseModel.scala
```diff
@@ -72,9 +72,8 @@ case class NextUpdateModel(start: LocalDate,
...
+  def getNextUpdateStatus(implicit dateService: DateService): NextUpdateStatus = if (!currentTime(dateService).isAfter(due)) Open(due) else Overdue(due)
```
avelx, line 76: this method is only used in the scope of unit tests, so we can move this logic in the helpers logic in the scope of unit tests;

avelx (changes requested):
just few minor comment, pls address these, otherwise looks good

avelx (approved):

PR #1518, opened 2022-11-18
MISUV-4456-last-update Fix for the Date field which now should be from the documentDate for MFA credit but Credit Summary page date should be a taxYear

andriikondiuk:


app/services/CreditHistoryService.scala
```diff
@@ -63,18 +65,24 @@ class CreditHistoryService @Inject()(incomeTaxViewChangeConnector: IncomeTaxView
     }
   }
 
+  def getTaxYearAsInt(taxYear: String): Int =
```
avelx, line 68: method to be remove during https://jira.tools.tax.service.gov.uk/browse/MISUV-4508 implementation

test/views/CreditsSummaryViewSpec.scala
```diff
@@ -40,7 +40,7 @@ class CreditsSummaryViewSpec extends TestSupport with FeatureSwitching with Impl
   lazy val mockAppConfig: FrontendAppConfig = app.injector.instanceOf[FrontendAppConfig]
 
   val testCalendarYear: Int = 2018
-  val expectedDate: String = "2018"
+  val expectedDate: String = "29 Mar 2018"
```
avelx, line 43: still not convinced that we should use String instead of LocalDate, as this is really representing moment in time

PR #1542, opened 2022-12-06
MISUV-4557: fix iv redirect

ayubAz:


conf/application.conf
```diff
@@ -302,6 +302,10 @@ time-machine {
     add-years = 1
 }
 
+iv {
```
avelx, line 305: there is existing identity-verification-frontend key in the config, we might don't need add anything new on this occasion; just re-use what we have so far

test/controllers/predicates/AuthenticationPredicateSpec.scala
```diff
@@ -174,5 +174,16 @@ class AuthenticationPredicateSpec extends TestSupport with MockAuthenticationPre
       }
 
     }
+
+    "called with a confidence level of 250" should {
+      "return Ok (200)" in {
+        enable(IvUplift)
```
avelx, line 180: I would assume this is not required ? so, ie. if IvUplift is not enabled we still should be able to login with CL250 (np);
but in case we are trying with CL200 and IvUplift is on we are redirected (so we are testing redirect etc)

avelx (changes requested):
minor changes requested

avelx (approved):
approve

PR #1543, opened 2022-12-07
MISUV-4575 : Prevent Negative Credit Amount Showing

AshokGitAccount:


app/models/financialDetails/BalanceDetails.scala
```diff
@@ -31,6 +33,15 @@ case class BalanceDetails(balanceDueWithin30Days: BigDecimal,
   val total: Option[BigDecimal] = availableCredit.map{ total =>
     total - firstPendingAmountRequested.getOrElse(0) - secondPendingAmountRequested.getOrElse(0)
   }
+
```
avelx, line 36: pls unit test this change

avelx (changes requested):
pls add required unit tests

app/models/financialDetails/BalanceDetails.scala
AshokGitAccount: Added Unit tests

avelx (approved):
approved

test/models/financialDetails/BalanceDetailsSpec.scala
```diff
@@ -95,5 +95,15 @@ class BalanceDetailsSpec extends UnitSpec {
         balanceDetailsModel().refundInProgress shouldBe false
       }
     }
+
+    "return absolute values" when {
```
avelx, line 99: we would need a unit test for the case where is no value;

PR #1563, opened 2022-12-30
MISUV-4701 : FE: Request/View refund: Redesign

AshokGitAccount:


app/controllers/CreditAndRefundController.scala
```diff
@@ -189,11 +189,9 @@ class CreditAndRefundController @Inject()(val authorisedFunctions: FrontendAutho
       case financialDetailsModel: List[FinancialDetailsModel] =>
         val balance: Option[BalanceDetails] = financialDetailsModel.headOption.map(balance => balance.balanceDetails)
         repaymentService.start(user.nino, balance.flatMap(_.getAbsoluteAvailableCreditAmount).getOrElse(BigDecimal(0))).flatMap {
-          case RepaymentJourneyModel(nextUrl) =>
+          case Right(nextUrl) =>
```
avelx, line 192: pls move logic above into service layer

app/services/RepaymentService.scala
```diff
@@ -17,27 +17,39 @@
...
+import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 
 @Singleton
-class RepaymentService @Inject()(val repaymentConnector: RepaymentConnector){
+class RepaymentService @Inject()(val repaymentConnector: RepaymentConnector) {
 
   def start(nino: String, fullAmount: BigDecimal)
-           (implicit headerCarrier: HeaderCarrier): Future[RepaymentJourneyResponseModel] = {
+           (implicit headerCarrier: HeaderCarrier): Future[Either[Throwable, String]] = {
```
avelx, line 32: you might pass fullAmount as Option[BigDecimal] here and do extract absolute value, for None value just raise an exception

app/services/RepaymentService.scala
```diff
@@ -17,27 +17,39 @@
...
     Logger("application").debug(s"Repayment journey start with nino: $nino and fullAmount: $fullAmount ")
-    repaymentConnector.start(nino, fullAmount)
+    repaymentConnector.start(nino, fullAmount).map {
+      case RepaymentJourneyModel(nextUrl) =>
+        Right(nextUrl)
+      case RepaymentJourneyErrorResponse(status, message) =>
+        Logger("application").error(s"Repayment journey start error with response code: $status and message: $message}")
```
avelx, line 38: pls apply formatting for the log entries, i.e => [CalculationService][getLatestCalculation] ,

app/services/RepaymentService.scala
```diff
@@ -17,27 +17,39 @@
...
+        Left(new InternalError)
+    }.recover { case ex: Exception => Left(ex) }
```
avelx, line 40: we need to have log error entry within recover so we can identify this during investigation etc

app/services/RepaymentService.scala
AshokGitAccount: done

app/controllers/CreditAndRefundController.scala
AshokGitAccount: done

app/services/RepaymentService.scala
AshokGitAccount: done

app/services/RepaymentService.scala
AshokGitAccount: done

avelx (approved):
approved

PR #1578, opened 2023-01-16
MISUV-4772 : Repayments enable logic by FS

AshokGitAccount:


avelx:
@AshokGitAccount  - please double check story and make sure you understand it correctly:
https://jira.tools.tax.service.gov.uk/browse/MISUV-4755

AshokGitAccount:
Changes made and added test case

app/controllers/CreditAndRefundController.scala
```diff
@@ -202,6 +209,8 @@ class CreditAndRefundController @Inject()(val authorisedFunctions: FrontendAutho
   private def handleStatusRefundRequest(isAgent: Boolean, itvcErrorHandler: ShowInternalServerError, backUrl: String)
                                  (implicit user: MtdItUser[_], hc: HeaderCarrier, ec: ExecutionContext, messages: Messages): Future[Result] = {
     repaymentService.view(user.nino).flatMap {
+      case _ if isDisabled(CreditsRefundsRepay) =>
```
avelx, line 212: we really should check is FS before we are calling repayment service

avelx, line 212: we really should check is FS before we are calling repayment service

avelx (approved):
looks ok

PR #1585, opened 2023-01-18
Undo revert to restore MISUV-4760

ayubAz:


avelx (approved):
looks good

avelx (approved):
confirmed by @tharun.prasath that these are changes expected

PR #1587, opened 2023-01-19
MISUV-4736: CL257 - Update error messages for Agents in Tax year summary page

AshokGitAccount:


avelx:
@AshokGitAccount  - I would believe you would need a simple tests to verify that all resources you added under C15~ exists; just a simple test to make sure key/value pair exists for Eng/Cy would work fine;

conf/messages
```diff
@@ -272,6 +272,76 @@ tax-year-summary.message.C559105                                   = the amount
...
+tax-year-summary.agent.message.C15015                              = your client has claimed to carry forward a loss to set against general income of the next year. They also need to make the claim in the same year the loss arose.
+tax-year-summary.agent.message.C15016                              = your client uses cash basis accounting. This means that they cannot claim to set losses against other taxable income.
+tax-year-summary.agent.message.C15102                              = the total amount of one-off Gift Aid payments is more than the total Gift Aid payments your client made.
+tax-year-summary.agent.message.C15103                              = Gift Aid payments made this year treated as paid in the previous year are more than the total Gift Aid payments your client made.
+tax-year-summary.agent.message.C15104                              = the value of qualifying investments your client gifted to non-UK charities are more than the value of these items they gifted to charity:<br><ul class="govuk-list govuk-list--bullet"> <li> qualifying shares and securities </li> <li> qualifying land and buildings </li> </ul>
```
avelx, line 283: this might be not a good idea to keep HTML formatting as a part of resource file, we have views for this

AshokGitAccount: It's been implemented for the individual so I have implemented the same for Agent

avelx: no html and styles in the message files pls

app/controllers/agent/errors/asd.sc
```diff
@@ -0,0 +1,11 @@
+import scala.util.matching.Regex
```
avelx, line 1: Pls drop this file from the PR

avelx: Why do we need this regexp btw ???

app/views/TaxYearSummary.scala.html
```diff
@@ -48,17 +48,24 @@
...
-        @errorMessage(model)
+@getMessageKey(isAgent: Boolean) = {
+  @if(isAgent) {tax-year-summary.agent.message} else {tax-year-summary.message}
+}
+
+@taxCalculationOverview(model: TaxYearSummaryViewModel, isAgent: Boolean) = {
+    @if(model.messages.exists(_.errorMessages.size > 0)) {
+       @{
+            val key = getMessageKey(isAgent)
+            errorMessage(model,s"${key}".trim(), isAgent)
```
avelx, line 67: you can move trim in getMessageKey function,

AshokGitAccount: I've removed the function and used a function literal instead. We dont need the trim() now.

app/controllers/agent/errors/asd.sc
AshokGitAccount: Deleted file

avelx (approved):
approved

PR #1591, opened 2023-01-24
MISUV-4607 : Fixed Table in Tax Year Summary Page

AshokGitAccount:


app/views/TaxYearSummary.scala.html
```diff
@@ -368,6 +368,7 @@ <h1 class="govuk-heading-xl" id="heading">
                 )
             )))
         }
+        @if(!errorPresent(model)) {
```
avelx, line 371: @AshokGitAccount  - I would assume this was not part of the IT test originally (was not covered ) before we started working on this story?

avelx: can you add IT test to cover this scenario ?

AshokGitAccount: @avelx - added check for forecast table to not be displayed if there are error messages in the existing IT

avelx: @AshokGitAccount - do we need to update spec for agent as well?

PR #1598, opened 2023-01-27
MISUV-4829: Claim Refund page - disable request/view refund logic for Agents

ayubAz:


avelx:
minor changes:
if you can mark cases related to individual, so we can see clear distinction on these;
if you can rename/create a new instance of tests data for the individual cases where we have 'testAgentAuthRetrievalSuccess', this is expected to be named for individual, you would have to change AffinityGroup set as Individual for this data;

PR #1609, opened 2023-02-06
MISUV-4898: Code refactoring MtdItUserBase

ayubAz:


app/audit/Utilities.scala
```diff
@@ -33,8 +35,8 @@ object Utilities {
     ("credId", user.credId) ++
     userType(user.userType)
 
-  def userType(userType: Option[String]): JsObject = userType match {
-    case Some("Agent") => Json.obj("userType" -> "Agent")
+  def userType(userType: Option[AffinityGroup]): JsObject = userType match {
+    case Some(Agent) => Json.obj("userType" -> "Agent")
```
avelx, line 39: we might need to have explicit case for Individual, as we also have Organisation's

PR #1633, opened 2023-03-06
MISUV 4414: Investigate: AuditServiceSpec re-enable disabled unit test

danielmaltz1:
Removed obsolete test and moved private logic into public methods for better logging


avelx (approved):
looks good

PR #1642, opened 2023-03-14
Misuv 4834 - WhatYouOweControllerISpec(s): remove usage of LocalDate.now

danielmaltz1:
Removed all instances of LocalDate.now from the two WhatYouOweControllerISpecs, replaced with a static date. All changes to FinancialDetailsIntegrationTestConstants.scala only affect constants which are only used by WhatYouOweControllerISpec(s)

### DIFF
--- a/it/controllers/WhatYouOweControllerISpec.scala
+++ b/it/controllers/WhatYouOweControllerISpec.scala
@@ -46,12 +46,13 @@ class WhatYouOweControllerISpec extends ComponentSpecBase {
   )(FakeRequest())
 
   val testTaxYear: Int = getCurrentTaxYearEnd.getYear - 1
+  val testDate: LocalDate = LocalDate.parse("2022-01-01")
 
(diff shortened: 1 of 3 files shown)


it/controllers/WhatYouOweControllerISpec.scala
```diff
@@ -46,6 +48,26 @@ class WhatYouOweControllerISpec extends ComponentSpecBase {
   )(FakeRequest())
 
   val testTaxYear: Int = getCurrentTaxYearEnd.getYear - 1
+  val testDate: LocalDate = LocalDate.parse("2022-01-01")
```
avelx, line 51: this is fine, but ideally its better to use https://docs.oracle.com/javase/8/docs/api/java/time/LocalDate.html
LocalDate.of 
reason: that the approach used above we using DateFormatting behind the scene and it not guaranteed that exact format YYYY-MM-DD will be used or for example YYYY-DD-YY; but I think should be fine in this instance

avelx (approved):
approved

PR #1654, opened 2023-03-20
Misuv 5128: FE: refactor FinancialDetail : extract getCreditType method

danielmaltz1:
Replaced old Boolean methods with single getCreditType: Option[String] method. Replaced all calls to old methods with call to new method. Added tests for new method

### DIFF
--- a/app/audit/models/ClaimARefundAuditModel.scala
+++ b/app/audit/models/ClaimARefundAuditModel.scala
@@ -54,9 +54,6 @@ case class ClaimARefundAuditModel(balanceDetails: Option[BalanceDetails],
       case None =>
         Logger("application").error(s"[ClaimARefundAuditModel][getCreditType] Missing or non-matching credit: not a valid credit type")
         "unknownCredit"
-      case error =>
-        Logger("application").error(s"[ClaimARefundAuditModel][getCreditType] Missing or non-matching credit: $error found")
-        "unknownCredit"
     }
   }
 
--- a/app/audit/models/ClaimARefundAuditModel.scala
+++ b/app/audit/models/ClaimARefundAuditModel.scala
@@ -43,17 +43,16 @@ case class ClaimARefundAuditModel(balanceDetails: Option[BalanceDetails],
   }
 
   private def getCreditType(credit: (DocumentDetailWithDueDate, FinancialDetail)): String = {
-    val isMFA: Boolean = credit._2.isMFACredit
-    val isCutOverCredit: Boolean = credit._2.isCutOverCredit
+    val creditType: Option[String] = credit._2.getCreditType
     val isPayment: Boolean = credit._1.documentDetail.paymentLot.isDefined
-    val isBalancingCreditCharge: Boolean = credit._2.isBalancingChargeCredit
-
-    (isMFA, isCutOverCredit, isBalancingCreditCharge, isPayment) match {
-      case (true, _, _, _) => "Credit from HMRC adjustment"
-      case (_, true, _, _) => "Credit from an earlier tax year"
-      case (_, _, true, _) => "Balancing charge credit"
-      case (_, _, _, true) => s"Payment made on ${getFullDueDate(credit._1.dueDate.get)}"
-
+    (creditType, isPayment) match {
+      case (Some("MFA Credit"), _) => "Credit from HMRC adjustment"
+      case (Some("Cutover Credit"), _) => "Credit from an earlier tax year"
+      case (Some("Balancing Charge Credit"), _) => "Balancing charge credit"
+      case (_, true) => s"Payment made on ${getFullDueDate(credit._1.dueDate.get)}"
+      case (None, _) =>
+        Logger("application").error(s"[ClaimARefundAuditModel][getCreditType] Missing or non-matching credit: not a valid credit type")
+        "unknownCredit"
       case error =>
         Logger("application").error(s"[ClaimARefundAuditModel][getCreditType] Missing or non-matching credit: $error found")
         "unknownCredit"
--- a/app/audit/models/ClaimARefundAuditModel.scala
+++ b/app/audit/models/ClaimARefundAuditModel.scala
@@ -20,6 +20,7 @@ import audit.Utilities.userAuditDetails
 import auth.MtdItUser
 import enums.AuditType.ClaimARefundResponse
 import enums.TransactionName.ClaimARefund
+import models.creditDetailModel.{BalancingChargeCreditType, CreditType, CutOverCreditType, MfaCreditType}
 import models.financialDetails.{BalanceDetails, DocumentDetailWithDueDate, FinancialDetail}
 import play.api.Logger
 import play.api.libs.json.{JsObject, JsValue, Json}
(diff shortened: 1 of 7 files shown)


app/audit/models/ClaimARefundAuditModel.scala
```diff
@@ -43,17 +43,16 @@ case class ClaimARefundAuditModel(balanceDetails: Option[BalanceDetails],
...
     val isPayment: Boolean = credit._1.documentDetail.paymentLot.isDefined
-    val isBalancingCreditCharge: Boolean = credit._2.isBalancingChargeCredit
-
-    (isMFA, isCutOverCredit, isBalancingCreditCharge, isPayment) match {
-      case (true, _, _, _) => "Credit from HMRC adjustment"
-      case (_, true, _, _) => "Credit from an earlier tax year"
-      case (_, _, true, _) => "Balancing charge credit"
-      case (_, _, _, true) => s"Payment made on ${getFullDueDate(credit._1.dueDate.get)}"
-
+    (creditType, isPayment) match {
```
avelx, line 48: I would really think that isPayment ise in this pattern matching not required, as its really used once

app/models/financialDetails/FinancialDetail.scala
```diff
@@ -88,11 +88,16 @@ case class FinancialDetail(taxYear: String,
     }
     .filter(_.getPaymentAllocationTextInChargeSummary.isDefined)
 
-  def isMFACredit: Boolean = MfaCreditUtils.validMFACreditType(mainType)
-
-  def isCutOverCredit: Boolean = mainType.contains("ITSA Cutover Credits")
-
-  def isBalancingChargeCredit: Boolean = mainType.contains("SA Balancing Charge Credit")
+  def getCreditType: Option[String] = {
```
avelx, line 91: pls use strong typing instead of Option[String] as this is suggested in the main story

app/views/CreditAndRefunds.scala.html
```diff
@@ -89,21 +89,21 @@
 }
 
 @getCreditsSummaryLink = {
-    @if(maybeCreditChargesHead._2.isMFACredit) {
+    @if(maybeCreditChargesHead._2.getCreditType == Some("MFA Credit")) {
```
avelx, line 92: these checks should be localised on the different level, so we already should check against strong types here

app/audit/models/ClaimARefundAuditModel.scala
```diff
@@ -43,17 +44,16 @@ case class ClaimARefundAuditModel(balanceDetails: Option[BalanceDetails],
...
-
+    creditType match {
+      case Some(MfaCreditType) => "Credit from HMRC adjustment"
+      case Some(CutOverCreditType) => "Credit from an earlier tax year"
+      case Some(BalancingChargeCreditType) => "Balancing charge credit"
+      case _ if isPayment => s"Payment made on ${getFullDueDate(credit._1.dueDate.get)}"
+      case None =>
+        Logger("application").error(s"[ClaimARefundAuditModel][getCreditType] Missing or non-matching credit: not a valid credit type")
+        "unknownCredit"
       case error =>
```
avelx, line 57: error line become redundant after this change, so you are free to drop it as input would be alway of type => Option[CreditType]

app/models/financialDetails/FinancialDetail.scala
```diff
@@ -88,11 +89,16 @@ case class FinancialDetail(taxYear: String,
...
-  def isMFACredit: Boolean = MfaCreditUtils.validMFACreditType(mainType)
-
-  def isCutOverCredit: Boolean = mainType.contains("ITSA Cutover Credits")
-
-  def isBalancingChargeCredit: Boolean = mainType.contains("SA Balancing Charge Credit")
+  def getCreditType: Option[CreditType] = {
+    val validMFA = MfaCreditUtils.validMFACreditType(mainType)
+    (validMFA, mainType) match {
+      case (true, _) => Some(MfaCreditType)
+      case (_, Some(credit)) => if (credit.contains("ITSA Cutover Credits")) Some(CutOverCreditType) else {
```
avelx, line 96: you re-write
`case (_, Some(credit)) => if (credit.contains("ITSA Cutover Credits")) Some(CutOverCreditType)`
into:
`case (_, Some(""ITSA Cutover Credits"")) => Some(CutOverCreditType)`

danielmaltz1: I wasn't sure if the mainType would be exactly "ITSA Cutover Credits" or whether it could have extra text and so we would need to do .contains. But if it will always be "ITSA Cutover Credits" and never "...ITSA Cutover Credits..." I will change it

avelx: you can check ATs project for exact data would be in this field; my assumption is that this would be exact match, but you can apply something like: ~.map(_.trim) to drop any extra space; but this is very unlikely

danielmaltz1: Where in the ATs would I found that?

app/audit/models/ClaimARefundAuditModel.scala
```diff
@@ -43,19 +44,15 @@ case class ClaimARefundAuditModel(balanceDetails: Option[BalanceDetails],
...
-      case (_, _, true, _) => "Balancing charge credit"
-      case (_, _, _, true) => s"Payment made on ${getFullDueDate(credit._1.dueDate.get)}"
-
-      case error =>
-        Logger("application").error(s"[ClaimARefundAuditModel][getCreditType] Missing or non-matching credit: $error found")
+    creditType match {
+      case Some(MfaCreditType) => "Credit from HMRC adjustment"
+      case Some(CutOverCreditType) => "Credit from an earlier tax year"
+      case Some(BalancingChargeCreditType) => "Balancing charge credit"
+      case _ if isPayment => s"Payment made on ${getFullDueDate(credit._1.dueDate.get)}"
```
avelx, line 53: its not part of you changes, but we should really avoid calling .get directly, might be better to call something like .getOrElse(LocaDate.Min), so you would return 'beginning of time' instead , and if data is wrong we not fail this part;

app/services/CreditHistoryService.scala
```diff
@@ -45,16 +45,16 @@ class CreditHistoryService @Inject()(incomeTaxViewChangeConnector: IncomeTaxView
...
-            (financialDetail.isMFACredit, document.credit.isDefined, financialDetail.isBalancingChargeCredit) match {
-              case (true, true, false) =>
+            (financialDetail.getCreditType, document.credit.isDefined) match {
+              case (Some(MfaCreditType), true) =>
                 Some(CreditDetailModel(date = document.documentDate, document, MfaCreditType, Some(financialDetailsModel.balanceDetails)))
-              case (false, true, true) =>
+              case (Some(BalancingChargeCreditType), true) =>
                 Some(CreditDetailModel(date = document.documentDate, document, BalancingChargeCreditType, Some(financialDetailsModel.balanceDetails)))
-              case (false, true, false) =>
+              case (_, true) =>
```
avelx, line 53: I think we can put in this case:
`case ( Some(CutOverCreditType), true) => ...`

PR #1657, opened 2023-03-24
Misuv 5180: Investigate: FE - remove FS use from the services level

danielmaltz1:
Removed FS from 3 out of 4 services using it. FS checks are now executed in Controllers/Tests and the variable passed as a parameter into the service. Removing TimeMachine check from DateService was not possible


avelx (approved):
approved

PR #1676, opened 2023-04-12
MISUV-5335 TxM event - CL218 (R8b) -'Forecast tax calculation ' page

edward-dennis-ey:


it/controllers/ForecastTaxCalcSummaryControllerISpec.scala
```diff
@@ -1,18 +1,75 @@
...
+import play.api.test.FakeRequest
+import testConstants.BaseIntegrationTestConstants.{testMtditid, testNino, testYear}
+import testConstants.NewCalcBreakdownItTestConstants.liabilityCalculationModelSuccessful
+import uk.gov.hmrc.auth.core.AffinityGroup.Individual
+
+object ForecastTaxSummaryControllerTestConstants {
+  val mtdItUser: MtdItUserWithNino[_] = MtdItUserWithNino(testMtditid, testNino, None, None, Some("1234567890"),
+    Some("12345-credId"), Some(Individual), None)(FakeRequest())
+
+  val taxableIncome = 12500
```
avelx, line 20: I would assume this should be an amount which should be BigDecimal ?

app/audit/models/ForecastTaxCalculationAuditModel.scala
```diff
@@ -0,0 +1,79 @@
...
+import utils.Utilities._
+
+
+case class ForecastTaxCalculationAuditModel(user: MtdItUserWithNino[_], endOfYearEstimate: EndOfYearEstimate)
+  extends ExtendedAuditModel with ImplicitDateParser {
+
+  override val transactionName: String = enums.TransactionName.ForecastTaxCalculation
+  override val auditType: String = enums.AuditType.ForecastTaxCalculation
+
+  private val totalEstimatedIncome: Option[Int] = endOfYearEstimate.totalEstimatedIncome
```
avelx, line 33: not sure if we should use Option[Int] here to represent amount

it/controllers/ForecastTaxCalcSummaryControllerISpec.scala
edward-dennis-ey: This is an Int in the IncomeSource case class so I think this is okay?

app/audit/models/ForecastTaxCalculationAuditModel.scala
avelx: clarification required from the BAs level regarding format for each amount we store in this event; I've asked @Olga to clarify requirements for this story;

avelx (approved):
approved

PR #1682, opened 2023-04-19
MISUV-5370: Incorrect date format getting displayed on coding out summary page

simondrugan16:
added .toString so that the ints aren't formatted incorrectly


avelx (approved):
approved

test/views/ChargeSummaryViewSpec.scala
```diff
@@ -95,6 +95,7 @@ class ChargeSummaryViewSpec extends ViewSpec with FeatureSwitching{
     val balancingCharge: String = messages("chargeSummary.balancingCharge.text")
     val paymentBreakdownNic2: String = messages("chargeSummary.paymentBreakdown.nic2")
     val codingOutMessage2017To2018: String = messages("chargeSummary.codingOutMessage", 2017, 2018)
+    val codingOutMessage2017To2018WithStringMessagesArgument: String = messages("chargeSummary.codingOutMessage", "2017", "2018")
```
avelx, line 98: I might suspect we can change existing method instead, but looks good anyway

PR #1685, opened 2023-04-21
MISUV-5399 SCP025 | Update Homepage with new 'Income sources' tile

edward-dennis-ey:


app/models/incomeSourceDetails/IncomeSourceDetailsResponse.scala
```diff
@@ -31,6 +31,8 @@ case class IncomeSourceDetailsModel(mtdbsa: String,
 
   val hasPropertyIncome: Boolean = property.nonEmpty
   val hasBusinessIncome: Boolean = businesses.nonEmpty
+  val hasOngoingBusinessOrPropertyIncome: Boolean = businesses.exists(b => b.cessation.forall(_.reason.isEmpty)) ||
```
joesheridan, line 34: as discussed, can we use cessationDate here

PR #1687, opened 2023-04-24
MISUV-5046 Update IncomeSourceDetailsService

danielmaltz1:


app/models/incomeSourceDetails/PropertyDetailsModel.scala
```diff
@@ -22,7 +22,9 @@ import play.api.libs.json.{Json, OFormat}
 
 case class PropertyDetailsModel(incomeSourceId: Option[String],
                                 accountingPeriod: Option[AccountingPeriodModel],
-                                firstAccountingPeriodEndDate: Option[LocalDate])
+                                firstAccountingPeriodEndDate: Option[LocalDate],
+                                incomeSourceType: Option[String],
```
avelx, line 26: * please add enumeration representing set of incomeSourceTypes and use instead of Option[String];
* tradingStartDate - must be Option[LocalDate] assuming on the naming convention used for the field;

app/models/incomeSourceDetails/BusinessDetailsModel.scala
```diff
@@ -23,7 +23,9 @@ import play.api.libs.json.{Format, Json}
 case class BusinessDetailsModel(incomeSourceId: Option[String],
                                 accountingPeriod: Option[AccountingPeriodModel],
                                 tradingName: Option[String],
-                                firstAccountingPeriodEndDate: Option[LocalDate])
+                                firstAccountingPeriodEndDate: Option[LocalDate],
+                                tradingStartDate: Option[String],
```
avelx, line 27: pls use Option[LocalDate] here as well

avelx (approved):
approved

PR #1696, opened 2023-05-02
MISUV-5405: - Add new SE & property income 

ayubAz:


avelx:
@ayubAz  - pls use AddIncomeSource naming convention across this change;

app/views/NewIncomeSources.scala.html
```diff
@@ -0,0 +1,131 @@
...
+    h1: h1,
+    h2: h2,
+    p: p,
+    link: link,
+    hr: hr,
+    govukTable: GovukTable
+)
+
+@(
+    soleTraderBusinesses: List[BusinessDetailsModel],
```
avelx, line 37: * pls create a required view model and wrap this into these field in it;
* as originally discussed, we expect a method in incomeSourceDetailsService to return this model;

app/controllers/AddIncomeSourceController.scala
```diff
@@ -16,31 +16,72 @@
...
-                                          mcc: MessagesControllerComponents,
-                                          val ec: ExecutionContext,
-                                          val itvcErrorHandlerAgent: AgentItvcErrorHandler)
-  extends ClientConfirmedController with FeatureSwitching with I18nSupport {
+  def show(): Action[AnyContent] = (checkSessionTimeout andThen authenticate andThen retrieveNino
+    andThen retrieveIncomeSources andThen retrieveBtaNavBar).async {
+    implicit user =>
+      handleRequest(
+        sources = user.incomeSources,
+        isAgent = false,
```
avelx, line 55: really an optional, but I would keen to read isAgent state from user.userType, we can have some extension method on this type to distinguish this, but we might have to create a story to address rest of the page as well

app/services/IncomeSourceDetailsService.scala
```diff
@@ -50,23 +51,44 @@ class IncomeSourceDetailsService @Inject()(val incomeTaxViewChangeConnector: Inc
...
+//            case error: IncomeSourceDetailsResponse => Future.successful(error)
+//          }
+//      }
+//    } else {
       incomeTaxViewChangeConnector.getIncomeSources()
-    }
+//    }
+  }
+
+  def incomeSourcesAsViewModel(sources: IncomeSourceDetailsModel): IncomeSourcesViewModel = {
```
avelx, line 74: this view model would be specific to the single view, so you can use mode specific naming conversion in method and view model name

app/services/IncomeSourceDetailsService.scala
```diff
@@ -50,23 +51,44 @@ class IncomeSourceDetailsService @Inject()(val incomeTaxViewChangeConnector: Inc
...
+    IncomeSourcesViewModel(
+      soleTraderBusinesses = sources.businesses.filterNot(_.isCeased).map {
+        case b if b.tradingName.nonEmpty && b.tradingStartDate.nonEmpty =>
```
avelx, line 77: instead of using variable name p / b - use full names, property and business name accordingly, this would simplify reading/maintaining this code in the future

conf/app.routes
```diff
@@ -150,9 +150,15 @@ GET          /how-to-claim-refund-external                                contro
...
 
 # Income Sources
-GET          /income-sources/manage/choose-income-sources                 controllers.ManageIncomeSourceController.show()
-GET          /agents/income-sources/manage/choose-income-sources          controllers.ManageIncomeSourceController.showAgent()
-GET          /income-sources/add/new-income-sources                       controllers.AddIncomeSourceController.show()
-GET          /agents/income-sources/add/new-income-sources                controllers.AddIncomeSourceController.showAgent()
-GET          /income-sources/cease/cease-an-income-source                 controllers.CeaseIncomeSourceController.show()
-GET          /agents/income-sources/cease/cease-an-income-source          controllers.CeaseIncomeSourceController.showAgent()
\ No newline at end of file
+GET          /income-sources/manage/choose-income-sources                                               controllers.ManageIncomeSourceController.show()
```
avelx, line 153: again, optional, you can separate these which are not yet implemented in a separate section, but it really up to you

test/models/IncomeSourceDetailsModelSpec.scala
```diff
@@ -97,11 +97,32 @@ class IncomeSourceDetailsModelSpec extends UnitSpec with Matchers {
           "XIAT0000000000A",
           Some((LocalDate.now.getYear - 1).toString),
```
avelx, line 98: Please don't use LocalDate.now, use static date instead

app/services/IncomeSourceDetailsService.scala
```diff
@@ -69,4 +72,49 @@ class IncomeSourceDetailsService @Inject()(val incomeTaxViewChangeConnector: Inc
       incomeTaxViewChangeConnector.getIncomeSources()
     }
   }
+
+  def incomeSourcesAsViewModel(sources: IncomeSourceDetailsModel): AddIncomeSourcesViewModel = {
+
```
avelx, line 77: looks like the same as:
`! ceasedBusinessExists`

app/controllers/AddIncomeSourceController.scala
```diff
@@ -16,31 +16,72 @@
 
 package controllers
 
-import auth.FrontendAuthorisedFunctions
+import auth.MtdItUser
```
avelx, line 19: remove not used imports pls

PR #1700, opened 2023-05-03
MISUV 5092-5480 Update FE

danielmaltz1:


it/controllers/agent/FinalTaxCalculationControllerISpec.scala
```diff
@@ -155,9 +156,9 @@ class FinalTaxCalculationControllerISpec extends ComponentSpecBase with SessionC
       PropertyDetailsModel(
         Some("testId2"),
         Some(AccountingPeriodModel(LocalDate.now, LocalDate.now.plusYears(1))),
```
avelx, line 158: I understand that this is not your code, but we really should avoid using LocalDate.now and use some fixed date instead, if you can apply this across files you have changed pls

PR #1706, opened 2023-05-05
MISUV-5405: - Refactor IncomeSourceDetailsModel for multiple properties

ayubAz:


app/services/IncomeSourceDetailsService.scala
```diff
@@ -75,17 +75,20 @@ class IncomeSourceDetailsService @Inject()(val incomeTaxViewChangeConnector: Inc
 
   def incomeSourcesAsViewModel(sources: IncomeSourceDetailsModel): AddIncomeSourcesViewModel = {
 
-    val soleTraderBusinessesExists = !sources.businesses.forall(_.isCeased)
+    val maybeSoleTraderBusinesses = sources.businesses.filterNot(_.isCeased)
```
avelx, line 78: we might miss this in the first PR: 
* method, if there are any exceptions raised within, need to return Either[Throwable, AddIncomeSourcesViewModel], on the level of the caller you would need to handle returned value, if this is Left, log error and use errorHandler etc

PR #1712, opened 2023-05-11
MISUV-5476: SCP001 | Build Cease an income source page | Cease an income source journey

simondrugan16:


app/controllers/incomeSources/cease/CeaseIncomeSourceController.scala
```diff
@@ -16,31 +16,87 @@
...
+            )
+        }
   }
-}
+
+  def handleRequest(sources: IncomeSourceDetailsModel, isAgent: Boolean, backUrl: String)
+                   (implicit user: MtdItUser[_]): Future[Result] = {
+    Future.successful(
+      if (isDisabled(IncomeSources)) {
+        Redirect(controllers.routes.HomeController.show())
```
avelx, line 79: * pls wrap redirect with: Future.successful : which is completed Future value;
* for the incomeSourceDetailsService.getCeaseIncomeSourceViewModel call below, pls use actual Future, as this would be subject to evaluation etc

conf/messages.cy
```diff
@@ -1145,6 +1145,20 @@ new-income-sources.table-head.business-name = Enw’r busnes
...
 new-income-sources.table-head.date-ended    = Y dyddiad dod i ben
 
+## cease income sources
+cease-income-sources.heading                  = Dod â ffynhonnell incwm i ben
+cease-income-sources.self-employment.h1       = Hunangyflogaeth (unig fasnachwr)
+cease-income-sources.table-head.business-name = Enw’r busnes
+cease-income-sources.table-head.date-started  = Dyddiad cychwyn
+cease-income-sources.cease                    = Dod i ben
+cease-income-sources.uk-property.h1           = Eiddo yn y DU
+cease-income-sources.uk-property.p1           = **Welsh to be translated**
```
avelx, line 1155: assuming we still did not receive Welsh translation?

app/controllers/incomeSources/cease/CeaseIncomeSourceController.scala
avelx: I would believe you should wrap incomeSourceDetailsService.getCeaseIncomeSourceViewModel into the Future as well; you can think that IF i.e getCeaseIncomeSourceViewModel would return Future after some rework/or possible changes, this would really require this whole section to be Future

simondrugan16: Added future!

app/views/incomeSources/cease/CeaseIncomeSources.scala.html
```diff
@@ -0,0 +1,218 @@
...
+)(implicit messages: Messages,
+    user: auth.MtdItUser[_])
+
+@import implicitDateFormatter.longDate
+
+@getMessage(key: String, args: String*) = @{
+    messages(s"cease-income-sources.$key", args: _*)
+}
+
+@soleTraderBusinesses = @{ sources.soleTraderBusinesses }
```
avelx, line 46: these 4 lines not really required and duplicating what we already have in the source view model;

app/views/incomeSources/cease/CeaseIncomeSources.scala.html
```diff
@@ -0,0 +1,218 @@
...
+    else controllers.incomeSources.cease.routes.CeaseUKPropertyController.show().url
+}
+
+@ceaseForeignPropertyURL = @{
+    if(isAgent) controllers.incomeSources.cease.routes.CeaseForeignPropertyController.showAgent().url
+    else controllers.incomeSources.cease.routes.CeaseForeignPropertyController.show().url
+}
+
+@ceaseMessage = @{
+    getMessage("cease")
```
avelx, line 67: not sure why we need this method?

simondrugan16: I couldnt get getMessage to work from inside triple quotes otherwise so the Welsh translation wouldn't work

simondrugan16: Not sure why I couldn't get it to work earlier, removed now

avelx (approved):
approved with comments + requested more tests to be added as a part of another PR

PR #1716, opened 2023-05-15
MISUV-5533: hide p text for non-empty properties

ayubAz:


app/views/incomeSources/add/AddIncomeSources.scala.html
```diff
@@ -78,25 +78,30 @@
...
-            TableRow(
-                content = Text(b.tradingName),
-                attributes = Map("id" -> s"table-row-trading-name-$index")),
-
-            TableRow(
-                content = Text(b.tradingStartDate.toLongDate),
-                attributes = Map("id" -> s"table-row-trading-start-date-$index"))
-        )}
-    ))
+    @if(soleTraderBusinesses.nonEmpty) {
```
avelx, line 81: I would suspect we should have a test related to this change?

avelx (approved):
approved

PR #1720, opened 2023-05-19
MISUV-5560 - Build - Choose income source

JordHemsworth:


app/views/incomeSources/manage/ManageIncomeSources.scala.html
```diff
@@ -0,0 +1,172 @@
...
+        @govukTable(Table(
+            head = Some(List(
+                HeadCell(
+                    content = Text(getMessage("table-head.business-name")),
+                    attributes = Map("id" -> "table-head-business-name")),
+                HeadCell(
+                    content = Text(getMessage("table-head.date-started")),
+                    attributes = Map("id" -> "table-head-date-started")),
+                HeadCell(
+                    content = Text("") //TODO see on extending line w/o blank cell
```
avelx, line 81: please avoid comments in the views

app/views/incomeSources/manage/ManageIncomeSources.scala.html
```diff
@@ -0,0 +1,172 @@
...
+                )
+            )),
+            rows = soleTraderBusinesses.zipWithIndex.map { case (business, index) => Seq(
+                TableRow(
+                    content = Text(business.tradingName),
+                    attributes = Map("id" -> s"table-row-trading-name-$index")),
+                TableRow(
+                    content = Text(business.tradingStartDate.map(_.toLongDate).getOrElse("Unknown")),
```
avelx, line 89: as discussed, we need to support both language, even you might not received translation for Welsh, this must be handled according to the language selected by the user

app/models/incomeSourceDetails/viewmodels/ViewIncomeSourcesViewModel.scala
```diff
@@ -0,0 +1,32 @@
...
+package models.incomeSourceDetails.viewmodels
+
+import java.time.LocalDate
+
+case class ViewIncomeSourcesViewModel(viewSoleTraderBusinesses: List[ViewBusinessDetailsViewModel],
+                                      viewUkProperty: Option[ViewPropertyDetailsViewModel],
+                                      viewForeignProperty: Option[ViewPropertyDetailsViewModel],
+                                      viewCeasedBusinesses: List[ViewCeasedBusinessDetailsViewModel])
+
+case class ViewBusinessDetailsViewModel(tradingName: String, tradingStartDate: Option[LocalDate])
```
avelx, line 26: from what I can see in API definition / story: tradingName and tradingStartDate - can be Optional; you might need to change these as well ;

app/views/incomeSources/manage/ManageIncomeSources.scala.html
```diff
@@ -0,0 +1,172 @@
...
+
+@viewSelfEmploymentDetailsURL = @{
+    if(isAgent){
+        controllers.incomeSources.manage.routes.ManageSelfEmploymentSourceController.showAgent().url
+    } else {
+        controllers.incomeSources.manage.routes.ManageSelfEmploymentSourceController.show().url
+    }
+}
+
+@ukProperty = @{ sources.viewUkProperty }
```
avelx, line 55: * its seem we are using copy/paste approach from the similar stories;
* I don't think we really need these variable definitions at all;

app/views/incomeSources/manage/ManageIncomeSources.scala.html
```diff
@@ -0,0 +1,172 @@
...
+
+        @govukTable(Table(
+            head = Some(List(
+                HeadCell(
+                    content = Text(getMessage("table-head.date-started")),
+                    attributes = Map("id" -> "table-head-date-started-uk"))
+            )),
+            rows = Seq(Seq(
+                TableRow(
+                    content = Text(ukProperty.map(_.tradingStartDate.map(_.toLongDate).getOrElse("Unknown")).get),
```
avelx, line 117: please add language support for this text value

app/views/incomeSources/manage/ManageIncomeSources.scala.html
```diff
@@ -0,0 +1,172 @@
...
+                HeadCell(
+                    content = Text(getMessage("table-head.date-ended")),
+                    attributes = Map("id" -> "table-head-date-ended-ceased"))
+            )),
+            rows = ceasedBusinesses.zipWithIndex.map { case (b, index) => Seq(
+                TableRow(
+                    content = Text(b.tradingName),
+                    attributes = Map("id" -> s"table-row-trading-name-ceased-$index")),
+                TableRow(
+                    content = Text(b.tradingStartDate.map(_.toLongDate).getOrElse("Unknown")),
```
avelx, line 164: again, language support for "Unknown" value

conf/messages.cy
```diff
@@ -1158,6 +1158,18 @@ cease-income-sources.foreign-property.p1      = **Welsh to be translated**
...
+view-income-sources.heading                  = Bwrw golwg dros eich ffynonellau incwm a’u rheoli
+view-income-sources.self-employed-h2         = Busnesau unig fasnachwr
+view-income-sources.uk-property-h2           = Eiddo yn y DU
+view-income-sources.foreign-property-h2      = Eiddo tramor
+view-income-sources.ceased-businesses-h2     = Busnesau sydd wedi dod i ben
+view-income-sources.table-head.business-name = Enw’r busnes
+view-income-sources.table-head.date-started  = Dyddiad cychwyn
+view-income-sources.table-head.date-ended    = Y dyddiad dod i ben
+view-income-sources.business-details         = Dewis
+view-income-sources.unknown-field            = Unknown
```
avelx, line 1172: for not translated into Welsh text, would be good to have similar approach as we already have like: "To be translated to Welsh" or similar, as its easy to find it

app/services/IncomeSourceDetailsService.scala
```diff
@@ -120,6 +121,54 @@ class IncomeSourceDetailsService @Inject()(val incomeTaxViewChangeConnector: Inc
...
+          Some(ViewPropertyDetailsViewModel(
+            maybeForeignProperty.flatMap(_.tradingStartDate)
+          ))
+        } else None,
+        viewCeasedBusinesses = if (ceasedBusinessExists) {
+          maybeCeasedBusinesses.map { business =>
+            ViewCeasedBusinessDetailsViewModel(
+              tradingName = business.tradingName.getOrElse("Unknown"),
+              tradingStartDate = business.tradingStartDate,
+              cessationDate = business.cessation.flatMap(_.date).getOrElse(LocalDate.parse("Unknown"))
```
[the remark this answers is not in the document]
JordHemsworth: Unsure how to display the unknown for business name when throwing the exception if its mandatory field

app/views/incomeSources/manage/ManageIncomeSources.scala.html
JordHemsworth: Done

app/views/incomeSources/manage/ManageIncomeSources.scala.html
JordHemsworth: Why would I not need them?

conf/messages.cy
JordHemsworth: Done

app/views/incomeSources/manage/ManageIncomeSources.scala.html
avelx: its really up to you, but use of these one line variables seems to be redundant

app/services/IncomeSourceDetailsService.scala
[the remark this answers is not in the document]
avelx: we had a discussion yesterday regarding similar cases;
Here is what we have agreed (this would be similar for any field where we have to display 'Unknown'): 
* set tradingName as Optional in the view model, so no need to throw error when you map API returned field to view model field;
* in the actual view, do next call tradingName.getElse(  @getMessage("~.relevant.resource.for.nnknown") )

so, we would show something in case there is no value provided for this field etc

app/controllers/incomeSources/manage/ManageIncomeSourceController.scala
```diff
@@ -16,31 +16,85 @@
...
+              isAgent = true,
+              backUrl = controllers.routes.HomeController.showAgent.url
+            )
+        }
   }
-}
+
+  def handleRequest(sources: IncomeSourceDetailsModel, isAgent: Boolean, backUrl: String)
+                   (implicit user: MtdItUser[_]): Future[Result] = {
+    Future.successful(
```
avelx, line 75: if you can wrap only redirect with Future.successful
on line 79: please use Future { .. }

avelx (approved):
approved

PR #1726, opened 2023-05-24
MISUV-5476-fix accessibility-issues

simondrugan16:
MISUV-5476-extra-tests-and-welsh-translation


app/views/incomeSources/cease/CeaseIncomeSources.scala.html
```diff
@@ -70,39 +71,34 @@
...
-                )
-            )},
-            attributes = Map("id" -> "sole-trader-businesses-table")
-        ))
+        <table class="govuk-table govuk-!-margin-bottom-3" id="sole-trader-businesses-table">
+            <thead class="govuk-table__head">
+                <tr class="govuk-table__row">
+                    <th class="govuk-table__header" scope="col" id="table-head-business-name">@getMessage("table-head.business-name")</th>
+                    <th class="govuk-table__header" scope="col" id="table-head-date-started">@getMessage("table-head.date-started")</th>
+                    <td class="govuk-table__header" id="empty0"></td>
```
avelx, line 79: this might fail one of accessibility rule: => _id="empty0"_

app/views/incomeSources/cease/CeaseIncomeSources.scala.html
```diff
@@ -113,32 +109,30 @@
...
+        <table class="govuk-table govuk-!-margin-bottom-3" id="uk-property-table">
+            <thead class="govuk-table__head">
+                <tr class="govuk-table__row">
+                    <th class="govuk-table__header" scope="col" id="table-head-date-started-uk">@getMessage("table-head.date-started")</th>
+                    <td class="govuk-table__header" id="empty1"></td>
+                </tr>
+            </thead>
+            <tbody>
+                <tr class="govuk-table__row">
+                    <td class="govuk-table__cell" id="table-row-trading-start-date-uk">@sources.ukProperty.map(_.tradingStartDate.toLongDate).get</td>
```
avelx, line 121: the whole point to have a separate model was to avoid these hidden bugs, essentially you not handling situation when ukProperty field is not set;

app/views/incomeSources/cease/CeaseIncomeSources.scala.html
```diff
@@ -149,61 +143,53 @@
...
+        <table class="govuk-table govuk-!-margin-bottom-3" id="foreign-property-table">
+            <thead class="govuk-table__head">
+                <tr class="govuk-table__row">
+                    <th class="govuk-table__header" scope="col" id="table-head-date-started-foreign">@getMessage("table-head.date-started")</th>
+                    <td class="govuk-table__header" id="empty2"></td>
+                </tr>
+            </thead>
+            <tbody>
+                <tr class="govuk-table__row">
+                    <td class="govuk-table__cell" id="table-row-trading-start-date-foreign">@sources.foreignProperty.map(_.tradingStartDate.toLongDate).get</td>
```
avelx, line 155: same issue as above, we should properly handle scenario when foreignProperty is None

app/views/incomeSources/cease/CeaseIncomeSources.scala.html
simondrugan16: @if(sources.ukProperty.nonEmpty) {
this line at the top of the block solves this i think. ukProperty field is an option so the table only shows when the option is Some(ukProperty)

app/views/incomeSources/cease/CeaseIncomeSources.scala.html
simondrugan16: I thought I solved this accessibility issue by using the td tag rather than the th tag like for the other headings

avelx: for heading we should use th in html table; but my concern was around using empty0 in the id attribute; lets leave this to a11y job;

avelx (approved):
approved; looks good

PR #1727, opened 2023-05-25
MISUV-4891: SCP025 | Add SE | Build - Add business start date page and Add business start date check page

ayubAz:


app/controllers/AddBusinessStartDateController.scala
```diff
@@ -0,0 +1,150 @@
...
+                                               implicit val dateFormatter: ImplicitDateFormatterImpl,
+                                               implicit val itvcErrorHandlerAgent: AgentItvcErrorHandler,
+                                               implicit override val mcc: MessagesControllerComponents,
+                                               val ec: ExecutionContext)
+  extends ClientConfirmedController with I18nSupport with FeatureSwitching {
+
+  lazy val backUrl: String = routes.AddBusinessStartDateCheckController.show().url
+  lazy val backUrlAgent: String = routes.AddBusinessStartDateCheckController.showAgent().url
+
+  def show(): Action[AnyContent] = (checkSessionTimeout andThen authenticate andThen retrieveNino
```
avelx, line 60: duplicated code in function definition

app/controllers/AddBusinessStartDateController.scala
```diff
@@ -0,0 +1,150 @@
...
+          Ok(addBusinessStartDate(
+            form = BusinessStartDateForm(),
+            postAction = postAction,
+            backUrl = backUrl,
+            isAgent = isAgent
+          ))
+        }
+      ) recover {
+        case ex: Exception =>
+          Logger("application").error(s"${if (isAgent) "[Agent]"}" +
```
avelx, line 104: I am not sure if this would be evaluated correctly;

app/controllers/AddBusinessStartDateController.scala
```diff
@@ -0,0 +1,150 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package controllers
+
+import auth.MtdItUser
```
avelx, line 19: pls remove not used imports

app/forms/incomeSources/add/BusinessStartDateForm.scala
```diff
@@ -0,0 +1,106 @@
...
+  override protected def maxDate(maximum: LocalDate, errorKey: String, args: Any*): Constraint[LocalDate] =
+    Constraint {
+      case date if date.isAfter(maximum) || date.equals(maximum) =>
+        Invalid(errorKey, args: _*)
+      case _ =>
+        Valid
+    }
+}
+
+object BusinessStartDateForm {
```
avelx, line 102: you are using mixing approach here: passing implict in apply method and passing these as params to instantiate object for DI ; 
there is really no need to use DI like other devs are doing, please avoid this;
here is example how you can do this:
https://github.com/hmrc/income-tax-view-change-frontend/compare/MISUV-4891-1_fix?expand=1

app/views/components/dateInput.scala.html
```diff
@@ -21,29 +21,28 @@
...
-    hintContent: Option[String] = None,
-    smallText: Boolean = false,
-    largeText: Boolean = true,
-    labelClass: Option[String] = None,
-    pageHeading: Boolean = true
+        field: Field,
+        hintContent: Option[Content] = None,
+        smallText: Boolean = false,
+        largeText: Boolean = true,
+        labelClass: Option[String] = None,
```
avelx, line 28: not sure if UI wise there will be ever situation when labelClass will be None? maybe not worth to model this field like Optional?

avelx (changes requested):
please address comments

app/forms/incomeSources/add/BusinessStartDateForm.scala
```diff
@@ -0,0 +1,106 @@
...
+      Invalid(Seq(ValidationError(dayRequired)))
+    case (day, month, year) if day.trim.nonEmpty && month.trim.isEmpty && year.trim.nonEmpty =>
+      Invalid(Seq(ValidationError(monthRequired)))
+    case (day, month, year) if day.trim.nonEmpty && month.trim.nonEmpty && year.trim.isEmpty =>
+      Invalid(Seq(ValidationError(yearRequired)))
+    case _ =>
+      Valid
+  }
+
+  override protected def maxDate(maximum: LocalDate, errorKey: String, args: Any*): Constraint[LocalDate] =
```
edward-dennis-ey, line 93: Not sure why this has been overridden?

edward-dennis-ey: The requirements are to stop entry above 7 days, so it should be after (not equals) as per my implementation of this function

app/controllers/AddBusinessStartDateController.scala
```diff
@@ -0,0 +1,150 @@
...
+          formWithErrors =>
+            BadRequest(addBusinessStartDate(
+              form = formWithErrors,
+              postAction = postAction,
+              backUrl = backUrl,
+              isAgent = false
+            )),
+          formData =>
+            Redirect(routes.AddBusinessStartDateCheckController.show())
+              .addingToSession(SessionKeys.businessStartDate -> formData.date.toString)
```
avelx, line 124: I don't see any immediate issues related to the format we store date, but just to note that this is a sensitive part

app/forms/incomeSources/add/BusinessStartDateForm.scala
```diff
@@ -0,0 +1,91 @@
...
+        month -> default(text(), ""),
+        year -> default(text(), ""))
+        .verifying(firstError(
+          checkRequiredFields,
+          validDate(dateMustBeReal)
+        )).transform[LocalDate](
+        {
+          case (day, month, year) => LocalDate.of(year.toInt, month.toInt, day.toInt)
+        },
+        date => (date.getDayOfMonth.toString, date.getMonthValue.toString, date.getYear.toString)
```
avelx, line 67: this would be easy to store these as Int, but lets leave these for next iteration / refactoring

avelx (approved):
approved

app/controllers/AddBusinessStartDateCheckController.scala
```diff
@@ -18,55 +18,155 @@ package controllers
...
+            )
+        }
+  }
+
+  def handleRequest(isAgent: Boolean,
+                    backUrl: String,
+                    homePageCall: Call,
+                    itvcErrorHandler: ShowInternalServerError)
+                   (implicit request: Request[_], mtdItUser: MtdItUser[_]): Future[Result] = {
+    Future.successful(
```
avelx, line 122: Please wrap only Redirect part with Future.success;
On line 126 you should use actual Future instead

app/forms/BusinessStartDateCheckForm.scala
```diff
@@ -0,0 +1,47 @@
...
+import play.api.data.Forms._
+
+
+object BusinessStartDateCheckForm {
+
+  val responseNo: String = "No"
+  val responseYes: String = "Yes"
+  val response: String = "add-business-start-date-check"
+  val radiosEmptyError: String = "add-business-start-date-check.radio.error"
+  val csrfToken: String = "csrfToken"
```
avelx, line 29: I dont really think we need this value => as this is something to be supported by the framework; but worth to confirm
`
val csrfToken: String = "csrfToken"
`

conf/messages
```diff
@@ -1212,4 +1230,14 @@ incomeSources.cease.UKPropertyEndDate.error.beforeStartDate      = The cessation
 
 incomeSources.ceaseForeignProperty.heading                       = Cease foreign property
 incomeSources.ceaseForeignProperty.checkboxLabel                 = I declare that I no longer get an income from any foreign property.
-incomeSources.ceaseForeignProperty.checkboxError                 = You must declare that you no longer get an income from any foreign property
\ No newline at end of file
+incomeSources.ceaseForeignProperty.checkboxError                 = You must declare that you no longer get an income from any foreign property
+
+#Generic date form messages
+dateForm.hint                                                       = For example, 27 3 2020
```
avelx, line 1236: I don't think this is the only one form have date input, so you need to have more details in the prefix i.e
> addbusinessstartdateform.hint ?

app/controllers/incomeSources/add/AddIncomeSourceController.scala
```diff
@@ -62,38 +66,40 @@ class AddIncomeSourceController @Inject()(val addIncomeSources: AddIncomeSources
...
-            } else {
-              Logger("application").error(
-                s"[AddIncomeSourceController][handleRequest] - Error: ${ex.getMessage}")
-              itvcErrorHandler.showInternalServerError()
-            }
-        }
+    if (isDisabled(IncomeSources)) {
+      Future.successful(Redirect(homePageCall))
+    } else {
+      incomeSourceDetailsService.getAddIncomeSourceViewModel(sources) match {
```
avelx, line 85: would be easier to wrap this whole expression with Future, but I guess we can leave things as is

avelx (approved):
approved

PR #1735, opened 2023-05-30
MISUV-4891: Create Confirmation start date page UI

ayubAz:


avelx:
@ayubAz  - please make sure we have required it:tests

PR #1763, opened 2023-06-12
MISUV-5382: SCP001 | Build - Check your details page for ceasing Foreign property business | Cease an income source

ayubAz:


app/services/UpdateIncomeSourceService.scala
```diff
@@ -44,4 +45,18 @@ class UpdateIncomeSourceService @Inject()(connector: IncomeTaxViewChangeConnecto
 
   }
 
+  def updateCessationDatev2(nino: String, incomeSourceId: String, cessationDate: String)
+                           (implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Either[UpdateIncomeSourceError.type, UpdateIncomeSourceSuccess.type]]= {
+    connector.updateCessationDate(
```
avelx, line 50: I would suggest to change next case classes signature for UpdateIncomeSourceError and  UpdateIncomeSourceSuccess to:

`
case class UpdateIncomeSourceError(reson: String)
case class UpdateIncomeSourceSuccess(incomeSourceId: String)
// and also method signature:
  def updateCessationDatev2(nino: String, incomeSourceId: String, cessationDate: String)
                           (implicit hc: HeaderCarrier, ec: ExecutionContext): **Future[Either[UpdateIncomeSourceError, UpdateIncomeSourceSuccess]]**
`

lets assume that caller of this service method (controller) would decided if we need to log these details or not

app/controllers/AddBusinessNameController.scala
```diff
@@ -76,12 +76,15 @@ class AddBusinessNameController @Inject()(authenticate: AuthenticationPredicate,
   def handleRequest(isAgent: Boolean, backUrl: String)(implicit user: MtdItUser[_], ec: ExecutionContext): Future[Result] = {
     Future {
       if (isDisabled(IncomeSources)) {
-        Redirect(controllers.routes.HomeController.show())
+        Redirect(
+          if(isAgent) controllers.routes.HomeController.showAgent
```
avelx, line 80: if you can wrap Redirect into Future.Future.successful which is a blocking call and rendering of the view below into actual Future;

avelx:
@ayubAz - if we can increase test coverage for CheckCeaseForeighPropertyDetailsController, see details below:

image

app/controllers/incomeSources/cease/CheckCeaseForeignPropertyDetailsController.scala
```diff
@@ -36,28 +41,137 @@ class CheckCeaseForeignPropertyDetailsController @Inject()(val authenticate: Aut
...
+                          successCall: Call,
+                          cessationDate: String,
+                          itvcErrorHandler: ShowInternalServerError)
+                         (implicit user: MtdItUser[_]): Future[Result] = {
+
+    if (isDisabled(IncomeSources)) {
+      Future.successful(Redirect(homePageCall))
+    } else {
+
+      val maybeIncomeSourceId = user.incomeSources.properties
```
avelx, line 155: - I would assume this is some sort of logic to verify that this incomeSource exists before executing Runtime logic to drop incomeSource?  
- Because we intended incomeSource to be cached in the user object, this part might not work well (caching might not be up to date etc);  this might be okis to fully reply on the call to drop incomeSource without checking if this incomeSource exist;

avelx (approved):
approved

PR #1768, opened 2023-06-16
MISUV-5730-business-accounting-method-page

simondrugan16:


app/controllers/incomeSources/add/BusinessAccountingMethodController.scala
```diff
@@ -0,0 +1,171 @@
...
+    } recover {
+      case ex: Exception =>
+        Logger("application").error(s"${if (isAgent) "[Agent]"}" +
+          s"Error getting BusinessEndDate page: ${ex.getMessage}")
+        errorHandler.showInternalServerError()
+    }
+  }
+
+  def shouldAutomaticallyRedirect(businesses: List[BusinessDetailsModel]): Boolean = {
+    val activeSoleTraderBusinesses: Boolean = !businesses.forall(_.isCeased)
```
avelx, line 97: this variable definition is redundant, used only once

simondrugan16: I think it's good for readability because this function is just the condition inside an if statement to decide whether to automatically redirect or not

it/controllers/incomeSources/add/BusinessAccountingMethodControllerISpec.scala
```diff
@@ -0,0 +1,111 @@
...
+import config.featureswitch.IncomeSources
+import forms.utils.SessionKeys.addBusinessAccountingMethod
+import helpers.ComponentSpecBase
+import helpers.servicemocks.IncomeTaxViewChangeStub
+import play.api.http.Status.{BAD_REQUEST, OK, SEE_OTHER}
+import testConstants.BaseIntegrationTestConstants.testMtditid
+import testConstants.IncomeSourceIntegrationTestConstants.noPropertyOrBusinessResponse
+
+class BusinessAccountingMethodControllerISpec extends ComponentSpecBase {
+
```
avelx, line 28: just wondering if we have similar it:test for agent? or this all incorporate in this single test?

simondrugan16: I noticed that none of the it:tests had separate agent tests so thought they weren't necessary (I also tried writing them myself but couldn't get anywhere). the controller unit tests have agent tests, however

simondrugan16: Hmm... I must've been looking in the wrong place as I can see them now. Will get on that

Joshtest1570 (approved):

PR #1769, opened 2023-06-19
MISUV-5735 - Build - Check your details

JordHemsworth:


app/controllers/incomeSources/add/CheckBusinessDetailsController.scala
```diff
@@ -16,62 +16,232 @@
...
+      user.session.data.get("addBusinessStartDate").orElse(Option(MissingKey("MissingKey: addBusinessStartDate"))),
+      user.session.data.get("addBusinessTrade").orElse(Option(MissingKey("MissingKey: addBusinessTrade"))),
+      user.session.data.get("addBusinessAddressLine1").orElse(Option(MissingKey("MissingKey: addBusinessAddressLine1"))),
+      user.session.data.get("addBusinessPostalCode").orElse(Option(MissingKey("MissingKey: addBusinessPostalCode")))
+    ).collect {
+      case Some(MissingKey(msg)) => MissingKey(msg)
+    }.map(e => e.msg)
+
+    val result: Option[CheckBusinessDetailsViewModel] = for {
+        businessName <- user.session.data.get("addBusinessName")
```
avelx, line 94: we have to use sessionKey defined in the SessionKeys.scala file instead of string values

app/controllers/incomeSources/add/CheckBusinessDetailsController.scala
```diff
@@ -16,62 +16,232 @@
...
+          implicit mtdItUser =>
+            getDetails(mtdItUser).toOption match {
+              case Some(viewModel: CheckBusinessDetailsViewModel) =>
+                businessDetailsService.createBusinessDetails(viewModel) map {
+                  case Left(ex) => Logger("application").error(
+                    s"[CheckBusinessDetailsController][handleRequest] - Unable to create income source: ${ex.getMessage}")
+                    itvcErrorHandler.showInternalServerError()
+
+                  case Right(_) =>
+                    mtdItUser.session.data - "addBusinessName" - "addBusinessStartDate" - "addBusinessTrade" - "addBusinessAddressLine1" - "addBusinessPostalCode" - "addBusinessAccountingMethod"
```
avelx, line 181: remove line 181,
please use 
Redirect("/home").withNewSession 
as discussed

app/controllers/incomeSources/add/CheckBusinessDetailsController.scala
```diff
@@ -16,62 +16,232 @@
...
+              case None => Logger("application").error(
+                s"[CheckBusinessDetailsController][submit] - Error: Unable to build view model on submit")
+                Future.successful(itvcErrorHandler.showInternalServerError())
+            }
+        }
+  }
+
+
+  def changeBusinessName(): Action[AnyContent] = Action {
+    Ok("Change Business Name WIP")
```
avelx, line 193: we need implementation to where were are going to redirect for each of these methods;
I would assume we have most of these controllers already created, so relevant public methods should be created otherwise create one with no implementation;

app/controllers/incomeSources/add/CheckBusinessDetailsController.scala
JordHemsworth: Done

app/controllers/incomeSources/add/CheckBusinessDetailsController.scala
JordHemsworth: done

app/controllers/incomeSources/add/CheckBusinessDetailsController.scala
JordHemsworth: As discussed will leave these as they are new unimplemented change pages, can be updated once work is there

app/models/incomeSourceDetails/viewmodels/CheckBusinessDetailsViewModel.scala
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.incomeSourceDetails.viewmodels
+
+import java.time.LocalDate
+case class CheckBusinessDetailsViewModel(businessName: Option[String], businessStartDate: Option[LocalDate],
```
avelx, line 20: looks like quite a few fields are missing:
ie.
businessDetails.accountingPeriodEndDate
please cross check with required fields for API which consume these in the spec
--

JordHemsworth: Ok will do

Joshtest1570 (approved):

PR #1780, opened 2023-06-23
MISUV-5188 Added correct headings, works in both staging and local, other updates

danielmaltz1:


app/connectors/AddressLookupConnector.scala
```diff
@@ -82,6 +85,58 @@ class AddressLookupConnector @Inject()(val appConfig: FrontendAppConfig,
...
             "ukMode" -> JsBoolean(true)
           )
+        ),
+        "labels" -> JsObject(
+          Seq(
+            "en" -> JsObject(
+              Seq(
+                "selectPageLabels" -> JsObject(
+                  Seq(
+                    "heading" -> JsString(messagesApi.preferred(Seq(Lang("en")))("add-business-address.select.heading"))
```
avelx, line 95: we don't now what language user selected, this should come from request context

avelx (approved):
approved

PR #1787, opened 2023-07-03
WIP Misuv 5664 - Add SE | Obligations (business added)

danielmaltz1:


app/controllers/incomeSources/add/BusinessAddedObligationsController.scala
```diff
@@ -0,0 +1,156 @@
...
+  def handleRequest(isAgent: Boolean, id: Option[String])(implicit user: MtdItUser[_], ec: ExecutionContext): Future[Result] = {
+    if (isDisabled(IncomeSources)) {
+      Future.successful(Redirect(controllers.routes.HomeController.show()))
+    } else {
+      id match {
+        case None => Logger("application").error(
+          s"[BusinessAddedObligationsController][handleRequest] - Error: No id supplied by reporting method page")
+          Future(itvcErrorHandler.showInternalServerError())
+        case Some(value) => {
+          if (!user.incomeSources.businesses.exists(x => x.incomeSourceId.contains(value))) {
```
avelx, line 87: this is redundant step, as you calling service, and handling this scenario already +
we should make assumption that incomeSource ids under specific user will always belong to the user

app/controllers/incomeSources/add/BusinessAddedObligationsController.scala
```diff
@@ -0,0 +1,167 @@
...
+  def handleRequest(isAgent: Boolean, id: Option[String])(implicit user: MtdItUser[_], ec: ExecutionContext): Future[Result] = {
+    if (isDisabled(IncomeSources)) {
+      Future.successful(Redirect(controllers.routes.HomeController.show()))
+    } else {
+      id match {
+        case None => Logger("application").error(
+          s"[BusinessAddedObligationsController][handleRequest] - Error: No id supplied by reporting method page")
+          Future(itvcErrorHandler.showInternalServerError())
+        case Some(value) =>
+          val addedBusiness: BusinessDetailsModel = user.incomeSources.businesses.find(x => x.incomeSourceId.contains(value)).get
```
avelx, line 85: this would cause error in case in case businessModel no provided
this would be much better to use for-comprehension in this situation, like:
`
maybeId :  Option[String] // in the method definition
.....
      for {
        incomeSourceId <- maybeId
        addedBusiness <- user.incomeSources.businesses.find(x => x.incomeSourceId.contains(incomeSourceId))
      } yield addedBusiness`

app/controllers/incomeSources/add/BusinessAddedObligationsController.scala
```diff
@@ -0,0 +1,167 @@
...
+            Logger("application").error(
+              s"[BusinessAddedObligationsController][handleRequest] - No business name for business with provided id")
+            itvcErrorHandler.showInternalServerError()
+          }
+          if (addedBusiness.tradingStartDate.isEmpty) {
+            Logger("application").error(s"[BusinessAddedObligationsController][handleRequest] - No business start date for business with provided id")
+            Future(itvcErrorHandler.showInternalServerError())
+          }
+
+          val businessName: String = addedBusiness.tradingName.get
```
avelx, line 97: again, calling .get is not safe way in we encourage to avoid this was of coding

app/controllers/incomeSources/add/BusinessAddedObligationsController.scala
```diff
@@ -0,0 +1,167 @@
...
+
+          val dates: Seq[DatesModel] = getObligationDates(id.get)
+
+          val quarterlyDates: Seq[DatesModel] = dates.filter(x => x.periodKey.contains("00")).sortBy(_.inboundCorrespondenceFrom)
```
avelx, line 101: its worth to consider refactor this code in to service or model

app/controllers/incomeSources/add/BusinessAddedObligationsController.scala
danielmaltz1: The if statement above handles the case where the name isn't defined and throws an error, so the get is safe. I understand if it's still bad coding practice and you'd like me to change it

app/controllers/incomeSources/add/BusinessAddedObligationsController.scala
danielmaltz1: I had a case for if the business with the id did not exist but in the above comment you said that was a "redundant step" so I removed it. Will implement something like this now

app/controllers/incomeSources/add/BusinessAddedObligationsController.scala
```diff
@@ -0,0 +1,167 @@
...
+          val quarterlyDatesByYear: (Seq[DatesModel], Seq[DatesModel]) = quarterlyDates.partition(x => dateService.getAccountingPeriodEndDate(x.inboundCorrespondenceTo) == dateService.getAccountingPeriodEndDate(quarterlyDates.head.inboundCorrespondenceTo))
+
+          val eopsDates: Seq[DatesModel] = dates.filter(x => x.periodKey == "EOPS")
+
+          val finalDeclarationDates: Seq[DatesModel] = getFinalDeclarationDates
+
+          val showPreviousTaxYears: Boolean = addedBusiness.tradingStartDate.get.isBefore(getStartOfCurrentTaxYear)
```
avelx, line 108: why field really can be a part of view model, so we don't have to worry about passing any extra parameters to the view ...; I would say this is optional at this stage

app/controllers/incomeSources/add/BusinessAddedObligationsController.scala
danielmaltz1: I think in terms of readability this is fine, there's nothing that requires it to be separated, and it's calling a service so keeping it in the connector is fine. I only took it out of the main handleRequest method for readability, it's just a chunk of code, but I think it belongs in the connector

app/controllers/incomeSources/add/BusinessAddedObligationsController.scala
danielmaltz1: True, I'll move this into the view model

app/controllers/incomeSources/add/BusinessAddedObligationsController.scala
```diff
@@ -0,0 +1,167 @@
...
+              controllers.incomeSources.add.routes.BusinessAddedObligationsController.agentSubmit(), agentBackUrl, isAgent = true, showPrevTaxYears = showPreviousTaxYears))
+            else Ok(obligationsView(businessName, ObligationsViewModel(quarterlyDatesByYear._1, quarterlyDatesByYear._2, eopsDates, finalDeclarationDates, dateService.getCurrentTaxYearEnd()),
+              controllers.incomeSources.add.routes.BusinessAddedObligationsController.submit(), backUrl, isAgent = false, showPrevTaxYears = showPreviousTaxYears))
+          }
+      }
+    }
+  }
+
+  private def getObligationDates(id: String)(implicit user: MtdItUser[_], ec: ExecutionContext): Seq[DatesModel] = {
+    Await.result(nextUpdatesService.getNextUpdates() map {
```
avelx, line 122: we are doing everything strictly in async way, no blocking calls pls
you have to return Future[Seq[DatesModel]]

app/controllers/incomeSources/add/BusinessAddedObligationsController.scala
```diff
@@ -0,0 +1,167 @@
...
+        Seq.empty
+      case NextUpdateModel(start, end, due, obligationType, dateReceived, periodKey) =>
+        Seq(DatesModel(start, end, due, obligationType, periodKey))
+      case ObligationsModel(obligations) =>
+        obligations.filter(x => x.identification == id).flatMap(obligation => obligation.obligations.map(x => DatesModel(x.start, x.end, x.due, x.obligationType, x.periodKey)))
+    }, Duration(100, MILLISECONDS))
+  }
+
+  private def getFinalDeclarationDates(implicit user: MtdItUser[_], ec: ExecutionContext): Seq[DatesModel] = {
+    Await.result(nextUpdatesService.getNextUpdates() map {
```
avelx, line 135: no blocking calls pls

app/controllers/incomeSources/add/BusinessAddedObligationsController.scala
```diff
@@ -0,0 +1,167 @@
...
+      case model: ObligationsModel => model.allCrystallised map {
+        source => DatesModel(source.obligation.start, source.obligation.end, source.obligation.due, source.obligation.obligationType, source.obligation.periodKey)
+      }
+    }, Duration(100, MILLISECONDS))
+  }
+
+  private def getStartOfCurrentTaxYear: LocalDate = {
+    val currentDate = dateService.getCurrentDate()
+    if (currentDate.isBefore(LocalDate.of(currentDate.getYear, APRIL, 6))) LocalDate.of(currentDate.getYear - 1, APRIL, 6)
```
avelx, line 144: manipulation with the dates, where required a new logic, should be moved under DateServiceInterface

app/controllers/incomeSources/add/BusinessAddedObligationsController.scala
```diff
@@ -0,0 +1,124 @@
...
+                                                   nextUpdatesService: NextUpdatesService)
+                                                  (implicit val appConfig: FrontendAppConfig,
+                                                   implicit val itvcErrorHandlerAgent: AgentItvcErrorHandler,
+                                                   implicit override val mcc: MessagesControllerComponents,
+                                                   val ec: ExecutionContext,
+                                                   dateService: DateServiceInterface)
+  extends ClientConfirmedController with I18nSupport with FeatureSwitching {
+
+
+  def show(id: String): Action[AnyContent] = (checkSessionTimeout andThen authenticate andThen retrieveNino
```
avelx, line 58: ideally we should use incomeSourceId instead of 'id', but this is really optional I would say

app/services/NextUpdatesService.scala
```diff
@@ -100,4 +103,43 @@ class NextUpdatesService @Inject()(val incomeTaxViewChangeConnector: IncomeTaxVi
...
+            DatesModel(source.start, source.end, source.due, source.periodKey, isFinalDec = true)
+        },
+          model.obligations.filter(x => x.identification == id).flatMap(obligation => obligation.obligations.map(x => DatesModel(x.start, x.end, x.due, x.periodKey, isFinalDec = false)))
+        ).flatten
+    }
+  }
+
+  def getObligationsViewModel(id: String, showPreviousTaxYears: Boolean)
+                             (implicit user: MtdItUser[_], ec: ExecutionContext, hc: HeaderCarrier): Future[ObligationsViewModel] = {
+    val processingRes = for {
```
avelx, line 129: there are chances that this transformation can failed:
* filtering on the line 134
* use of .head on line 135

app/views/incomeSources/add/BusinessAddedObligations.scala.html
```diff
@@ -0,0 +1,183 @@
...
+            <table class="govuk-table">
+                <thead class="govuk-table__head">
+                <tr class="govuk-table__row">
+                    <th scope="col" class="govuk-table__header govuk" style="width:55%;">@getMessage("table.heading-1")</th>
+                    <th scope="col" class="govuk-table__header govuk" style="width:30%;">@getMessage("table.heading-2")</th>
+                </tr>
+                </thead>
+
+                <tbody class="govuk-table__body">
+                    @sources.finalDeclarationDates.distinct.sortBy(_.inboundCorrespondenceFrom).map{ x =>
```
avelx, line 154: this would be better to do distinct / and sorting before you created view model;

app/services/DateService.scala
```diff
@@ -42,6 +42,12 @@ class DateService @Inject()(implicit val frontendAppConfig: FrontendAppConfig) e
     if (isBeforeLastDayOfTaxYear(isTimeMachineEnabled)) currentDate.getYear else currentDate.getYear + 1
   }
 
+  def getCurrentTaxYearStart(isTimeMachineEnabled: Boolean): LocalDate = {
```
avelx, line 45: we expect test coverage for any new method added to the service

avelx (changes requested):
please address comments

app/services/NextUpdatesService.scala
danielmaltz1: I thought we agreed the .head was fine because x.partition(y => y==x.head) just returns two empty lists if x is empty, and .filter will be the same thing?

avelx (approved):
approved

PR #1788, opened 2023-07-04
MISUV-5796-fix-business-accounting-method

simondrugan16:


avelx:
@simondrugan16  - assuming this was raised as a bug, we would expect at least one Or few test to cover this change ...

app/controllers/incomeSources/add/BusinessAccountingMethodController.scala
```diff
@@ -57,22 +57,23 @@ class BusinessAccountingMethodController @Inject()(val authenticate: Authenticat
...
     val postAction: Call = if (isAgent) controllers.incomeSources.add.routes.BusinessAccountingMethodController.submitAgent() else
       controllers.incomeSources.add.routes.BusinessAccountingMethodController.submit()
 
     val errorHandler: ShowInternalServerError = if (isAgent) itvcErrorHandlerAgent else itvcErrorHandler
 
     if (incomeSourcesEnabled) {
       val userBusinesses: List[BusinessDetailsModel] = user.incomeSources.businesses
-      if (shouldAutomaticallyRedirect(userBusinesses)) {
+      if (user.incomeSources.checkBusinessesCashOrAccrualsTypes && !userBusinesses.forall(_.isCeased)) {
+        val accountingMethod: String = userBusinesses.head.cashOrAccruals.getOrElse(throw MissingFieldException("cashOrAccruals field missing"))
```
avelx, line 70: * lines 69 - 70 can be moved within relevant model and we are not handling error which can be raised here

app/models/incomeSourceDetails/IncomeSourceDetailsResponse.scala
```diff
@@ -27,13 +29,20 @@ case class IncomeSourceDetailsModel(mtdbsa: String,
...
-                                   ) extends IncomeSourceDetailsResponse {
+                                   ) extends IncomeSourceDetailsResponse with Logging {
 
   val hasPropertyIncome: Boolean = properties.nonEmpty
   val hasBusinessIncome: Boolean = businesses.nonEmpty
   val hasOngoingBusinessOrPropertyIncome: Boolean = businesses.exists(b => b.cessation.forall(_.date.isEmpty)) ||
     properties.exists(p => p.cessation.forall(_.date.isEmpty))
 
+  def doAllBusinessesCashOrAccrualsMatch(): Boolean = if (businesses.filterNot(_.isCeased).flatMap(_.cashOrAccruals).distinct.size > 1) {
+    logger.warn("user businesses had different cashOrAccruals types")
```
avelx, line 40: logging on the level of the model might be not a good idea

app/controllers/incomeSources/add/BusinessAccountingMethodController.scala
```diff
@@ -53,26 +54,36 @@ class BusinessAccountingMethodController @Inject()(val authenticate: Authenticat
...
 
     if (incomeSourcesEnabled) {
-      val userBusinesses: List[BusinessDetailsModel] = user.incomeSources.businesses
-      if (shouldAutomaticallyRedirect(userBusinesses)) {
+
+      doAllBusinessesCashOrAccrualsMatch(user)
+
+      val userActiveBusinesses: List[BusinessDetailsModel] = user.incomeSources.businesses.filterNot(_.isCeased)
+      if (!userActiveBusinesses.forall(_.isCeased)) {
+        val accountingMethod: String = userActiveBusinesses.head.cashOrAccruals.getOrElse(throw MissingFieldException("cashOrAccruals field missing"))
```
avelx, line 80: you are not handling error in the right way; in case this would happen, you should not raise error

avelx (changes requested):
as discussed

avelx (approved):
approved

avelx (approved):
approved

PR #1797, opened 2023-07-12
MISUV-5377-add-foreign-property-check-details

simondrugan16:


app/views/incomeSources/add/ForeignPropertyCheckDetails.scala.html
```diff
@@ -0,0 +1,123 @@
...
+        captionClasses = Some("govuk-caption-l")
+    ))
+
+    @formWithCSRF(action = postAction) {
+
+        <dl class="govuk-summary-list" id="refund-to-taxpayer-table">
+            <div class="govuk-summary-list__row">
+                <dt class="govuk-summary-list__key">@getMessage("start-date")</dt>
+                <dd class="govuk-summary-list__value" id="foreign-property-date-value">@{
+                    viewModel.tradingStartDate.get.toLongDate
```
avelx, line 89: tradingStartDate - this field should really start as mandatory in the view model so we are not calling .get over Option; when mapping we can raise error if this date is not provided;

app/controllers/incomeSources/add/ForeignPropertyCheckDetailsController.scala
```diff
@@ -0,0 +1,203 @@
...
+          } else {
+            Logger("application").error(
+              s"[ForeignPropertyCheckDetailsController][handleRequest] - Error: Unable to construct Future ${ex.getMessage}")
+            itvcErrorHandler.showInternalServerError()
+          }
+      }
+    }
+  }
+
+  def getDetails(implicit user: MtdItUser[_]): Either[Throwable, CheckForeignPropertyViewModel] = {
```
avelx, line 106: as per agreement we have recently in tech-channel*,
result of this function should be Future[Either[Throwable, CheckForeignPropertyViewModel]] - as this simplify integrating reading these data from Mongo and this is a new code

app/controllers/incomeSources/add/ForeignPropertyCheckDetailsController.scala
```diff
@@ -0,0 +1,207 @@
...
+          } else {
+            Logger("application").error(
+              s"[ForeignPropertyCheckDetailsController][handleRequest] - Error: Unable to construct Future ${ex.getMessage}")
+            errorHandler.showInternalServerError()
+          }
+      }
+    }
+  }
+
+  def getDetails(implicit user: MtdItUser[_]): Future[Either[Throwable, CheckForeignPropertyViewModel]] = {
```
avelx, line 106: I was expecting the all definition to be wrapped in the Future, but thats fine, as long as we returning correct type, as this would be replaced

avelx (approved):
approved

PR #1809, opened 2023-07-20
Misuv 5914 dynamic stub creation

tharunprasath:


app/utils/Utilities.scala
```diff
@@ -26,4 +27,20 @@ object Utilities {
...
+      case Some(data) => "Gov-Test-Scenario" -> data
+      case _ => "Gov-Test-Scenario" -> ""
+    }
+    headerCarrier.withExtraHeaders(updatedHeaders)
+  }
+
+  // Map list of action names with its test header values
+  def govUKTestHeaderValuesMap(): Map[String, String] = {
+    Map {
+      "uk-property-reporting-method" -> "ISCreated" // UK Property Select reporting method
```
joesheridan, line 44: can we avoid hardcoding the uk-property-reporting-method path?

app/utils/Utilities.scala
```diff
@@ -26,4 +27,20 @@ object Utilities {
     }
   }
 
+  //Checks and adding the value to the test header
+  def checkAndAddTestHeaders(requestPath: String, headerCarrier: HeaderCarrier): HeaderCarrier = {
+    val updatedHeaders = govUKTestHeaderValuesMap.get(requestPath.split('/').last) match {
```
joesheridan: updatedHeaders -> updatedHeader (there is only 1 of them)

app/utils/Utilities.scala
```diff
@@ -17,6 +17,7 @@
 package utils
 
 import play.api.libs.json.{JsObject, Json, Writes}
+import uk.gov.hmrc.http.HeaderCarrier
 
 object Utilities {
```
joesheridan, line 22: probably should be in a file called headers.scala

app/utils/Utilities.scala
joesheridan: I'm also not convinced we have the right name here "ISCreated" - lets discuss..

joesheridan: afterIncomeSourceCreated might be better

joesheridan (approved):

PR #1810, opened 2023-07-20
MISUV-5923-manage-your-details-joint-ticket

simondrugan16:


app/services/IncomeSourceDetailsService.scala
```diff
@@ -224,5 +230,24 @@ class IncomeSourceDetailsService @Inject()(val incomeTaxViewChangeConnector: Inc
...
+        val nonNoneFields = List(
+          Some(address.addressLine1),
+          address.addressLine2,
+          address.addressLine3,
+          address.addressLine4,
+          address.postCode,
+          Some(address.countryCode)
+        ).flatten
+
+        Some(nonNoneFields.mkString("<br>"))
```
avelx, line 247: remove any UI related formatting from the service layer, 
you can do these formatting within views/templates

app/views/incomeSources/manage/ManageSelfEmployment.scala.html
```diff
@@ -0,0 +1,176 @@
...
+        </div>
+        @if(viewModel.itsaHasMandatedOrVoluntaryStatusCurrentYear.getOrElse(throw MissingFieldException("Missing itsaHasMandatedOrVoluntaryStatusCurrentYear field"))
+                && viewModel.latencyDetails.isDefined) {
+            <div class="govuk-summary-list__row">
+                <dt class="govuk-summary-list__key">@getMessage("reporting-method", viewModel.latencyDetails.get.taxYear1.toString, (viewModel.latencyDetails.get.taxYear2).toString)</dt>
+                <dd class="govuk-summary-list__value" id="reporting-method-1">@{
+                    viewModel.latencyDetails.get.latencyIndicator1 match {
+                        case method: String if method == "A" => getMessage("annually")
+                        case method: String if method == "Q" => getMessage("quarterly")
+                        case _ => "Field Missing"
```
avelx, line 142: you would need read this value from the resources like: getMessage, I guess there is translation in Welsh for this

app/views/incomeSources/manage/ManageSelfEmployment.scala.html
```diff
@@ -0,0 +1,176 @@
...
+                    case Some(x) => x match {
+                        case value if value == "cash" => getMessage("cash-accounting")
+                        case value if value == "accruals" => getMessage("traditional-accounting")
+                        case _ => throw new MissingFieldException("Unexpected value for businessAccountingMethod field")
+                    }
+                    case None => messages("incomeSources.generic.unknown")
+                }
+            }</dd>
+        </div>
+        @if(viewModel.itsaHasMandatedOrVoluntaryStatusCurrentYear.getOrElse(throw MissingFieldException("Missing itsaHasMandatedOrVoluntaryStatusCurrentYear field"))
```
avelx, line 134: pls avoid throwing exception in the views

app/views/incomeSources/manage/ManageSelfEmployment.scala.html
```diff
@@ -0,0 +1,176 @@
...
+                    </dd>
+                }
+            </div>
+            <div class="govuk-summary-list__row">
+                <dt class="govuk-summary-list__key">@getMessage("reporting-method", viewModel.latencyDetails.get.taxYear2.toString, (viewModel.latencyDetails.get.taxYear2 + 1).toString)</dt>
+                <dd class="govuk-summary-list__value" id="reporting-method-2">@{
+                    viewModel.latencyDetails.get.latencyIndicator2 match {
+                        case method: String if method == "A" => getMessage("annually")
+                        case method: String if method == "Q" => getMessage("quarterly")
+                        case _ => "Field Missing"
```
avelx, line 159: same, as above, you need to get this value from resource / use getMessage

app/views/incomeSources/manage/ManageSelfEmployment.scala.html
```diff
@@ -0,0 +1,176 @@
...
+                    }
+                }</dd>
+                @if(!viewModel.taxYearTwoCrystallised.get){
```
avelx, line 162: are you sure its safe to call .get in this context?

app/services/IncomeSourceDetailsService.scala
simondrugan16: so move this method to a code block? If I do this i need to add the whole address to the view model. Can I use AddressModel in the view model or do i need to make a put the address into a separate view model called addressViewModel or can I just add each of the fields to the view model separately?

app/views/incomeSources/manage/ManageSelfEmployment.scala.html
simondrugan16: This field should always be "Q" or "A" when coming from the api so should I just remove this line and not have a case _

app/views/incomeSources/manage/ManageSelfEmployment.scala.html
```diff
@@ -0,0 +1,176 @@
...
+            }</dd>
+        </div>
+        <div class="govuk-summary-list__row">
+            <dt class="govuk-summary-list__key">@getMessage("accounting-method")</dt>
+            <dd class="govuk-summary-list__value" id="business-accounting-method">@{
+                viewModel.businessAccountingMethod match {
+                    case Some(x) => x match {
+                        case value if value == "cash" => getMessage("cash-accounting")
+                        case value if value == "accruals" => getMessage("traditional-accounting")
+                        case _ => throw new MissingFieldException("Unexpected value for businessAccountingMethod field")
```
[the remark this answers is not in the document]
simondrugan16: this should only ever come back from the api as "cash" or "accruals" so maybe i just remove this line

app/services/IncomeSourceDetailsService.scala
avelx: This should be fine for next field Option[AddressModel] "travel" all way up to the view as a part of view model; 
You can create a generic method/re-usable block to render address as it was discussed (as its always easy to change this code there, rather than search for this somewhere in the scala code) + you can also promote this code to be a component if team happy to re-use it in other places (but this is more optional I would say);

app/views/incomeSources/manage/ManageSelfEmployment.scala.html
simondrugan16: this should never be empty so can I just do .get?

avelx: in case in this context you are 100% sure that value would be provided, you can call .get; 
but its worth to check if this field really need to be Optional  ?

app/services/IncomeSourceDetailsService.scala
simondrugan16: yeah I guess... I just thought it would be better to add a method to just do it in service as it makes it much easier and there's less to go wrong in service but i will change it

app/views/incomeSources/manage/ManageSelfEmployment.scala.html
simondrugan16: going to remove line

app/views/incomeSources/manage/ManageSelfEmployment.scala.html
simondrugan16: yes I removed the optional

app/views/incomeSources/manage/ManageSelfEmployment.scala.html
simondrugan16: removing line now

app/views/incomeSources/manage/ManageSelfEmployment.scala.html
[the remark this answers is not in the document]
simondrugan16: removing line now

app/views/incomeSources/manage/ManageSelfEmployment.scala.html
```diff
@@ -0,0 +1,188 @@
...
+@changeToValue(currentMethod: String) = @{
+    if (currentMethod == "Q") {
+        "annual"
+    } else {
+        "quarterly"
+    }
+}
+
+@taxYearValue(taxYear: Int) = @{
+    s"${taxYear.toString}-${(taxYear + 1).toString}"
```
avelx, line 71: should work fine without converting .toString

PR #1815, opened 2023-07-24
MISUV-5872 - SCP025 | Add Foreign Property | Build - select reporting method [TESTS]

ayubAz:


it/helpers/agent/ComponentSpecBase.scala
```diff
@@ -330,6 +330,11 @@ trait ComponentSpecBase extends TestSuite with CustomMatchers
       val formData = form.toFormMap.map { case (k, v) => (k -> Seq(v.getOrElse(""))) }
       post(s"/income-sources/add/uk-property-reporting-method?id=$testPropertyIncomeId", additionalCookies = additionalCookies)(formData)
     }
+
+    def postAddForeignPropertyReportingMethod(form: AddForeignPropertyReportingMethodForm)(additionalCookies: Map[String, String] = Map.empty): WSResponse = {
+      val formData = form.toFormMap.map { case (k, v) => (k -> Seq(v.getOrElse(""))) }
```
avelx, line 335: maybe worth to raise and error instead of calling .getOrElse ?
I think in this context should be fine to call .get

app/controllers/incomeSources/add/CheckForeignPropertyDetailsController.scala
```diff
@@ -48,7 +48,7 @@ class CheckForeignPropertyDetailsController @Inject()(val authenticate: Authenti
   extends ClientConfirmedController with FeatureSwitching with I18nSupport {
 
   def show(origin: Option[String] = None): Action[AnyContent] = Action {
-    Ok("Check Foreign Property Details - WIP")
+    Redirect(controllers.incomeSources.add.routes.ForeignPropertyReportingMethodController.show("XAIS00000000004"))
```
avelx, line 51: as discuss with QAs, we need to read incomeSourceId from the query string parameters , id key

avelx (approved):
approved

PR #1824, opened 2023-07-27
[WIP]: SCP025 |Manage SE, UK & foreign property | Build -Joint ticket - "Confirm you want to report <annually/quarterly> for the TY1 to TY2" pages (3 out of 4)

ayubAz:


app/controllers/incomeSources/manage/ConfirmReportingMethodSharedController.scala
```diff
@@ -0,0 +1,165 @@
...
+          )
+        case _ =>
+          Logger("application")
+            .error(s"[ConfirmReportingMethodSharedController][handleRequest]: Could not find property or business with incomeSourceId: $id")
+          itvcErrorHandler.showInternalServerError()
+      }
+    )
+  }
+
+  private def getTaxYearStartYearEndYear(year: String): Option[TaxYear] = {
```
avelx, line 145: this function is something we should move into the model or into another service;
Its seems the best candidate for this is DateService as there might be already similar method there + this need to be unit tested

app/controllers/incomeSources/manage/ConfirmReportingMethodSharedController.scala
```diff
@@ -0,0 +1,165 @@
...
+
+  private def getReportingMethodKey(reportingMethod: String): Option[String] = {
+    reportingMethod.toLowerCase match {
+      case "annual" => Some("annual")
+      case "quarterly" => Some("quarterly")
+      case _ => None
+    }
+  }
+}
+
```
avelx, line 164: this need to be move into the models, as we dealing with taxation across application and seems reasonable  to have exposed this to app level; check if we don't have this already defined somewhere

app/controllers/incomeSources/manage/ConfirmReportingMethodSharedController.scala
```diff
@@ -0,0 +1,165 @@
...
+    def isValidYear(year: String) = year.length == 4 && year.forall(_.isDigit)
+
+    year.split('-') match {
+      case Array(yearOne, yearTwo) if isValidYear(yearOne) && isValidYear(yearTwo) =>
+        Some(TaxYear(yearOne, yearTwo))
+      case _ => None
+    }
+  }
+
+  private def getReportingMethodKey(reportingMethod: String): Option[String] = {
```
avelx, line 156: not sure why we need this method in term of the usage;
also its trivial duplication of the next logic:
  
Set("annual", "quarterly").find(_.toLowerCase == "w")

app/controllers/incomeSources/manage/ConfirmReportingMethodSharedController.scala
```diff
@@ -0,0 +1,165 @@
...
+    lazy val isSoleTraderBusiness = user.incomeSources.businesses.exists(business =>
+      business.incomeSourceId.contains(id) && !business.isCeased)
+
+    lazy val isForeignProperty = user.incomeSources.properties.exists(property =>
+      property.incomeSourceId.contains(id) && property.isForeignProperty && !property.isCeased)
+
+    lazy val isUKProperty = user.incomeSources.properties.exists(property =>
+      property.incomeSourceId.contains(id) && property.isUkProperty && !property.isCeased)
+
+    Future(
```
avelx, line 96: we have to use .recover on the future to handle Future failure as well

app/views/incomeSources/manage/ConfirmForeignPropertyReportingMethod.scala.html
```diff
@@ -0,0 +1,69 @@
...
+        taxYearEndYear: String,
+        reportingMethodKey: String
+)(implicit messages: Messages, user: auth.MtdItUser[_])
+
+@getMessage(key: String, args: String*) = @{
+    messages(s"incomeSources.manage.propertyReportingMethod.$key", args: _*)
+}
+
+@mainTemplate(pageTitle = getMessage(s"heading.$reportingMethodKey"),
+    backUrl = Some(""),
```
avelx, line 40: we need backUrl in all 3 views

app/controllers/incomeSources/manage/ConfirmReportingMethodSharedController.scala
```diff
@@ -0,0 +1,165 @@
...
+      (isEnabled(IncomeSources), getTaxYearStartYearEndYear(taxYear), getReportingMethodKey(changeTo)) match {
+        case (false, _, _) =>
+          Ok(
+            customNotFoundErrorView()
+          )
+        case (_, _, None) =>
```
avelx, line 102: we might don't really need this case, as this all will be covered by the last case defined in th

app/models/incomeSourceDetails/TaxYear.scala
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.incomeSourceDetails
+
+case class TaxYear(startYear: String, endYear: String)
```
avelx, line 19: assuming you are doing all required validation and converting string into your custom type, I think this would make more sense to keep startYear and endYear as Integer + you might need to add:
required(startYear + 1 == endYear), as this is more accurate description of the data for this new type

test/models/TaxYearModelSpec.scala
```diff
@@ -0,0 +1,67 @@
...
+
+package models
+
+import models.incomeSourceDetails.TaxYear
+import testUtils.UnitSpec
+
+class TaxYearModelSpec extends UnitSpec {
+
+
+  "TaxYear.getTaxYearStartYearEndYear" when {
```
avelx, line 25: its seems a good use case to use scalaCheck here, as this will shorten your code, but really up to you

app/models/incomeSourceDetails/TaxYear.scala
```diff
@@ -16,20 +16,29 @@
 
 package models.incomeSourceDetails
 
-import java.lang.reflect.UndeclaredThrowableException
 import scala.util.Try
 
-case class TaxYear(startYear: Int, endYear: Int)
+case class TaxYear(startYear: String, endYear: String)
```
avelx, line 21: if we can keep this, as this more accurate description of this type
`case class TaxYear(startYear: Int, endYear: Int)`

avelx (approved):
approve

PR #1825, opened 2023-07-27
MISUV-5921 SCP025 | Manage SE, UK&foreign property | Build - ("Obligations" pages), joint ticket

danielmaltz1:


app/views/incomeSources/manage/ManageObligations.scala.html
```diff
@@ -0,0 +1,204 @@
...
+
+@getMessage(key: String, args: String*) = @{
+    messages(s"business-added.$key", args: _*)
+}
+
+@getUniqueMessage(key: String, args: String*) = @{
+    messages(s"incomeSources.add.manageObligations.$key", args: _*)
+}
+
+@getBusinessName(name: String, mode: String) = @{
```
avelx, line 58: mode - parameter seems to be rendandant

app/views/incomeSources/manage/ManageObligations.scala.html
```diff
@@ -0,0 +1,204 @@
...
+            </div>
+        </div>
+
+        <h2 class="govuk-label-wrapper">
+            <label class="govuk-label govuk-label--l govuk-!-margin-top-8">
+                @getUniqueMessage("heading")
+            </label>
+        </h2>
+
+        @if(sources.quarterlyObligationsDatesYearOne.nonEmpty) {
```
avelx, line 105: this might be really better to use .forEach instead here, 
usage of .get or .head is really discouraged, this make code less flexible

app/views/incomeSources/manage/ManageObligations.scala.html
```diff
@@ -0,0 +1,204 @@
...
+                <table class="govuk-table">
+                    <thead class="govuk-table__head">
+                        <tr class="govuk-table__row">
+                            <th scope="col" class="govuk-table__header">@(getMessage("table.heading-1") + " " + sources.quarterlyObligationsDatesYearOne.head.inboundCorrespondenceFrom.getYear + " " + getMessage("to") + " " + (sources.quarterlyObligationsDatesYearOne.head.inboundCorrespondenceFrom.getYear + 1))</th>
+                            <th scope="col" class="govuk-table__header">@getMessage("table.heading-2")</th>
+                        </tr>
+                    </thead>
+
+                    <tbody class="govuk-table__body">
+                    @sources.quarterlyObligationsDatesYearOne.map { x =>
```
avelx, line 119: preferable to use .foreach over .map as there is not return value here, but we evaluating view/render final HTML here; not a mistake, but easy to read

app/views/incomeSources/manage/ManageObligations.scala.html
danielmaltz1: I need to check if it's empty to see if I need to render the table at all. Displaying the heads of the table depends on whether there is any data, it isn't done for every piece of data

app/views/incomeSources/manage/ManageObligations.scala.html
```diff
@@ -0,0 +1,204 @@
...
+            <table class="govuk-table">
+                <thead class="govuk-table__head">
+                    <tr class="govuk-table__row">
+                        <th scope="col" class="govuk-table__header">@getMessage("table.heading-1")</th>
+                        <th scope="col" class="govuk-table__header">@getMessage("table.heading-2")</th>
+                    </tr>
+                </thead>
+
+                <tbody class="govuk-table__body">
+                @sources.finalDeclarationDates.map { x =>
```
avelx, line 187: again, preferable to use .foreach, and you might move this to  re-usable code block is patterns is it repeating pattern

app/views/incomeSources/manage/ManageObligations.scala.html
```diff
@@ -0,0 +1,204 @@
...
+        p: p,
+        link: link,
+        hr: hr,
+        govukTable: GovukTable
+)
+
+
+@(
+        sources: ObligationsViewModel,
+        mode: String,
```
avelx, line 39: there is no actual usage of mode in the view, so this is really something redundant

app/controllers/incomeSources/manage/ManageObligationsController.scala
```diff
@@ -0,0 +1,222 @@
...
+            s"[BusinessAddedObligationsController][handleRequest] - invalid tax year provided")
+          if (isAgent) Future(itvcErrorHandlerAgent.showInternalServerError())
+          else Future(itvcErrorHandler.showInternalServerError())
+        }
+      }
+    }
+  }
+
+  def getIncomeSourceId(mode: String, id: String)(implicit user: MtdItUser[_]): Either[Throwable, String] = {
+    mode match {
```
avelx, line 195: pls avoid use of String type to model enumeration, this is not best practice;
under Scala 2.x.x we use you should use sealed traits for this, this new type can be used within a view if you need this; + there are other benefit provided by compiler to verify if there are missed cases in your logic, this is not achievable when you use String type ..

app/controllers/incomeSources/manage/ManageObligationsController.scala
```diff
@@ -0,0 +1,222 @@
...
+      case "SE" => Right(id)
+      case "UK" => user.incomeSources.properties.find(x => x.isUkProperty) match {
+        case Some(value) => Right(value.incomeSourceId.getOrElse(""))
```
avelx, line 198: its unlikely we would have anything when incomeSourceId is an empty string, this should lead to error right away

app/controllers/incomeSources/manage/ManageObligationsController.scala
```diff
@@ -0,0 +1,222 @@
...
+        case None => Left(new Error("Failed to find incomeSource Id"))
+      }
+      case "FP" => user.incomeSources.properties.find(x => x.isForeignProperty) match {
+        case Some(value) => Right(value.incomeSourceId.getOrElse(""))
+        case None => Left(new Error("Failed to find incomeSource Id"))
```
avelx, line 203: pls add incomSourceId details to the error as this would help to investigate this error

app/views/incomeSources/manage/ManageObligations.scala.html
danielmaltz1: replacing map with foreach has meant the rows aren't rendered, is there another step I need to add?

app/controllers/incomeSources/manage/ManageObligationsController.scala
danielmaltz1: incomeSourceId is only passed in the url if it is the self employment page, I've already covered the case where there is no SE with the supplied id above with "      if (mode == "SE" && !user.incomeSources.businesses.exists(x => x.incomeSourceId.contains(incomeSourceId))) {
        Logger("application").error("

app/controllers/incomeSources/manage/ManageObligationsController.scala
danielmaltz1: If we are in mode "UK" or "FP" (so either of the property pages) then no incomeSourceId is supplied so I cannot output it

app/controllers/incomeSources/manage/ManageObligationsController.scala
avelx: there are few option to avoid calling .getOrElse here, either to add guard Or to map over .incomeSourceId after find in line 197, its really up to you how you prefer to do this,

app/controllers/incomeSources/manage/ManageObligationsController.scala
```diff
@@ -0,0 +1,238 @@
...
+      handleRequest(
+        mode = SelfEmployment,
+        isAgent = false,
+        taxYear,
+        changeTo,
+        id
+      )
+  }
+
+  def showAgentSelfEmployment(changeTo: String, taxYear: String, id: String): Action[AnyContent] = Authenticated.async {
```
avelx, line 81: rename id to incomeSourceId as this is not exposed to UI we free to use required naming convention; this would simplify code support

app/models/incomeSourceDetails/TaxYear.scala
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.incomeSourceDetails
+
+case class TaxYear(startYear: String, endYear: String)
```
avelx, line 19: there was a few suggestion with respect to this model ( double check with @ayubAz  what is his plan to apply changes etc)
startYear / endYear - to be converted to Int and we might need to add 
required(startYear + 1 == endYear, "Incorrect taxYear definition ..") to definition

app/views/incomeSources/manage/ManageObligations.scala.html
```diff
@@ -0,0 +1,210 @@
...
+
+    @getMessage(key: String, args: String*) = @{
+        messages(s"business-added.$key", args: _*)
+    }
+
+    @getUniqueMessage(key: String, args: String*) = @{
+        messages(s"incomeSources.add.manageObligations.$key", args: _*)
+    }
+
+    @getBusinessName(name: String) = @{
```
avelx, line 58: I don't think you really need lines 59 and 60, you can remove these;
String is usually not the best choice to model enumeration, but I guess it really your choice here

app/controllers/incomeSources/manage/ManageObligationsController.scala
```diff
@@ -0,0 +1,238 @@
...
+    else {
+      val backUrl: String = if (isAgent) controllers.incomeSources.manage.routes.ManageConfirmController.showAgent().url else controllers.incomeSources.manage.routes.ManageConfirmController.show().url
+      val postUrl: Call = if (isAgent) controllers.incomeSources.manage.routes.ManageObligationsController.agentSubmit() else controllers.incomeSources.manage.routes.ManageObligationsController.submit()
+
+
+      if (mode == SelfEmployment && !user.incomeSources.businesses.exists(x => x.incomeSourceId.contains(incomeSourceId))) {
+        showError(isAgent, s"unable to find incomeSource by id: $incomeSourceId")
+      }
+      else {
+        val addedBusinessName: String = mode match {
```
avelx, line 164: Ideally we should reduce number of lines in this method as it about 40 lines, which is okish ..
Extracting method to return addedBusinessName would be best candidate for this simplification;

avelx (approved):

PR #1837, opened 2023-08-04
MISUV-5923-manage-uk-and-foreign-property-pages

simondrugan16:


app/models/incomeSourceDetails/BusinessDetailsModel.scala
```diff
@@ -28,10 +30,14 @@ case class BusinessDetailsModel(incomeSourceId: Option[String],
                                 cessation: Option[CessationModel],
                                 cashOrAccruals: Option[String] = None,
                                 address: Option[AddressModel] = None,
-                                latencyDetails: Option[LatencyDetails] = None) {
+                                latencyDetails: Option[LatencyDetails]) extends LatencyInformation {
```
avelx, line 33: BusinessDetailsModel as far as I understand is part of response from the relevant API; so this is "given" and we have not much choice, so I am not sure if this is a good idea to manipulate with definition of this model;
from the other hand you are free to use this model to create any other required view model base on the requirement dictated in the UI logic etc

app/views/incomeSources/manage/ManageSelfEmployment.scala.html
```diff
@@ -108,22 +109,24 @@
...
-            <dt class="govuk-summary-list__key">@getMessage("business-address")</dt>
-            <dd class="govuk-summary-list__value" id="business-address">@{
-                getLongAddressFromBusinessAddressDetails(viewModel.address) match {
-                    case Some(x) => HtmlFormat.raw(x)
-                    case None => messages("incomeSources.generic.unknown")
-                }
-            }</dd>
-        </div>
+    <dl class="govuk-summary-list" id="manage-details-table">
+        @{if (viewModel.incomeSourceType == SELF_EMPLOYMENT)} {
```
avelx, line 113: you are free to use or "if/else" Or pattern matching, up to you what you prefer

PR #1853, opened 2023-08-15
MISUV-6011: SCP025| Add SE, property| Consolidate View Template for Add Business/property start date and check page

ayubAz:


app/controllers/incomeSources/add/AddIncomeSourceStartDateCheckController.scala
```diff
@@ -0,0 +1,288 @@
...
+      (getStartDate(incomeSourceType), getCalls(isAgent, incomeSourceType)) match {
+        case (None, _) =>
+          Logger("application").error(s"[AddIncomeSourceStartDateCheckController][handleSubmitRequest]: " +
+            s"failed to get start date from session for incomeSourceType: $incomeSourceType")
+          Future(
+            getErrorHandler(isAgent).showInternalServerError()
+          )
+        case (Some(startDate), (backCall, postAction, successCall)) =>
+
+          form(incomeSourceType.addIncomeSourceStartDateCheckMessagesPrefix)
```
avelx, line 179: if we can move this specific code, related to form creation into separate method ?

app/controllers/incomeSources/add/AddIncomeSourceStartDateCheckController.scala
```diff
@@ -0,0 +1,288 @@
...
+      }
+    }
+  } recover {
+    case ex: Exception =>
+      Logger("application").error(s"[AddIncomeSourceStartDateCheckController][handleSubmitRequest]: " +
+        s"Error getting AddIncomeSourceStartDateCheck page: ${ex.getMessage}")
+      getErrorHandler(isAgent).showInternalServerError()
+  }
+
+  private def getStartDate(incomeSourceType: IncomeSourceType)(implicit user: MtdItUser[_]): Option[String] = {
```
avelx, line 234: few things can be done here:
* you can wrap results into Future as this would be extracted from the mongo (this changes is optional);
* attempt to convert String into LocalDate, as you are using LocalDate down in the context; so result can be something like String => Either[Throwable, LocalDate]

avelx: we are planning work related to mongo in the future, so we would read user stored session data from mongo in the future; in order to support this transition in we would need these method to return Future, but this is really something optional at the moment;

avelx (approved):
approved

app/controllers/incomeSources/add/AddIncomeSourceStartDateController.scala
```diff
@@ -195,54 +200,56 @@ class AddIncomeSourceStartDateController @Inject()(authenticate: AuthenticationP
...
           controllers.incomeSources.add.routes.AddIncomeSourceController.showAgent(),
           controllers.incomeSources.add.routes.AddIncomeSourceStartDateController.submitForeignPropertyAgent,
-          controllers.incomeSources.add.routes.ForeignPropertyStartDateCheckController.showAgent(),
-          ForeignProperty.messagesPrefix
+          controllers.incomeSources.add.routes.AddIncomeSourceStartDateCheckController.showForeignPropertyAgent
         )
     }
   }
 
+  private def getMessagesPrefix(incomeSourceType: IncomeSourceType): String = {
```
avelx, line 245: looks like redundant method as you can do next call: 
`
private def getMessagesPrefix(incomeSourceType: IncomeSourceType): String = incomeSourceType.addIncomeSourceStartDateMessagesPrefix
`

PR #1862, opened 2023-08-21
WIP Fix compiler warnings

danielmaltz1:


app/audit/models/ChargeSummaryAudit.scala
```diff
@@ -98,7 +98,7 @@ case class ChargeSummaryAudit(mtdItUser: MtdItUser[_], docDateDetail: DocumentDe
   }
 
   private val interestPeriod: Option[String] = (docDateDetail.documentDetail.interestFromDate, docDateDetail.documentDetail.interestEndDate) match {
-    case (Some(fromDate), Some(endDate)) => Some(fromDate + " to " + endDate)
+    case (Some(fromDate), Some(endDate)) => Some(fromDate.toString + " to " + endDate.toString)
```
avelx, line 101: string interpolation will be shorted in this case, ie:
Some(s"$fromDate to $endDate")

app/controllers/PaymentController.scala
```diff
@@ -56,7 +56,7 @@ class PaymentController @Inject()(val checkSessionTimeout: SessionTimeoutPredica
       case Some(utr) =>
         payApiConnector.startPaymentJourney(utr, paymentAmountInPence, isAgent).map {
           case model: PaymentJourneyModel => Redirect(model.nextUrl)
-          case _: PaymentJourneyErrorResponse => throw new Exception("Failed to start payments journey due to downstream response")
+          case _: PaymentJourneyResponse => throw new Exception("Failed to start payments journey due to downstream response")
```
avelx, line 59: we should match on PaymentJourneyErrorResponse(status, message)
and provide error details in exception like, 
Exception(s"Failed to start payments journey due to downstream response: $status - $message")

app/controllers/incomeSources/add/AddIncomeSourceStartDateCheckController.scala
```diff
@@ -199,6 +201,10 @@ class AddIncomeSourceStartDateCheckController @Inject()(authenticate: Authentica
           )
       case (Some(form.responseYes), _) =>
         Redirect(successCall)
+      case (Some(_), _) => Logger("application").error(s"[AddIncomeSourceStartDateCheckController][handleValidForm] - Unexpected response")
```
avelx, line 204: can we change log messages below so they would not be identical and we need details if this is an Agent or Individual

app/controllers/incomeSources/add/IncomeSourcesAccountingMethodController.scala
```diff
@@ -116,7 +119,7 @@ class IncomeSourcesAccountingMethodController @Inject()(val authenticate: Authen
           routes.AddBusinessAddressController.show().url
       case UkProperty.key =>
           routes.AddIncomeSourceStartDateCheckController.show(UkProperty.key, isAgent, isChange = false).url
-      case ForeignProperty.key =>
+      case _ =>
```
avelx, line 122: I would say this that we should have conversion from String => IncomeSourceType before pattern matching and raise failure if conversion is not succesfull

app/controllers/incomeSources/add/IncomeSourcesAccountingMethodController.scala
```diff
@@ -104,41 +107,56 @@ class IncomeSourcesAccountingMethodController @Inject()(val authenticate: Authen
     )(user, messages)))
   }
 
+  def convertToIncomeSourceType(incomeSource: String): Either[Throwable, IncomeSourceType] = {
+    incomeSource match{
+      case SelfEmployment.key => Right(SelfEmployment)
+      case UkProperty.key => Right(SelfEmployment)
+      case ForeignProperty.key => Right(ForeignProperty)
+      case _ => Left(new Error)
```
avelx, line 115: if you can provide error details in the error, as this method contract against "external world" would be something like: 
  Success Or Failed because of ...

app/controllers/incomeSources/add/IncomeSourcesAccountingMethodController.scala
```diff
@@ -104,41 +107,56 @@ class IncomeSourcesAccountingMethodController @Inject()(val authenticate: Authen
...
+      case Right(incomeSource) =>
+        val backUrl = incomeSource match {
+          case SelfEmployment =>
+            if (isAgent)
+              routes.AddBusinessAddressController.showAgent().url
+            else
+              routes.AddBusinessAddressController.show().url
+          case UkProperty =>
+            routes.AddIncomeSourceStartDateCheckController.show(UkProperty.key, isAgent, isChange = false).url
+          case _ =>
```
avelx, line 139: instead of default case you should be able to put: ForeignProperty,
as this type is under sealed trait compile would not raise warning in this case;

danielmaltz1: Done

app/controllers/incomeSources/add/IncomeSourcesAccountingMethodController.scala
```diff
@@ -104,41 +108,56 @@ class IncomeSourcesAccountingMethodController @Inject()(val authenticate: Authen
     )(user, messages)))
   }
 
+  def convertToIncomeSourceType(incomeSource: String): Either[Exception, IncomeSourceType] = {
```
avelx, line 111: * there is already such conversion specified in /Users/hmrc/devcore/itvc/income-tax-view-change-frontend/app/enums/IncomeSourceJourney/IncomeSourceType.scala
if we can use single method instead of duplicating;

danielmaltz1: Ok, replaced. This simplified a lot

avelx (approved):
approved

PR #1868, opened 2023-08-24
MISUV-6004: SCP025 | Manage SE, UK/foreign property | Reporting method change not saved error

ayubAz:


app/controllers/incomeSources/add/AddIncomeSourceStartDateCheckController.scala
```diff
@@ -58,106 +58,94 @@ class AddIncomeSourceStartDateCheckController @Inject()(authenticate: Authentica
...
+                                  isChange: Boolean)
                                  (implicit mtdItUser: MtdItUser[_]): Future[Result] = {
 
     val messagesPrefix = incomeSourceType.addStartDateCheckMessagesPrefix
-
     Future.successful(
       if (isEnabled(IncomeSources))
-        (getAndValidateStartDate(incomeSourceType), getCalls(isAgent, isUpdate, incomeSourceType)) match {
-          case (Right(startDate), (backCall, postAction, successCall)) =>
+        getAndValidateStartDate(incomeSourceType) match {
```
avelx, line 120: if we can change signature for this method pls as this seems not logical to return Either[Throwable, String];
Please use => Either[Throwable, LocalDate]

avelx (approved):
approved

PR #1874, opened 2023-08-31
MISUV-6204-property-businesses-ceased

simondrugan16:


app/models/incomeSourceDetails/viewmodels/CeasedBusinessDetailsViewModel.scala
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+package models.incomeSourceDetails.viewmodels
+
+import enums.IncomeSourceJourney.IncomeSourceType
+
+import java.time.LocalDate
+
+case class CeasedBusinessDetailsViewModel(tradingName: Option[String],
```
avelx, line 23: I think tradingName has some length limitation, so you can apply these as well on the type level (this is optional);

test/testConstants/PropertyDetailsTestConstants.scala
```diff
@@ -115,7 +127,16 @@ object PropertyDetailsTestConstants {
     firstAccountingPeriodEndDate = None,
     incomeSourceType = Some(foreignIncomeType),
     tradingStartDate = Some(testStartDate),
-    cessation = Some(testCessation)
+    cessation = Some(testPropertyCessation)
+  )
+
+  val ceasedForeignPropertyDetailsCessation2 = PropertyDetailsModel(
```
avelx, line 133: naming convention suggestion, this can be something like:
ceasedForeignPropertyDetailsCessation2 => ceasedForeignPropertyDetailsCessationNoEndDate

app/views/incomeSources/add/AddIncomeSources.scala.html
```diff
@@ -45,10 +47,15 @@
     messages("incomeSources.add.addIncomeSources.unknown")
 }
 
-@ukProperty = @{ sources.ukProperty }
-@foreignProperty = @{ sources.foreignProperty }
-@ceasedBusinesses = @{ sources.ceasedBusinesses }
-@soleTraderBusinesses = @{ sources.soleTraderBusinesses }
+@ukProperty = @{
```
avelx, line 50: these code blocks redundant as these are one line of code,

app/services/IncomeSourceDetailsService.scala
```diff
@@ -224,5 +205,30 @@ class IncomeSourceDetailsService @Inject()(val incomeTaxViewChangeConnector: Inc
       }
     }.toEither
   }
+
+  private def getCeasedBusinesses(sources: IncomeSourceDetailsModel): List[CeasedBusinessDetailsViewModel] = {
```
avelx, line 209: we usually return Either[Throwable, ViewModel] from the service, and log that error in the the caller, which is usually controller

app/models/incomeSourceDetails/viewmodels/CeasedBusinessDetailsViewModel.scala
avelx: apply required on the level on level of the case class definition, something like:
`
case class CeasedBusinessDetailsViewModel(tradingName: Option[String], ... )
   required(tradingName.length < 35 , "Trading name must be ... ")
`
this would guaranty that view model constructed would be always correct, you just need to handle this exception correctly, but this is really something optional

avelx (approved):
approved

PR #1876, opened 2023-09-01
MISUV-6203 added SessionService service to ease mongo session storage…

joesheridan:
… migration


app/services/SessionService.scala
```diff
@@ -0,0 +1,33 @@
...
+
+import auth.MtdItUser
+
+import javax.inject.{Inject, Singleton}
+import scala.concurrent.{ExecutionContext, Future}
+
+@Singleton
+class SessionService @Inject()() {
+
+  def get(key: String)(implicit user: MtdItUser[_], ec: ExecutionContext): Future[Either[Throwable, Option[String]]] = {
```
avelx, line 27: what about set method?

PR #1881, opened 2023-09-06
MISUV-6238: fix - favicon path resolution

SiddharthAtos:


conf/app.routes
```diff
@@ -1,8 +1,5 @@
 # microservice specific routes
 
-# Play frontend hmrc routes
-->          /hmrc-frontend                                                hmrcfrontend.Routes
```
EdwardDennis, line 4: This seems to be defined in prod.routes as well, can you remove if it's not needed?

SiddharthAtos: We need that. 
The route evaluation was using the app.routes for  `hmrc-frontend` as it was first in sequence and It comes under the V&C frontend route.

PR #1882, opened 2023-09-07
MISUV-6238 fix browser tab icon

SiddharthAtos:


app/views/layouts/unifiedLayout.scala.html
```diff
@@ -143,6 +143,7 @@
         bodyEndBlock = None,
         headerBlock = Some(siteHeader),
         footerBlock = Some(footerLinks()),
-        scriptsBlock = Some(scripts())
+        scriptsBlock = Some(scripts()),
+        assetPath = Some(uk.gov.hmrc.hmrcfrontend.controllers.routes.Assets.at("govuk").url)
```
EdwardDennis, line 147: This will need to be added in the if logic as well or it won't work on the home page (where main class is defined)

EdwardDennis (approved):

PR #1891, opened 2023-09-13
Misuv 6186 SCP025| Add SE, property | Refactoring ticket for moving to Mongo (add journey)

tharunprasath:


app/utils/IncomeSourcesUtils.scala
```diff
@@ -40,6 +43,35 @@ trait IncomeSourcesUtils extends FeatureSwitching {
     }
   }
 
+  def newWithIncomeSourcesRemovedFromSession(redirect: Result, sessionService: SessionService, errorRedirect: Result)(implicit user: MtdItUser[_], ec: ExecutionContext): Future[Result] = {
+    val incomeSourcesSessionKeys = Seq(
```
avelx, line 47: we do have SessionKeys with all names of keys we have, so this would duplication;
if you need list of all key, you should define this in the SessionKeys object

app/utils/IncomeSourcesUtils.scala
```diff
@@ -71,49 +103,49 @@ object IncomeSourcesUtils {
 
   case class MissingKey(msg: String)
 
-  def getBusinessDetailsFromSession(implicit user: MtdItUser[_]): Either[Throwable, CheckBusinessDetailsViewModel] = {
+  def getBusinessDetailsFromSession(sessionService: SessionService)(implicit user: MtdItUser[_], ec: ExecutionContext): Future[CheckBusinessDetailsViewModel] = {
```
avelx, line 106: this method has no direct relationship to IncomeSources, so not sure why we moving this under IncomeSourceUtils

avelx, line 106: Please return Future[Either[Throwable,CheckBusinessDetailsViewModel ]] instead

app/utils/IncomeSourcesUtils.scala
tharunprasath: I have revamped the existing method withIncomeSourcesRemovedFromSession to handle the sessionService flow to remove from Mongo, Once all the IncomeSource sessions are moved to the sessionService we need to remove the duplicated method withIncomeSourcesRemovedFromSession.

app/utils/IncomeSourcesUtils.scala
tharunprasath: Moved to CheckBusinessDetailsController

app/utils/IncomeSourcesUtils.scala
avelx: in incomeSourcesSessionKeys sequence, you should use variable defined in SessionKeys

app/controllers/incomeSources/add/CheckBusinessDetailsController.scala
```diff
@@ -180,4 +173,50 @@ class CheckBusinessDetailsController @Inject()(val checkBusinessDetails: CheckBu
...
+
+    result.flatMap {
+      case (Right(businessName), Right(businessStartDate), Right(Some(businessTrade)),
+      Right(Some(businessAddressLine1)), Right(Some(businessAccountingMethod)), Right(accountingPeriodEndDate),
+      Right(businessAddressLine2), Right(businessAddressLine3), Right(businessAddressLine4), Right(businessPostalCode),
+      Right(businessCountryCode), Right(incomeSourcesAccountingMethod)) =>
+        Future.successful(CheckBusinessDetailsViewModel(
+          businessName = businessName,
+          businessStartDate = businessStartDate.map(LocalDate.parse(_)),
+          accountingPeriodEndDate = accountingPeriodEndDate.map(LocalDate.parse(_)).get,
```
avelx, line 206: if accountingPeriodEndDate is mandatory field, we need to handle parsing in a right way
Please don't call .get directly over Option

app/utils/IncomeSourcesUtils.scala
tharunprasath: Changed to Future[Either[Throwable,CheckBusinessDetailsViewModel ]]

app/utils/IncomeSourcesUtils.scala
tharunprasath: Changed to use SessionKeys.

app/controllers/incomeSources/add/CheckBusinessDetailsController.scala
tharunprasath: Handled the parsing for mandatory field

app/controllers/incomeSources/add/CheckBusinessDetailsController.scala
```diff
@@ -180,4 +175,50 @@ class CheckBusinessDetailsController @Inject()(val checkBusinessDetails: CheckBu
...
+          businessAddressLine2 = businessAddressLine2,
+          businessAddressLine3 = businessAddressLine3,
+          businessAddressLine4 = businessAddressLine4,
+          businessPostalCode = businessPostalCode,
+          businessCountryCode = businessCountryCode,
+          incomeSourcesAccountingMethod = incomeSourcesAccountingMethod,
+          cashOrAccrualsFlag = businessAccountingMethod,
+          skippedAccountingMethod = skipAccountingMethod
+        )))
+      case ex => Future.failed(MissingSessionKey("[IncomeSourcesUtils][getBusinessDetailsFromSession]" + ex))
```
avelx, line 220: you need to return Left with exception wrap in it on this occasion

app/controllers/incomeSources/add/CheckBusinessDetailsController.scala
```diff
@@ -180,4 +175,50 @@ class CheckBusinessDetailsController @Inject()(val checkBusinessDetails: CheckBu
         }
   }
 
+  private def getBusinessDetailsFromSession(implicit user: MtdItUser[_], ec: ExecutionContext): Future[Either[Throwable, CheckBusinessDetailsViewModel]] = {
+    val userActiveBusinesses: List[BusinessDetailsModel] = user.incomeSources.businesses.filterNot(_.isCeased)
+    val skipAccountingMethod: Boolean = userActiveBusinesses.isEmpty
+
+    val result = for {
+      businessName <- sessionService.get(SessionKeys.businessName)
```
avelx, line 183: this seems quite strange approach as you keep querying SessionService regardless there was an error or not,
if will be fine to return first error occurred ; you can re-use foldLeft logic we used in this PR already

avelx:
@tharunprasath  - if you can resolve merge conflicts as well, thanks

app/controllers/incomeSources/add/CheckBusinessDetailsController.scala
```diff
@@ -180,4 +175,57 @@ class CheckBusinessDetailsController @Inject()(val checkBusinessDetails: CheckBu
...
+          val finalMap = m + (fieldName -> optVal)
+          Right(finalMap)
+        case (_, _) =>
+          futureAcc
+      }
+    }
+
+    for {
+      res <- xs
+    } yield res match { // => Add extra validation for the date parsing + others where required
```
avelx, line 206: I was hoping we would go data validation in each of fields we calling .get when constructing model, but assuming this is a temp code and we would use something provided by relevant service, we might skip this on this occasion

avelx (approved):
approved

PR #1902, opened 2023-09-21
MISUV-6342 SCP025| add SE, property| Obligations controller - pages refactoring before the TxM events

danielmaltz1:


app/controllers/incomeSources/add/IncomeSourceAddedController.scala
```diff
@@ -0,0 +1,164 @@
...
+        } yield (startDate, None)
+    }
+  }
+
+  private def handleRequest(isAgent: Boolean, incomeSourceId: String, incomeSourceType: IncomeSourceType)(implicit user: MtdItUser[_], ec: ExecutionContext): Future[Result] = {
+    withIncomeSourcesFS {
+      val backUrl = getBackUrl(incomeSourceId, isAgent, incomeSourceType)
+      getIncomeSourceData(incomeSourceType, incomeSourceId) match {
+        case Some(incomeSourceData) =>
+          val showPreviousTaxYears: Boolean = incomeSourceData._1.isBefore(dateService.getCurrentTaxYearStart())
```
avelx, line 117: avoid resolve Tuple type in the ways above/below: =>
._1 or ._2
you can alway resolve these values in the variable with semantical meaning defined by the actual variable name

app/controllers/incomeSources/add/IncomeSourceAddedController.scala
```diff
@@ -0,0 +1,164 @@
...
+            }
+            case ForeignProperty => nextUpdatesService.getObligationsViewModel(incomeSourceId, showPreviousTaxYears) map { viewModel =>
+              Ok(obligationsView(viewModel, backUrl, isAgent = isAgent, incomeSourceType = ForeignProperty))
+            }
+          }
+        case None => Logger("application").error(
+          s"[IncomeSourceAddedController][handleRequest] - unable to find incomeSource by id: $incomeSourceId")
+          if (isAgent) Future(itvcErrorHandlerAgent.showInternalServerError())
+          else Future(itvcErrorHandler.showInternalServerError())
+      }
```
avelx, line 139: if you can add recover for the resulting Future of this method, as we should handle critical error as well by logging them etc

app/controllers/incomeSources/add/IncomeSourceAddedController.scala
```diff
@@ -0,0 +1,164 @@
...
+        if (isAgent) controllers.incomeSources.add.routes.UKPropertyReportingMethodController.showAgent _ else
+          controllers.incomeSources.add.routes.UKPropertyReportingMethodController.show _
+      case ForeignProperty =>
+        if (isAgent) controllers.incomeSources.add.routes.ForeignPropertyReportingMethodController.showAgent _ else
+          controllers.incomeSources.add.routes.ForeignPropertyReportingMethodController.show _
+    }
+    baseRoute(incomeSourceId).url
+  }
+
+  def getIncomeSourceData(incomeSourceType: IncomeSourceType, incomeSourceId: String)(implicit user: MtdItUser[_]): Option[(LocalDate, Option[String])] = {
```
avelx, line 84: this method is a good candidate to be moved somewhere else away from controller:
* service layer / separate trait etc
in the controller we tend to "join" results and render final result to user

app/controllers/incomeSources/add/IncomeSourceAddedController.scala
```diff
@@ -0,0 +1,164 @@
...
+                }
+              }
+            case UkProperty => nextUpdatesService.getObligationsViewModel(incomeSourceId, showPreviousTaxYears) map { viewModel =>
+              Ok(obligationsView(viewModel, backUrl, isAgent = isAgent, incomeSourceType = UkProperty))
+            }
+            case ForeignProperty => nextUpdatesService.getObligationsViewModel(incomeSourceId, showPreviousTaxYears) map { viewModel =>
+              Ok(obligationsView(viewModel, backUrl, isAgent = isAgent, incomeSourceType = ForeignProperty))
+            }
+          }
+        case None => Logger("application").error(
```
avelx, line 135: if you can provide exact details of the journey failed in the exception

app/controllers/incomeSources/add/IncomeSourceAddedController.scala
```diff
@@ -0,0 +1,142 @@
...
+import services.{DateServiceInterface, IncomeSourceDetailsService, NextUpdatesService}
+import uk.gov.hmrc.auth.core.AuthorisedFunctions
+import utils.IncomeSourcesUtils
+import views.html.incomeSources.add.IncomeSourceAddedObligations
+
+import java.time.LocalDate
+import javax.inject.Inject
+import scala.concurrent.{ExecutionContext, Future}
+
+class IncomeSourceAddedController @Inject()(authenticate: AuthenticationPredicate,
```
avelx, line 37: * you might need to mark this controller as 'Shared' as this is shared between few endpoints; thought personally I am against shared controllers and this seem to me as an 'anti-patterns', as we should be able to move shared logic into trait in the worst case scenarios and keep existing controller as they are;

danielmaltz1: All the shared controllers are called 'IncomeSource"XX rather than Business/Foreign/UK, so I think this is clear already.

app/controllers/incomeSources/add/IncomeSourceAddedController.scala
```diff
@@ -0,0 +1,142 @@
...
+    baseRoute(incomeSourceId).url
+  }
+
+  private def handleRequest(isAgent: Boolean, incomeSourceId: String, incomeSourceType: IncomeSourceType)(implicit user: MtdItUser[_], ec: ExecutionContext): Future[Result] = {
+    withIncomeSourcesFS {
+      val backUrl = getBackUrl(incomeSourceId, isAgent, incomeSourceType)
+       incomeSourceDetailsService.getIncomeSourceFromUser(incomeSourceType, incomeSourceId) match {
+        case Some((startDate, businessName)) =>
+          val showPreviousTaxYears: Boolean = startDate.isBefore(dateService.getCurrentTaxYearStart())
+          incomeSourceType match {
```
avelx, line 90: if you can move this part (under incomeSourceType) - under separate private method for better code readability

app/controllers/incomeSources/add/IncomeSourceAddedController.scala
avelx: this type of naming convention would be useful in the future, as I this would help us to identify such controller when we would need to de-compose these back into separate controllers ie.

danielmaltz1: I guess this would be a separate ticket since it would affect many controllers other than this one. Also some controllers have been built as joint so never had separate controllers for different journeys (e.g. the manage journey)

avelx (approved):
approved

PR #1906, opened 2023-09-22
MISUV-6279: SCP025 | Manage SE & Property | TxM event - View and manage income sources page

ayubAz:


app/controllers/incomeSources/manage/ManageIncomeSourceController.scala
```diff
@@ -39,55 +41,53 @@ class ManageIncomeSourceController @Inject()(val manageIncomeSources: ManageInco
...
   def handleRequest(sources: IncomeSourceDetailsModel, isAgent: Boolean, backUrl: String)
                    (implicit user: MtdItUser[_]): Future[Result] = {
-    if (isDisabled(IncomeSources)) {
-      Future.successful(Redirect(controllers.routes.HomeController.show()))
-    } else {
+
+    withIncomeSourcesFS {
       incomeSourceDetailsService.getViewIncomeSourceViewModel(sources) match {
         case Right(viewModel) =>
+          auditingService
```
avelx, line 72: if we can move this code block between lines 81 and 82, this would reduce chances to have false audit record in case of Mongo failure to remove session data, essentially as I understand we audit in exact moment before rendering a view

app/enums/TransactionName/TransactionName.scala
```diff
@@ -116,4 +116,8 @@ case object RefundToTaxPayer extends TransactionName {
 
 case object CreditsSummary extends TransactionName {
   val name = "credits-summary"
+}
+
+case object ManageIncomeSourcesHome extends TransactionName {
+  val name = "view-and-manage-income-sources-home"
```
avelx, line 122: in the story we have auditType set as "ViewAndManageIncomeSourcesHome", 
so we need to update either story or code to be consistent;

avelx (approved):
approved

PR #1910, opened 2023-09-26
MISUV-6282 Manage SE & Property | TxM event - Obligations page

danielmaltz1:


app/enums/AuditType/AuditType.scala
```diff
@@ -121,3 +121,7 @@ case object CreditsSummaryResponse extends AuditType {
 case object MangeIncomeSourcesHome extends AuditType {
   override val name: String = "ViewAndManageIncomeSourcesHome"
 }
+
+case object Obligations extends AuditType {
+  override val name: String = "Obligations"
```
avelx, line 126: its seems quite short name for the auditType, might be something you need to raise with the BAs; but its is correct with respect what is specified in the story

app/audit/models/ObligationsAuditModel.scala
```diff
@@ -0,0 +1,103 @@
...
+      Json.obj(
+        "journeyType" -> journey,
+        "incomeSourceInfo" ->
+          Json.obj(
+            "businessName" -> name,
+            "reportingMethod" -> repMethod,
+            "taxYear" -> s"${taxYear.startYear}-${taxYear.endYear}"
+          ),
+        "quarterlyUpdates" ->
+          quarterly.map { value =>
```
avelx, line 73: maybe something like to resolve values out of Tuple2 ?
```
          quarterly.collect {
            case (taxYear, dataModel) =>
              Json.obj(
                "taxYear" -> s"${taxYear}-${taxYear + 1}",
                "quarter" -> dataModel.map { entry =>
                  Json.obj(
                    "startDate" -> entry.inboundCorrespondenceFrom,
                    "endDate" -> entry.inboundCorrespondenceTo,
                    "deadline" -> entry.inboundCorrespondenceDue
                  )
                }
              )
          },
```

app/enums/AuditType/AuditType.scala
danielmaltz1: I think the idea is to share this event between all of the obligations pages, journey and incomeSourceType get specified in the details

app/audit/models/ObligationsAuditModel.scala
```diff
@@ -0,0 +1,104 @@
...
+
+  override val auditType: String = enums.AuditType.Obligations
+
+  val journey: String = incomeSourceType match {
+    case SelfEmployment => "SE"
+    case UkProperty => "UKPROPERTY"
+    case ForeignProperty => "FOREIGNPROPERTY"
+  }
+
+  val name: String = if (businessName == "Not Found") "Sole trader business" else businessName
```
avelx, line 43: looks like some part of the logic relying on the "Not Found" value rather than on None; we would need to apply changes there for sure

avelx (approved):
approved

PR #1911, opened 2023-09-27
Misuv 6383 update mongo migration

joesheridan:


joesheridan:
test this build

joesheridan:
retest

joesheridan:
test this please

joesheridan:
test this please

avelx:
@joesheridan  - I would assume mongo IT tests not in scope of this work?
as I was expecting mongo repo spec like:
https://github.com/hmrc/trusts-individual-check/blob/ece02a72356f665cef1352d3967f7708bad7201d/it/repositories/IndividualCheckRepositorySpec.scala

joesheridan:
> @joesheridan - I would assume mongo IT tests not in scope of this work? as I was expecting mongo repo spec like: https://github.com/hmrc/trusts-individual-check/blob/ece02a72356f665cef1352d3967f7708bad7201d/it/repositories/IndividualCheckRepositorySpec.scala

yes I was thinking of something like this - will add

app/repositories/UIJourneySessionDataRepository.scala
```diff
@@ -0,0 +1,107 @@
...
+    collection
+      .updateOne(
+        filter = dataFilter(data),
+        update = Updates.set("lastUpdated", Instant.now(clock))
+      )
+      .toFuture()
+      .map(_ => true)
+
+  def get(sessionId: String, journeyType: String): Future[Option[UIJourneySessionData]] = {
+    val data = UIJourneySessionData(sessionId, journeyType, None)
```
EdwardDennis, line 72: I don't think this None is needed

app/services/SessionService.scala
```diff
@@ -17,26 +17,88 @@
...
+    val uiJourneySessionData = UIJourneySessionData(hc.sessionId.get.value, journeyType.toString, None)
+    val jsonAccessorPath = journeyType.operation match {
+      case Add => AddIncomeSourceData.getJSONKeyPath(key)
+      case Manage => ManageIncomeSourceData.getJSONKeyPath(key)
+      case Cease => CeaseIncomeSourceData.getJSONKeyPath(key)
+    }
+    uiJourneySessionDataRepository.updateData(uiJourneySessionData, jsonAccessorPath, value).map(
+      result => result.wasAcknowledged() match {
+        case true => Right(true)
+        case false => Left(new Error("Mongo Save data operation was not acknowledged"))
```
EdwardDennis, line 97: I think this should be an exception rather than an error as it's non-fatal, maybe a RuntimeException?
https://stackoverflow.com/questions/5813614/what-is-difference-between-errors-and-exceptions#:~:text=An%20Error%20%22indicates%20serious%20problems,should%20not%20try%20to%20catch.%22&text=An%20Exception%20%22indicates%20conditions%20that,Exception%20classes%20are%20checked%20exceptions.

joesheridan: I think Error and Exception here are interchangeable, its basically the Throwable aspect that we really need. That stackoverflow thread is highly opinionated..

app/repositories/UIJourneySessionDataRepository.scala
joesheridan: its not really doing any harm, but ok - will remove.

app/services/SessionService.scala
avelx: Screenshot 2023-09-28 at 17 04 40
Error - is a branch of JVM level exception, where is 
Exception is program level exceptions

PR #1920, opened 2023-10-02
MISUV-6356: remove null fields

ayubAz:


avelx:
@ayubAz  - what was the method you use to verify that null will be not in the resulting json ?

app/audit/models/SwitchReportingMethodAuditModel.scala
```diff
@@ -38,10 +39,10 @@ case class SwitchReportingMethodAuditModel(journeyType: String,
   override val detail: JsValue = {
     Utilities.userAuditDetails(user) ++
```
avelx, line 40: please remove usage of  messagesApi: MessagesApi from this model as this is not required and creates more problem with support in the future, we really don't need this type of dynamic logic here

avelx (approved):
approved

PR #1921, opened 2023-10-06
MISUV-6379 Manage SE, property | Mongo session storage migration

danielmaltz1:


app/controllers/incomeSources/manage/ManageIncomeSourceDetailsController.scala
```diff
@@ -153,11 +156,13 @@ class ManageIncomeSourceDetailsController @Inject()(val view: ManageIncomeSource
                 incomeSourceType = SelfEmployment
               )
 
-              result.flatMap(
-                sessionService.set(SessionKeys.incomeSourceId, id, _)
-              ).flatMap {
-                case Right(result) => Future.successful(result)
-                case Left(exception) => Future.failed(exception)
+              sessionService.createSession(JourneyType(Manage, SelfEmployment).toString).flatMap {
+                case true =>
```
avelx, line 160: * createSession would alway return value of true, so checking if result true or false is redundant (maybe we need to return Unit from .createSession ?);
* I am not sure about this, but I think that we need to createSession first before handle actual request, as the outcome of creating session can be fail; but this something we need to agreed across all journeys (if we should apply any order at all ?)

app/controllers/incomeSources/manage/ManageObligationsController.scala
```diff
@@ -169,7 +170,9 @@ class ManageObligationsController @Inject()(val checkSessionTimeout: SessionTime
         case Some(years) =>
           if (changeTo == "annual" || changeTo == "quarterly") {
             getIncomeSourceId(mode, incomeSourceId, isAgent = isAgent) match {
-              case Left(error) => showError(isAgent, {
+              case Left(error) =>
+                Logger("application").info(s"BEEP BEEP 1 ${error.getMessage}, mode: $mode, id: $incomeSourceId")
```
avelx, line 174: this log message is not valuable for investigation, if you can provide controller details pls

app/controllers/incomeSources/manage/ManageObligationsController.scala
```diff
@@ -231,10 +234,10 @@ class ManageObligationsController @Inject()(val checkSessionTimeout: SessionTime
     (incomeSourceType, id) match {
       case (SelfEmployment, Some(id)) => Right(id)
       case _ =>
-        val placeholder = if (incomeSourceType == UkProperty)
-          incomeSourceDetailsService.getActiveUkOrForeignPropertyBusinessFromUserIncomeSources(isUkProperty = true)
+        val placeholder = if (incomeSourceType == UkProperty || incomeSourceType == ForeignProperty)
```
avelx, line 237: Consider chaining code in the next way (usually single variable is flag for this scenarios):

```
        Seq(UkProperty, ForeignProperty).find(x => x == incomeSourceType)
          .map(_ => incomeSourceDetailsService.getActiveUkOrForeignPropertyBusinessFromUserIncomeSources(incomeSourceType == UkProperty))
          .getOrElse(Left(new Error("No id supplied for Self Employment business")))
        match {
          case Right(property: PropertyDetailsModel) => Right(property.incomeSourceId)
          case Left(error: Error) => Left(error)
          case _ => Left(new Error(s"Unknown error. IncomeSourceType: $incomeSourceType"))
        }
```
but again, current approach is fine

app/controllers/incomeSources/manage/ManageIncomeSourceDetailsController.scala
danielmaltz1: To create a session you need to know the operation (Add/Manage/Cease) and also the incomeSourceType(Self Employment/Uk/Foreign), this is the earliest point in the journey where you can know that the user has chosen a Self Employment business

avelx: I was actually referring to next change, which working fine as ATs are passing;
there is not big different from how is works right now, but I really think that hitting Mongo first and get result 
 would be more appropriate before processing request

```
              sessionService.createSession(JourneyType(Manage, SelfEmployment).toString).flatMap {
                case true =>
                  sessionService.setMongoKey(ManageIncomeSourceData.incomeSourceIdField, id, JourneyType(Manage, SelfEmployment)).flatMap {
                    case Right(_) =>
                      handleRequest(
                        sources = mtdItUser.incomeSources,
                        isAgent = true,
                        backUrl = controllers.incomeSources.manage.routes.ManageIncomeSourceController.show(true).url,
                        id = Some(id),
                        incomeSourceType = SelfEmployment
                      )
                    case Left(exception) => Future.failed(exception)
                  }
                case false => Future.failed(new Error("Failed to create mongo session"))
              }
```

avelx (approved):
approved

PR #1923, opened 2023-10-09
Misuv 6445 The back button for a individual user points to agent url from View and Manage Income Sources

tharunprasath:


app/controllers/incomeSources/manage/ManageIncomeSourceController.scala
```diff
@@ -57,8 +57,8 @@ class ManageIncomeSourceController @Inject()(val manageIncomeSources: ManageInco
       sources = user.incomeSources,
       isAgent = isAgent,
       backUrl = {
-        if(isAgent) controllers.routes.HomeController.show()
-        else controllers.routes.HomeController.showAgent
+        if(isAgent) controllers.routes.HomeController.showAgent
```
avelx, line 60: any tests associated with this change ?

tharunprasath: We haven't written a test case for the backlink as part of UT or IT. But this should be covered according to the AT journey coverage. Do we still need to write UT for this fix?

avelx: ATs not failing for this change, so I would think that this change is not covered on any level of tests

avelx (approved):
approved

PR #1924, opened 2023-10-09
MISUV-6378-cease-journey-mongo-migration

simondrugan16:


app/controllers/incomeSources/cease/IncomeSourceEndDateController.scala
```diff
@@ -262,33 +263,43 @@ class IncomeSourceEndDateController @Inject()(val authenticate: AuthenticationPr
...
-        })
+                case _ =>
+                  val propertyEndDate = validatedInput.date.toString
+                  val result = Redirect(redirectAction)
+                  sessionService.setMongoKey(key = CeaseIncomeSourceData.dateCeasedField, value = propertyEndDate, journeyType = JourneyType(Cease, incomeSourceType)).flatMap {
+                    case Right(_) => Future.successful(result)
+                    case Left(exception) => Future.failed(exception)
+                  }
+              }
+            case false => Future.failed(new Error("Failed to create mongo session"))
```
avelx, line 300: this is not reachable branch of the logic; result will be always "true" during evaluation;
you can simply flatMap with mo pattern matching for result value;

app/controllers/incomeSources/cease/IncomeSourceEndDateController.scala
```diff
@@ -262,33 +263,43 @@ class IncomeSourceEndDateController @Inject()(val authenticate: AuthenticationPr
...
+                case (SelfEmployment, Some(incomeSourceId)) =>
+                  val result = Redirect(redirectAction)
+                  sessionService.setMongoKey(
+                    CeaseIncomeSourceData.dateCeasedField, validatedInput.date.toString, JourneyType(Cease, incomeSourceType)
+                  ).flatMap {
+                    case Right(true) => {
+                      sessionService.setMongoKey(
+                        CeaseIncomeSourceData.incomeSourceIdField, incomeSourceId, JourneyType(Cease, incomeSourceType)
+                      ).flatMap {
+                        case Right(true) =>
```
avelx, line 284: this is also not reachable code (line 286), result is always true, so no need to test if this is true or false

app/models/incomeSourceDetails/UIJourneySessionData.scala
```diff
@@ -95,11 +95,17 @@ object ManageIncomeSourceData {
 }
 
 case class CeaseIncomeSourceData(
-                                  selectedIncomeSourceId: String,
-                                  dataCeased: Option[LocalDate]
+                                  incomeSourceId: Option[String],
+                                  endDate: Option[String],
```
avelx, line 99: * what is the reason to change dateCeased from type LocalDate to String ? 
* Type String is not correct representation of the date; if you need any logic to correctly convert LocalDate to any string format, use existing helper method or any a new one;

app/repositories/UIJourneySessionDataRepository.scala
```diff
@@ -105,4 +105,15 @@ class UIJourneySessionDataRepository @Inject()(
       .deleteOne(dataFilter(data))
       .toFuture()
       .map(_ => true)
+
+  private def sessionFilter(sessionID: String): Bson = {
```
avelx, line 109: this method is redundant, not sure if we really need it

app/repositories/UIJourneySessionDataRepository.scala
```diff
@@ -105,4 +105,15 @@ class UIJourneySessionDataRepository @Inject()(
...
+    import Filters._
+    and(equal("sessionId", sessionID))
+  }
+  
+  def deleteOne(sessionId: String): Future[Boolean] =
```
avelx, line 114: * this method is only use in scope of the Spec/tests;
* you should be able to use existing method instead, have a look at example in: => UIJourneySessionDataISpec

```
    "deleteOne should remove a sessionData item" in {
      await(repository.set(UIJourneySessionData("session-12345", "ADD-UKPROP", Some(AddIncomeSourceData(Some("business1"))))))
      val result = await(repository.deleteOne(UIJourneySessionData("session-12345", "ADD-UKPROP")))
      result shouldBe true
      val sessionData = await(repository.get("session-12345", "ADD-UKPROP"))
      sessionData shouldBe None
    }
```

it/controllers/agent/incomeSources/cease/CheckCeaseIncomeSourceDetailsControllerISpec.scala
```diff
@@ -65,6 +86,11 @@ class CheckCeaseIncomeSourceDetailsControllerISpec extends ComponentSpecBase {
     cessation = Some(Cessation(true, Some(LocalDate.parse(cessationDate))))
   )
 
+  override def beforeEach(): Unit = {
+    super.beforeEach()
```
avelx, line 90: you call always call relevant repo directly:

```
 val repository = app.injector.instanceOf[UIJourneySessionDataRepository]
.........
    //await(repository.deleteOne(UIJourneySessionData("testSessionId", .. )))
```

app/models/incomeSourceDetails/UIJourneySessionData.scala
simondrugan16: All the code at the moment uses Strings so I thought replacing everything with LocalDates would be outside the scope of the ticket

app/repositories/UIJourneySessionDataRepository.scala
simondrugan16: This was Solar's implementation for clearing the data between tests. You use this in the beforeAll() and supply the sessionID which clears all data as opposed to the existing method requires a data: UIJourneySessionData which is annoying to use in

app/repositories/UIJourneySessionDataRepository.scala
simondrugan16: This was required for the deleteOne(sessionId: String) method

app/models/incomeSourceDetails/UIJourneySessionData.scala
avelx: I am not sure what you mean by "all the code", there should not be a problem to deal with LocalDate

app/repositories/UIJourneySessionDataRepository.scala
avelx: * it is not correct to add business logic just to satisfy tests scenario;
 * using repo and call required collection method (delete should be fine);

app/controllers/incomeSources/cease/IncomeSourceEndDateController.scala
avelx: you can use Right(_) or any other approach, testing value if its "true" or "false" is not required and it a redundant code we should remove

app/repositories/UIJourneySessionDataRepository.scala
avelx: yes, but deleteOne you added is redundant

app/controllers/incomeSources/cease/IncomeSourceEndDateController.scala
simondrugan16: I don't think I'm testing if it's true or false anywhere? I'm flatmapping for the Left or Right return of sessionService.setMongoKey. I did Right(true) since that's the only return within the Right

avelx (approved):
approved

PR #1929, opened 2023-10-11
MISUV-6449 Incorrect apostrophe formatting in messages

tharunprasath:


app/testOnly/controllers/MessageCheckController.scala
```diff
@@ -0,0 +1,59 @@
...
+import javax.inject.Inject
+
+
+class MessageCheckController @Inject()(messageCheckView: MessageCheckView,
+                                       mcc: MessagesControllerComponents,
+                                       appConfig: FrontendAppConfig,
+                                       languageUtils: LanguageUtils) extends ItvcLanguageController(mcc, appConfig, languageUtils) {
+
+
+  def show(): Action[AnyContent] = Action { implicit req =>
```
avelx, line 35: I don't think we need to read anything from the message files at all;
you should have full access to the next Map[String, Map[String, String]] via next path =>
 mcc.messagesApi.messages

test/testUtils/MessageCheckViewSpec.scala
```diff
@@ -0,0 +1,59 @@
...
+import play.twirl.api.Html
+import testOnly.views.html.MessageCheckView
+
+class MessageCheckViewSpec extends TestSupport {
+
+  val messageCheckView: MessageCheckView = app.injector.instanceOf[MessageCheckView]
+  lazy val msgs: MessagesApi = app.injector.instanceOf[MessagesApi]
+  implicit val lang: Lang = Lang("cy")
+  class Setup(language: String) {
+    val messageKeys = messagesApi.messages.filter(_._1 == language).flatMap(_._2).keys.toList
```
avelx, line 31: Please avoid use of tuples as they reduce code readability dramatically, you can use something like:
    al messageKeys : List[String] = messagesApi.messages.collect {
      case (key, values) if key == language => values.keys
    }.flatten(_.toList).toList

test/testUtils/MessageCheckViewSpec.scala
```diff
@@ -0,0 +1,64 @@
...
+import play.twirl.api.Html
+import testOnly.views.html.MessageCheckView
+
+class MessageCheckViewSpec extends TestSupport {
+
+  val messageCheckView: MessageCheckView = app.injector.instanceOf[MessageCheckView]
+  lazy val msgs: MessagesApi = app.injector.instanceOf[MessagesApi]
+  implicit val lang: Lang = Lang("cy")
+
+  class Setup(language: String) {
```
avelx, line 31: * trait is more preferable way to avoid code duplication;
* under Scala 2.x you can define field in the trait if you like to pass a parameter etc;

test/testUtils/MessageCheckViewSpec.scala
```diff
@@ -0,0 +1,64 @@
...
+          val messageValue = ele.select(".messageValue").text()
+          Some(messageValue) shouldBe messagesMap.get(messageKey)
+        }
+      }
+
+      "return correct Welsh format" in new Setup("cy") {
+        messageKeys.foreach { keys =>
+          Some(msgs(keys)) shouldBe messagesMap.get(keys)
+        }
+        //        document.select(".messageRow").forEach { ele =>
```
avelx, line 56: I would assume we don't need this code??

tharunprasath: Yes. Removed the unused code

it/helpers/ComponentSpecBase.scala
```diff
@@ -269,6 +269,8 @@ trait ComponentSpecBase extends TestSuite with CustomMatchers
 
     def getPaymentHistory: WSResponse = get(s"/payment-refund-history")
 
+    def getMessagesCheck: WSResponse = get(s"/test-only/message-check")
```
avelx, line 272: these are redundant as we are not testing testOnly controller anyway

avelx (approved):
approved with comments

it/helpers/ComponentSpecBase.scala
tharunprasath: Language-specific keys are added in both the message and message.cy files so that some keys might be available in the message and the same won't be there in messacy.cy(blank) and vice versa. I thought of that scenario and designed both routes to test.

PR #1949, opened 2023-10-23
MISUV-6306 Implement twirl components: IS Manage journey

danielmaltz1:


danielmaltz1:
There are some sections that will not be able to be implemented in twirl:

- The banner on the obligations page requires text both above and below the bold heading as the Welsh has the business name in the middle of the sentence.
- The initial Manage page displays the user's income sources using a gov-uk-table, but has actions in the third column, which is not supported by the twirl implementation of tables.

PR #1962, opened 2023-10-27
MISUV-6557-cease-journey-twirlification

simondrugan16:


app/views/incomeSources/cease/CeaseIncomeSources.scala.html
```diff
@@ -64,7 +59,9 @@
     else controllers.incomeSources.cease.routes.DeclarePropertyCeasedController.show(ForeignProperty).url
 }
 
-@noneToUnknown(value:Option[String]) = @{value.getOrElse(getMessage("unknown"))}
+@noneToUnknown(value: Option[String]) = @{
```
avelx, line 62: this would be applicable only to English, what about Welsh? might need to manipulate with Language resources here instead

app/views/incomeSources/cease/CeaseIncomeSources.scala.html
```diff
@@ -64,7 +59,9 @@
...
+    value.getOrElse(getMessage("unknown"))
```
avelx, line 63: just wondering if we need to do any changes to support Welsh?

test/views/incomeSources/cease/CeaseIncomeSourcesViewSpec.scala
```diff
@@ -100,6 +100,10 @@ class CeaseIncomeSourcesViewSpec extends TestSupport with ImplicitDateFormatter
       "all fields have value" in new Setup(false) {
         val table = document.getElementById("sole-trader-businesses-table")
 
+        println("ZZZZZZZ")
```
avelx, line 103: If we can remove this, pls double check PR for warnings, as these are raise by PR-bot

simondrugan16: yes my bad

app/views/incomeSources/cease/CeaseIncomeSources.scala.html
simondrugan16: This gets message "cease-income-sources.unknown" which is in both messages files

avelx (approved):
approved

PR #1963, opened 2023-10-27
MISUV-6544 BusinessReportingMethodController: Refactor

danielmaltz1:


app/controllers/incomeSources/add/BusinessReportingMethodController.scala
```diff
@@ -189,68 +190,44 @@ class BusinessReportingMethodController @Inject()(val authenticate: Authenticati
...
       case ex: Exception =>
-        Logger("application").error(s"[BusinessReportingMethodController][updateReportingMethod]: " +
+        Logger("application").error(s"[ForeignPropertyReportingMethodController][updateReportingMethod]: " +
           s"Error updating tax year specific reporting method: ${ex.getMessage}")
         Redirect(redirectErrorUrl)
     }
   }
 
+  @tailrec
+  private def manageReportingMethodUpdateResponses(results: Seq[UpdateIncomeSourceResponse], redirectUrl: Call): Either[Throwable, Result] = {
```
avelx, line 208: In most of the cases should be sufficient to use foldLeft instead of custom recursion as this is hard to support in the future;

avelx: as I understand underlying logic, this should be fine to find first error if any (or find all errors); 
otherwise collect success result in the log ( foldLeft would work fine for this) and redirect with success

avelx: Here is potential solution for this processing:
```

      val errors = results.collect {
        case e: UpdateIncomeSourceResponseError => e
      }
      val success = results.collect {
        case c: UpdateIncomeSourceResponseModel => c
      }
      (errors, success) match {
        case (es: Seq[UpdateIncomeSourceResponseError], ss: UpdateIncomeSourceResponseModel) if es.length == 0 && ss.length == 0=>
          Logger("application").info(s"[BusinessReportingMethodController][updateReportingMethod]: " +
            s"Updated tax year specific reporting method: ${es.mkString(" - ")}")
          Redirect(redirectUrl)
        case (es: Seq[UpdateIncomeSourceResponseError] , _) if es.length > 0 =>
          Logger("application").info(s"[BusinessReportingMethodController][updateReportingMethod]: " +
            s"Updated tax year specific reporting method: ${es.mkString(" - ")}")
          Redirect(redirectUrl)
           // TODO: add other cases here ...
      }
```

app/controllers/incomeSources/add/BusinessReportingMethodController.scala
```diff
@@ -183,71 +184,37 @@ class BusinessReportingMethodController @Inject()(val authenticate: Authenticati
...
+    } yield {
+      val errors = results.collect {
+        case error: UpdateIncomeSourceResponseError => error
+      }
+      val success = results.collect {
+        case success: UpdateIncomeSourceResponseModel => success
+      }
+      (errors, success) match {
+        case (es: Seq[UpdateIncomeSourceResponseError], _) if es.isEmpty =>
+          Logger("application").info(s"[BusinessReportingMethodController][updateReportingMethod]: " +
```
avelx, line 200: I would assume this should be log level error?

avelx (approved):
approved

PR #1971, opened 2023-11-06
MISUV-6506: Add Journey check details unification

danielmaltz1:


app/controllers/incomeSources/add/IncomeSourceCheckDetailsController.scala
```diff
@@ -0,0 +1,229 @@
...
+    {
+      incomeSourceType match {
+        case SelfEmployment => getBusinessModel
+        case _ => getPropertyModel(incomeSourceType)
+      }
+    }.flatMap {
+      case Right(viewModel) =>
+        businessDetailsService.createRequest(viewModel).flatMap {
+          case Right(CreateIncomeSourceResponse(id)) =>
+            sessionService.deleteMongoData(JourneyType(Add, incomeSourceType))
```
avelx, line 213: * this would not work as expected, as you "cutting off" Future to be completed: line => deleteMongoData

app/models/incomeSourceDetails/viewmodels/CheckDetailsViewModel.scala
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+package models.incomeSourceDetails.viewmodels
+
+import enums.IncomeSourceJourney.IncomeSourceType
+
+import java.time.LocalDate
+
+
+case class CheckDetailsViewModel(businessName: Option[String] = None,
```
avelx, line 24: based on this model definition its just 3 field would be mandatory, can you confirm that this is correct?

app/models/incomeSourceDetails/viewmodels/CheckDetailsViewModel.scala
```diff
@@ -0,0 +1,51 @@
...
+                                 businessAddressLine4: Option[String] = None,
+                                 businessPostalCode: Option[String] = None,
+                                 businessCountryCode: Option[String] = None,
+                                 incomeSourcesAccountingMethod: Option[String] = None,
+                                 cashOrAccrualsFlag: String,
+                                 showedAccountingMethod: Boolean = true,
+                                 incomeSourceType: IncomeSourceType
+ ){
+
+  def countryName: Option[String] = Some("United Kingdom")
```
avelx, line 40: I would assume we should point to relevant language resource here, in the same fashion as below

test/views/incomeSources/add/ForeignPropertyCheckDetailsViewSpec.scala
```diff
@@ -14,74 +14,90 @@
...
-      val accChangeLink: Element = layoutContent.getElementById("accounting-method-change-details-link")
-      accChangeLink.hasCorrectHref(changeAccMethodIndiv)
-      accChangeLink.text() shouldBe ForeignCheckMessages.change
-
-      val button: Element = layoutContent.getElementById("confirm-button")
-      button.text() shouldBe ForeignCheckMessages.button
-    }
-  }
-}
+///*
```
avelx, line 17: Do we need this code?

test/views/incomeSources/add/CheckUKPropertyDetailsViewSpec.scala
```diff
@@ -14,114 +14,130 @@
...
-      document.select("dl:nth-of-type(2) > div > dt.govuk-summary-list__key").text() shouldBe accountingMethodLabel
-      document.select("dl:nth-of-type(2) > div > dd.govuk-summary-list__value").text() shouldBe cash
-      document.getElementById("change-accounting-method-link").attr("href") shouldBe changeAccountingMethodUrl
-      document.getElementById("change-accounting-method-link").text() shouldBe s"$change $change"
-
-      document.getElementById("check-uk-property-details-form").getElementsByTag("button").get(0).text() shouldBe continue
-    }
-  }
-}
+///*
```
avelx, line 17: Do we need this code[2]?

test/views/incomeSources/add/CheckBusinessDetailsViewSpec.scala
```diff
@@ -14,135 +14,152 @@
  * limitations under the License.
  */
 
-package views.incomeSources.add
+///*
```
avelx, line 17: Do we need this code[3]?

app/views/incomeSources/add/IncomeSourceCheckDetails.scala.html
```diff
@@ -0,0 +1,208 @@
...
+                        <a class="govuk-link" href=@changeBusinessNameURL>
+                        @getMessage("change")
+                        </a>
+                    </dd>
+                </div>
+            }
+            <div class="govuk-summary-list__row">
+                <dt class="govuk-summary-list__key">@getMessage("start-date")</dt>
+                <dd class="govuk-summary-list__value" id="start-date-value">@{
+                    viewModel.businessStartDate.get.toLongDate
```
avelx, line 145: if we can avoid calling .get and .head in the view? do we have a default value or we should raise an exception when constructing model ? it seem that we this field should not be Optional;

if this is different across view (entities: in some Optional and in some mandatory ) we would have to use different model definition for this

app/views/incomeSources/add/IncomeSourceCheckDetails.scala.html
```diff
@@ -0,0 +1,208 @@
...
+    ))
+
+    @formWithCSRF(action = postAction) {
+
+        <dl class="govuk-summary-list" id="check-details-table">
+            @if(viewModel.incomeSourceType == SelfEmployment){
+                <div class="govuk-summary-list__row">
+                    <dt class="govuk-summary-list__key">@getMessage("business-name")</dt>
+                    <dd class="govuk-summary-list__value" id="business-name-value">@{
+                        viewModel.businessName.getOrElse("")
```
avelx, line 133: if there is default value so this would be easy to do this when you constructing model and change field type to be not Optional

app/views/incomeSources/add/IncomeSourceCheckDetails.scala.html
```diff
@@ -0,0 +1,208 @@
...
+                        <a class="govuk-link" href=@changeBusinessTradeURL>
+                        @getMessage("change")
+                        </a>
+                    </dd>
+                </div>
+                <div class="govuk-summary-list__row">
+                    <dt class="govuk-summary-list__key">@getMessage("business-address")</dt>
+                    <dd class="govuk-summary-list__value" id="business-address-value">@{
+                        HtmlFormat.raw(getLongAddressFromBusinessAddressDetails(
+                            address1 = viewModel.businessAddressLine1.getOrElse(""),
```
avelx, line 169: as far as I remember businessAddressLine1 would be always provided so we can change this to be not Optional

app/services/CreateBusinessDetailsService.scala
```diff
@@ -50,41 +51,52 @@ class CreateBusinessDetailsService @Inject()(val createIncomeSourceConnector: Cr
...
         Future.successful(Left(new Error(s"Failed to create incomeSources: $ex")))
     }
   }
 
   private def removeEmptyStrings(strOpt: Option[String]): Option[String] = {
-      strOpt match {
-        case Some(str) => if (str == "") None else Some(str)
-        case None => None
-      }
+    strOpt match {
```
avelx, line 66: If this can be moved into extension method like (we can create story for this, as this is generic function we need across the project):

```
  1 implicit class OptionExtension(in: Option[String]) {
  2   def trim2(): Option[String] = {
  3      in match {
  4        case Some("") => None
  5        case _ => in
  6      }
  7   }
  8 }
  9
 10 val s : Option[String] = Some("")
 11
 12 println(s.trim2 )
```

app/models/incomeSourceDetails/viewmodels/CheckDetailsViewModel.scala
danielmaltz1: We don't have a message for this, or the Welsh. We would need to ask Olga to request the Welsh. Maybe best to leave to another ticket?

avelx: as long as this is raised/communicated, should be fine

app/services/CreateBusinessDetailsService.scala
avelx: Next story created to cover this change: 
https://jira.tools.tax.service.gov.uk/browse/MISUV-6629

PR #1982, opened 2023-11-14
MISUV-6581 - TxM event -Create Income Source: success/failure (API#1566)

JordHemsworth:


app/audit/Utilities.scala
```diff
@@ -35,6 +35,15 @@ object Utilities {
     ("credId", user.credId) ++
     userType(user.userType)
 
+  def userAuditDetailsNino(user: MtdItUserBase[_]): JsObject = Json.obj(
+    "nino" -> user.nino,
+    "mtditid" -> user.mtditid
+  ) ++
+    ("agentReferenceNumber", user.arn) ++
```
joesheridan, line 42: dupe code here from above - can you extract a common private method eg getBaseDetails (without nino fields) and then apply either nino or nationalinsurancenumber depending on the called method (ie userAuditDetails or userAuditdetailsNino)?

app/audit/models/CreateIncomeSourceAuditModel.scala
```diff
@@ -0,0 +1,92 @@
...
+
+import audit.Utilities.{userAuditDetails, userAuditDetailsNino}
+import auth.MtdItUser
+import enums.IncomeSourceJourney.{IncomeSourceType, SelfEmployment}
+import models.createIncomeSource.CreateIncomeSourceResponse
+import models.incomeSourceDetails.viewmodels.{CheckBusinessDetailsViewModel, CheckDetailsViewModel, CheckPropertyViewModel}
+import play.api.libs.json.{JsObject, JsValue, Json}
+import utils.Utilities._
+
+
```
joesheridan, line 27: could fix: two spacing lines is unnecessary - 1 is always enough 😄

app/audit/models/CreateIncomeSourceAuditModel.scala
```diff
@@ -0,0 +1,92 @@
...
+        seDetails
+
+      case propertyViewModel: CheckPropertyViewModel =>
+        val propDetails = Json.obj("dateStarted" -> propertyViewModel.tradingStartDate)
+
+        propDetails
+    }
+
+    val accountingMethod = if (viewModel.cashOrAccruals.toLowerCase.equals("cash")) {
+      "Cash basis accounting"
```
joesheridan, line 82: instead of "cash basis accounting" - can we preserve the API vals eg CASH or ACCRUALS?

JordHemsworth: We can however the example json from ticket does have -   "accountingMethod":"cash basis accounting"   
Are we better keeping the vals?

app/audit/Utilities.scala
JordHemsworth: resolved in recent commit

app/audit/models/CreateIncomeSourceAuditModel.scala
JordHemsworth: good eye

app/audit/models/CreateIncomeSourceAuditModel.scala
JordHemsworth: resolved in recent commit

PR #1988, opened 2023-11-17
MISUV-6439 add SE & property | Build: Back browser button from Final Obligations Page

danielmaltz1:


wjstanistreet:
Nothing to preview

app/utils/IncomeSourcesUtils.scala
```diff
@@ -18,13 +18,19 @@ package utils
...
 import play.api.mvc.Results.Redirect
+import services.SessionService
 import uk.gov.hmrc.auth.core.AffinityGroup.Agent
+import uk.gov.hmrc.http.HeaderCarrier
 
-import scala.concurrent.Future
+import scala.concurrent.{ExecutionContext, Future}
 
 trait IncomeSourcesUtils extends FeatureSwitching {
+
```
avelx, line 33: * IncomeSourceUtils trait will be remove when this part of domain will be finally activated / enabled in prod, no need to add any extra logic within it, you can create a new trait ie. like trait JourneyChecker or JourneyManager and extend it with IncomeSourcesUtils trait

app/utils/IncomeSourcesUtils.scala
```diff
@@ -35,4 +41,27 @@ trait IncomeSourcesUtils extends FeatureSwitching {
       codeBlock
     }
   }
+  def withIncomeSourcesFSWithSessionCheck(sessionService: SessionService, journeyType: JourneyType)(codeBlock: => Future[Result])(implicit user: MtdItUser[_], hc: HeaderCarrier, ec: ExecutionContext): Future[Result] = {
```
avelx, line 44: remove code duplicate ,ie. in this way:

 ```
 def withIncomeSourcesFSWithSessionCheck(sessionService: SessionService, journeyType: JourneyType)(codeBlock: => Future[Result])(implicit user: MtdItUser[_], hc: HeaderCarrier, ec: ExecutionContext): Future[Result] = {
    withIncomeSourcesFS {
      journeyChecker(sessionService, journeyType).flatMap {
        case true => user.userType match {
          case Some(Agent) => Future.successful(Redirect(controllers.incomeSources.add.routes.YouCannotGoBackErrorController.showAgent(journeyType.businessType)))
          case _ => Future.successful(Redirect(controllers.incomeSources.add.routes.YouCannotGoBackErrorController.show(journeyType.businessType)))
        }
        case false => codeBlock
      }
    }
  }
```

File: app/utils/IncomeSourcesUtils.scala [hunk 3]
```diff
@@ -35,4 +41,27 @@ trait IncomeSourcesUtils extends FeatureSwitching {
...
+        case true => user.userType match {
+          case Some(Agent) => Future.successful(Redirect(controllers.incomeSources.add.routes.YouCannotGoBackErrorController.showAgent(journeyType.businessType)))
+          case _ => Future.successful(Redirect(controllers.incomeSources.add.routes.YouCannotGoBackErrorController.show(journeyType.businessType)))
+        }
+        case false => codeBlock
+      }
+    }
+  }
+
+  def journeyChecker(sessionService: SessionService, journeyType: JourneyType)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Boolean] = {
```
avelx, line 61: you suppressing error details in here, this is fine in case you really need this;
but I would consider to add error logging in so we can trace this even in Kibana

danielmaltz1: For some reason the default false value in Mongo can't be retrieved so it usually gives an error even though things are acting as they should (the value has not been set to true yet). I can add the error logging in but I think it will be mostly false errors

app/utils/IncomeSourcesUtils.scala
```diff
@@ -18,13 +18,19 @@ package utils
...
 import play.api.mvc.Result
 import play.api.mvc.Results.Redirect
+import services.SessionService
 import uk.gov.hmrc.auth.core.AffinityGroup.Agent
+import uk.gov.hmrc.http.HeaderCarrier
 
-import scala.concurrent.Future
+import scala.concurrent.{ExecutionContext, Future}
 
 trait IncomeSourcesUtils extends FeatureSwitching {
```
EdwardDennis, line 32: You may be able to avoid passing in sessionService and ec by doing this:

trait IncomeSourcesUtils extends FeatureSwitching {
  self =>

  implicit val ec: ExecutionContext

  val sessionService: SessionService

EdwardDennis: Also there is a lot of unformatted code in this PR

app/controllers/incomeSources/add/IncomeSourceReportingMethodController.scala
```diff
@@ -104,32 +106,44 @@ class IncomeSourceReportingMethodController @Inject()(val authenticate: Authenti
...
+                Redirect(redirectUrl(isAgent, incomeSourceType, id.value))
+              }
+          }
+      }
+      }.recover {
+        case ex: Exception =>
+          val errorHandler = if (isAgent) itvcErrorHandlerAgent else itvcErrorHandler
+          Logger("application").error(
+            s"[UKPropertyReportingMethodController][handleRequest]:" +
+              s"Unable to display IncomeSourceReportingMethod page for $incomeSourceType: ${ex.getMessage}")
```
avelx, line 143: + if you can add ex.getCause in error log entry as this helpful at the time of investigation

app/utils/JourneyChecker.scala
```diff
@@ -0,0 +1,50 @@
...
+        case true => user.userType match {
+          case Some(Agent) => Future.successful(Redirect(controllers.incomeSources.add.routes.YouCannotGoBackErrorController.showAgent(journeyType.businessType)))
+          case _ => Future.successful(Redirect(controllers.incomeSources.add.routes.YouCannotGoBackErrorController.show(journeyType.businessType)))
+        }
+        case false => codeBlock
+      }
+    }
+  }
+
+  def journeyChecker(sessionService: SessionService, journeyType: JourneyType)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Boolean] = {
```
avelx, line 43: if we can mark this method as private as its not used anywhere else except this trait

app/controllers/incomeSources/add/YouCannotGoBackErrorController.scala
```diff
@@ -0,0 +1,74 @@
...
+                                               val itvcErrorHandler: ItvcErrorHandler,
+                                               val itvcErrorHandlerAgent: AgentItvcErrorHandler) extends ClientConfirmedController with IncomeSourcesUtils {
+
+
+  def handleRequest(isAgent: Boolean, incomeSourceType: IncomeSourceType)
+                   (implicit user: MtdItUser[_]): Future[Result] = withIncomeSourcesFS {
+    Future.successful(Ok(cannotGoBackError(isAgent, incomeSourceType)))
+  }
+
+  def show(incomeSourceType: IncomeSourceType): Action[AnyContent] = (checkSessionTimeout andThen authenticate
```
avelx, line 53: if we can add withIncomeSourcesFS check on these controllers as well

app/controllers/incomeSources/add/YouCannotGoBackErrorController.scala
```diff
@@ -0,0 +1,74 @@
...
+import play.api.mvc._
+import services.{CreateBusinessDetailsService, IncomeSourceDetailsService, SessionService}
+import uk.gov.hmrc.auth.core.AuthorisedFunctions
+import utils.IncomeSourcesUtils
+import views.html.incomeSources.add.{IncomeSourceNotAddedError, YouCannotGoBackError}
+
+import javax.inject.Inject
+import scala.concurrent.{ExecutionContext, Future}
+
+class YouCannotGoBackErrorController @Inject()(val checkSessionTimeout: SessionTimeoutPredicate,
```
avelx, line 34: Maybe something more neutral in terms of naming convention: NoBackErrorController ie.?

avelx (approved):

app/controllers/incomeSources/add/YouCannotGoBackErrorController.scala
danielmaltz1: In this controller there's withIncomeSourcesFS on the handleRequest. If I do the version that checks for the session service value, then this page will infinitely redirect to itself. I'll go and add it to the other error pages though if you think that's worthwhile

PR #1990, opened 2023-11-20
MISUV-6662-Implement-IncomeSourceIdHash-type

simondrugan16:


app/models/incomeSourceDetails/incomeSourceIds/IncomeSourceId.scala
```diff
@@ -0,0 +1,46 @@
...
+    if (validateStringAsIncomeSourceId(incomeSourceAsString = incomeSourceAsString)) {
+      new IncomeSourceId(incomeSourceAsString)
+    } else {
+      Logger("application").info(s"[IncomeSourceId][mkIncomeSourceId] incomeSourceId was not the correct length or contained special characters")
+      new IncomeSourceId(incomeSourceAsString)
+    }
+
+  }
+
+  def validateStringAsIncomeSourceId(incomeSourceAsString: String): Boolean = {
```
avelx, line 41: We are already in the IncomeSourceId space, so you can call method simply: 
=> validate

app/models/incomeSourceDetails/incomeSourceIds/IncomeSourceIdHash.scala
```diff
@@ -0,0 +1,45 @@
...
+}
+
+object IncomeSourceIdHash {
+  def mkFromIncomeSourceId(id: IncomeSourceId): IncomeSourceIdHash = {
+    val hash = id.value.hashCode().abs.toString
+    new IncomeSourceIdHash(hash)
+  }
+
+  def mkFromQueryString(id: String): Either[Throwable, IncomeSourceIdHash] = Try {
+    val hash = id.hashCode().abs.toString
```
avelx, line 41: you can rename id to hashCodeAsString and there is not need to generate hash again

test/models/incomeSourceDetails/incomeSourceIds/IncomeSourceIdSpec.scala
```diff
@@ -0,0 +1,38 @@
...
+import testUtils.UnitSpec
+
+class IncomeSourceIdSpec extends UnitSpec {
+
+  "IncomeSourceId object" should {
+    "return the overridden toString of the incomeSourceId" when {
+      "on an incomeSourceId object" in {
+        val incomeSourceId: IncomeSourceId = mkIncomeSourceId(testSelfEmploymentId)
+        val incomeSourceIdValue = incomeSourceId.value
+        val incomeSourceIdToString: String = s"IncomeSourceId: $incomeSourceIdValue"
```
avelx, line 31: .toString is a helper method, no need to test , you can actually compare with value created out of mkFromQueryString

app/models/incomeSourceDetails/incomeSourceIds/IncomeSourceIdHash.scala
```diff
@@ -0,0 +1,45 @@
...
+  def oneOf(ids: List[IncomeSourceId]): Option[IncomeSourceId] = {
+    ids.find(id => {
+      val hashId = mkFromIncomeSourceId(id)
+      hashId.hash == this.hash
+    })
+  }
+}
+
+object IncomeSourceIdHash {
+  def mkFromIncomeSourceId(id: IncomeSourceId): IncomeSourceIdHash = {
```
avelx, line 35: if you can keep original naming, here type not relevant in the naming conversion

simondrugan16: I can change it back. I thought this was better as we have 2 different methods to make the IncomeSourceHash with different inputs 🤷

test/models/incomeSourceDetails/incomeSourceIds/IncomeSourceIdHashSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+    "return IncomeSourceIdHash objects" when {
+
+      "calling both the mkFromIncomeSourceId and mkFromQueryString methods" in {
+        val incomeSourceId: IncomeSourceId = mkIncomeSourceId(testSelfEmploymentId)
+        val incomeSourceIdHashFromIncomeSourceId: IncomeSourceIdHash = IncomeSourceIdHash.mkFromIncomeSourceId(incomeSourceId)
+
+        val incomeSourceIdHashFromQueryString: Either[Throwable, IncomeSourceIdHash] = IncomeSourceIdHash.mkFromQueryString(testSelfEmploymentId)
+
+        incomeSourceIdHashFromQueryString match {
+          case Left(error) => Failed("IncomeSourceIdHash.mkFromQueryString returned an Either[Left]")
```
avelx, line 38: feel free to covert to Option and use .get, as it should be fine for the scope of unit test, we don't have to handle error etc

app/models/incomeSourceDetails/incomeSourceIds/IncomeSourceIdHash.scala
simondrugan16: Are you sure? This is for creating a IncomeSourceIdHash from a string of the id. The input is the actual id, not the hash of the id, right?

test/models/incomeSourceDetails/incomeSourceIds/IncomeSourceIdHashSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+
+  "IncomeSourceIdHash class" should {
+
+    "return IncomeSourceIdHash objects" when {
+
+      "calling both the mkFromIncomeSourceId and mkFromQueryString methods" in {
+        val incomeSourceId: IncomeSourceId = mkIncomeSourceId(testSelfEmploymentId)
+        val incomeSourceIdHashFromIncomeSourceId: IncomeSourceIdHash = IncomeSourceIdHash.mkFromIncomeSourceId(incomeSourceId)
+
+        val incomeSourceIdHashFromQueryString: Either[Throwable, IncomeSourceIdHash] = IncomeSourceIdHash.mkFromQueryString(testSelfEmploymentId)
```
avelx, line 35: we creating IncomeSourceIdHash value out of it string representation, can be something like: "afasv", "asffq" so actual hash code etc

test/models/incomeSourceDetails/incomeSourceIds/IncomeSourceIdSpec.scala
simondrugan16: Since this is the only test in the IncomeSourceIdSpec should I just remove the spec altogether, or is there something else I can test

app/models/incomeSourceDetails/incomeSourceIds/IncomeSourceIdHash.scala
avelx: you can look at this as a suggestion, I don't mind new naming convention as long as this would be consistent in other part of code base

app/models/incomeSourceDetails/incomeSourceIds/IncomeSourceIdHash.scala
avelx: yes, I am sure, if you can rename id to be a hashCodeAsString or something, as this is a value we would read from url/query params;

simondrugan16: Ok but calling the input _hashCodeAsString_ implies that it's already a hashCode, even though you hash it later on... If anything, it should be:
    val hashCodeAsString = id.hashCode().abs.toString

test/models/incomeSourceDetails/incomeSourceIds/IncomeSourceIdHashSpec.scala
simondrugan16: in the actual method?

avelx (approved):
approved

PR #1996, opened 2023-11-22
MISUV-6613: WIP / Add journey / convert incomeSourceId from String to IncomeSourceId type

avelx:
* changes in main / Add journey

### DIFF
--- a/app/controllers/incomeSources/add/IncomeSourceCheckDetailsController.scala
+++ b/app/controllers/incomeSources/add/IncomeSourceCheckDetailsController.scala
@@ -81,7 +81,9 @@ class IncomeSourceCheckDetailsController @Inject()(val checkDetailsView: IncomeS
         }
   }
 
-  private def handleRequest(sources: IncomeSourceDetailsModel, isAgent: Boolean, incomeSourceType: IncomeSourceType)
+  private def handleRequest(sources: IncomeSourceDetailsModel,
+                            isAgent: Boolean,
+                            incomeSourceType: IncomeSourceType)
                            (implicit user: MtdItUser[_]): Future[Result] = withIncomeSourcesFS {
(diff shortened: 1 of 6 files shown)


app/controllers/incomeSources/add/IncomeSourceCheckDetailsController.scala
```diff
@@ -120,25 +123,27 @@ class IncomeSourceCheckDetailsController @Inject()(val checkDetailsView: IncomeS
...
+                  Right(CheckPropertyViewModel(
+                    tradingStartDate = date,
+                    cashOrAccrualsFlag = method,
+                    incomeSourceType = incomeSourceType
+                  ))
+                case (_, _) =>
+                  Left(new Error(s"Start date or accounting method not found in session. Start date: $dateMaybe, AccMethod: $methodMaybe"))
+              }
+            case (_, _) => Left(new Error(s"Error while retrieving date started or accounting method from session"))
+          }
```
simondrugan16, line 144: I thought about adding another constructor for Option[IncomeSourceId] like so:

`object IncomeSourceId {
  def mkIncomeSourceId(incomeSourceAsString: String): IncomeSourceId = {
    new IncomeSourceId(incomeSourceAsString)
  }

  def mkIncomeSourceIdFromOption(incomeSourceAsStringMaybe: Option[String]): Either[Throwable, Option[IncomeSourceId]] = {
    incomeSourceAsStringMaybe match {
      case Some(incomeSourceIdAsString: String) => Right(Option(new IncomeSourceId(incomeSourceIdAsString)))
      case None => Left(new Error(s"Error while creating incomeSourceId object. incomeSourceAsStringMaybe was empty"))
    }
  }
}`

What do you think?

avelx: this method is redundant, you call always do something like:

```
val idMaybe: Option[String] = Some("String")
val incomeSourceIdMaybe = idMaybe.map(mkIncomeSourceId)
```

PR #1998, opened 2023-11-23
Misuv 6526 enhance hash unit test

avelx:


test/models/incomeSourceDetails/IncomeSourceIdSpecification.scala
```diff
@@ -0,0 +1,42 @@
...
+
+package models.incomeSourceDetails
+
+import models.core.IncomeSourceIdHash
+import org.scalacheck.{Gen, Properties}
+import IncomeSourceIdHash._
+import models.core.IncomeSourceId._
+
+object IncomeSourceIdSpecification extends Properties("IncomeSourceId") {
+  import org.scalacheck.Prop.forAll
```
simondrugan16, line 25: Move import to top perhaps?

test/models/incomeSourceDetails/IncomeSourceIdSpecification.scala
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package models.incomeSourceDetails
+
+import models.core.IncomeSourceIdHash
+import org.scalacheck.{Gen, Properties}
+import IncomeSourceIdHash._
+import models.core.IncomeSourceId._
+
+object IncomeSourceIdSpecification extends Properties("IncomeSourceId") {
```
simondrugan16, line 24: Is there a reason for this to be an object/named IncomeSourceIdSpecification rather than IncomeSourceIdSpec/not use matchers?

avelx: Spec name changed to IncomeSourceIdHashSpecification
IncomeSourceIdSpec is existing spec with unit tests

test/models/incomeSourceDetails/IncomeSourceIdSpecification.scala
avelx: not sure if this is make big difference, as its been added by IDE, move in imports just in case

simondrugan16 (approved):

PR #2002, opened 2023-11-24
MISUV-6414-use-hashed-values-in-links

simondrugan16:


app/models/core/IncomeSourceId.scala
```diff
@@ -31,6 +32,17 @@ object IncomeSourceId {
 
   }
 
+  def compare(id: Option[IncomeSourceIdHash])(implicit user: MtdItUser[_]): Option[IncomeSourceId] = {
```
avelx, line 35: there is already method you can use -> oneOf
no need to add a new one

avelx, line 35: we don't need to add this complexity to the type level
we should keep time as simple as possible

app/utils/IncomeSourceIdUtils.scala
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package utils
+
+import auth.MtdItUser
+import models.core.IncomeSourceId.mkIncomeSourceId
+import models.core.IncomeSourceIdHash.mkFromQueryString
+import models.core.{IncomeSourceId, IncomeSourceIdHash}
+
+trait IncomeSourceIdUtils {
```
avelx, line 24: no need for these, pls use one provided by the actual type, as this is duplication

app/models/core/IncomeSourceIdHash.scala
```diff
@@ -38,8 +38,8 @@ object IncomeSourceIdHash {
     new IncomeSourceIdHash(s"${hashB}${hashA}")
   }
 
-  def mkFromQueryString(hashCodeAsString: String): Either[Throwable, IncomeSourceIdHash] = Try {
+  def mkFromQueryString(hashCodeAsString: String): Option[IncomeSourceIdHash] = Try {
```
avelx, line 41: I think we can keep original method signature as it was, and convert Either to Option where we need this, as we rely suppressing error details here

simondrugan16: If we convert to option we suppress the error details anyway

app/controllers/incomeSources/manage/ManageIncomeSourceDetailsController.scala
```diff
@@ -306,13 +310,15 @@ class ManageIncomeSourceDetailsController @Inject()(val view: ManageIncomeSource
...
+  def handleRequest(sources: IncomeSourceDetailsModel, isAgent: Boolean, backUrl: String, maybeIncomeSourceIdHash: Option[IncomeSourceIdHash],
+                    incomeSourceType: IncomeSourceType)(implicit user: MtdItUser[_], hc: HeaderCarrier): Future[Result] = {
+
+    val incomeSourceIdMaybe: Option[IncomeSourceId] = compare(incomeSourceIdHash = maybeIncomeSourceIdHash)
 
     withIncomeSourcesFS {
       for {
         value <- if (incomeSourceType == SelfEmployment) {
-          getManageIncomeSourceViewModel(sources = sources, incomeSourceId = id.get, isAgent = isAgent)
+          getManageIncomeSourceViewModel(sources = sources, incomeSourceId = incomeSourceIdMaybe.getOrElse(mkIncomeSourceId("")), isAgent = isAgent)
```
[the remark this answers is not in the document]
simondrugan16: Yes. I thought adding validation was outside the scope of this ticket since we don't really know the constraints on income source id. With current implementation, getManageIncomeSourceViewModel returns an error when passed an empty string

app/models/core/IncomeSourceIdHash.scala
avelx: yes, but this would decided by the caller and can still go back to preserver actual error in the future

app/models/incomeSourceDetails/IncomeSourceDetailsResponse.scala
```diff
@@ -99,6 +96,12 @@ case class IncomeSourceDetailsModel(nino: String,
       case _ => None
     }
   }
+
+  def compareHashToQueryString(incomeSourceIdHash: Option[IncomeSourceIdHash])(implicit user: MtdItUser[_]): Option[IncomeSourceId] = {
```
avelx, line 100: I would suspect that this should be:

`def compareHashToQueryString(incomeSourceIdHash: IncomeSourceIdHash)(implicit user: MtdItUser[_]): Option[IncomeSourceId]`

avelx (approved):
approved

PR #2010, opened 2023-11-30
MISUV-6439:  refactoring proposal

avelx:
* proposal to refactor existing as long as apply this approach to other pages we enforce noGoBack logic;

### DIFF
--- a/app/controllers/incomeSources/add/AddBusinessNameController.scala
+++ b/app/controllers/incomeSources/add/AddBusinessNameController.scala
@@ -117,7 +117,7 @@ class AddBusinessNameController @Inject()(authenticate: AuthenticationPredicate,
     }
 
   def handleRequest(isAgent: Boolean, backUrl: String, isChange: Boolean)(implicit user: MtdItUser[_], ec: ExecutionContext): Future[Result] = {
-    withIncomeSourcesFSWithSessionCheck(JourneyType(Add, SelfEmployment)) {
+    withCustomSession(JourneyType(Add, SelfEmployment)) {
       getBusinessName(isChange).flatMap {
         nameOpt =>
           val filledForm = nameOpt.fold(BusinessNameForm.form)(name =>
--- a/app/controllers/incomeSources/add/AddBusinessNameController.scala
+++ b/app/controllers/incomeSources/add/AddBusinessNameController.scala
@@ -73,13 +73,7 @@ class AddBusinessNameController @Inject()(authenticate: AuthenticationPredicate,
   private def getBusinessName(isChange: Boolean, sessionData: Option[UIJourneySessionData])
                              (implicit user: MtdItUser[_]): Future[Option[String]] = {
     if (isChange)
-      sessionData match {
-        case Some(session) => session.addIncomeSourceData match {
-          case Some(data) => Future.successful(data.businessName)
-          case None => Future.failed(new Error("No session data found"))
-        }
-        case None => Future.failed(new Error("No session data found"))
-      }
+      Future.successful(sessionData.get.addIncomeSourceData.get.businessName)
     else
       sessionService.createSession(journeyType.toString).flatMap {
         case true => Future.successful(None)
--- a/app/controllers/incomeSources/add/AddBusinessNameController.scala
+++ b/app/controllers/incomeSources/add/AddBusinessNameController.scala
@@ -24,7 +24,7 @@ import controllers.predicates._
 import enums.IncomeSourceJourney.SelfEmployment
 import enums.JourneyType.{Add, JourneyType}
 import forms.incomeSources.add.BusinessNameForm
-import models.incomeSourceDetails.AddIncomeSourceData
+import models.incomeSourceDetails.{AddIncomeSourceData, UIJourneySessionData}
 import play.api.Logger
 import play.api.i18n.I18nSupport
 import play.api.mvc._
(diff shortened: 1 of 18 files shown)


app/utils/JourneyChecker.scala
```diff
@@ -32,23 +33,45 @@ trait JourneyChecker extends IncomeSourcesUtils {
...
-  def withIncomeSourcesFSWithSessionCheck(journeyType: JourneyType)(codeBlock: => Future[Result])(implicit user: MtdItUser[_], hc: HeaderCarrier): Future[Result] = {
+
+  def withCustomSession(journeyType: JourneyType)(codeBlock: => Future[Result])
+                       (implicit user: MtdItUser[_], hc: HeaderCarrier): Future[Result] = {
     withIncomeSourcesFS {
-      journeyChecker(journeyType).flatMap {
-        case true => user.userType match {
-          case Some(Agent) => Future.successful(Redirect(controllers.incomeSources.add.routes.YouCannotGoBackErrorController.showAgent(journeyType.businessType)))
-          case _ => Future.successful(Redirect(controllers.incomeSources.add.routes.YouCannotGoBackErrorController.show(journeyType.businessType)))
+      customSessionStarted(journeyType).flatMap { state =>
```
joesheridan, line 40: Can we not get the whole data UIJourneySessionData and pass back to the code block, so we only have to do this once? Otherwise we are duplicating the mongodb calls...
something like:
```
withSessionData {
   sessionData => 
      ...stuff
}
```

app/utils/JourneyChecker.scala
```diff
@@ -32,23 +33,48 @@ trait JourneyChecker extends IncomeSourcesUtils {
...
       }
     }
   }
 
-  private def journeyChecker(journeyType: JourneyType)(implicit hc: HeaderCarrier): Future[Boolean] = {
-    sessionService.getMongoKeyTyped[Boolean](AddIncomeSourceData.hasBeenAddedField, journeyType).flatMap {
-      case Right(Some(true)) => Future(true)
+  private def customSessionStarted(sessionData: UIJourneySessionData): Future[Boolean] = {
+    sessionData.addIncomeSourceData match {
+      case Some(value) if value.hasBeenAdded.contains(true) => Future(true)
```
EdwardDennis, line 59: There is no need to directly create a future like this unless running a long-running async computation. This can lead to thread blocking and performance issues, see:
https://blog.colinbreck.com/future-successful-blocking-your-future-success/

app/controllers/incomeSources/add/AddBusinessNameController.scala
```diff
@@ -70,12 +70,15 @@ class AddBusinessNameController @Inject()(authenticate: AuthenticationPredicate,
...
                              (implicit user: MtdItUser[_]): Future[Option[String]] = {
     if (isChange)
-      sessionService.getMongoKeyTyped[String](AddIncomeSourceData.businessNameField, journeyType).flatMap {
-        case Right(nameOpt) => Future.successful(nameOpt)
-        case Left(ex) => Future.failed(ex)
+      sessionData match {
+        case Some(session) => session.addIncomeSourceData match {
+          case Some(data) => Future.successful(data.businessName)
+          case None => Future.failed(new Error("No session data found"))
+        }
```
EdwardDennis, line 80: The checks for this have already been executed in the utils file. Throwing an additional error here could result in loss of stack trace. It's not necessary to generate errors throughout every level of the application. If there is a need to throw something, perhaps consider using an exception rather than an error (errors should only be used for JVM/fatal issues).
https://www.baeldung.com/scala/nonfatal-vs-exception#:~:text=Error%20represents%20fatal%20issues%20that,be%20handled%20or%20recovered%20from.

app/utils/JourneyChecker.scala
```diff
@@ -32,23 +33,48 @@ trait JourneyChecker extends IncomeSourcesUtils {
...
+  def withCustomSession(journeyType: JourneyType)(codeBlock: Option[UIJourneySessionData] => Future[Result])
+                       (implicit user: MtdItUser[_], hc: HeaderCarrier): Future[Result] = {
     withIncomeSourcesFS {
-      journeyChecker(journeyType).flatMap {
-        case true => user.userType match {
-          case Some(Agent) => Future.successful(Redirect(controllers.incomeSources.add.routes.YouCannotGoBackErrorController.showAgent(journeyType.businessType)))
-          case _ => Future.successful(Redirect(controllers.incomeSources.add.routes.YouCannotGoBackErrorController.show(journeyType.businessType)))
-        }
-        case false => codeBlock
+      sessionService.getMongo(journeyType.toString).flatMap{
```
EdwardDennis, line 40: Not formatted

app/controllers/incomeSources/add/AddBusinessNameController.scala
```diff
@@ -70,13 +70,10 @@ class AddBusinessNameController @Inject()(authenticate: AuthenticationPredicate,
...
 
-  private def getBusinessName(isChange: Boolean)
+  private def getBusinessName(isChange: Boolean, sessionData: Option[UIJourneySessionData])
                              (implicit user: MtdItUser[_]): Future[Option[String]] = {
     if (isChange)
-      sessionService.getMongoKeyTyped[String](AddIncomeSourceData.businessNameField, journeyType).flatMap {
-        case Right(nameOpt) => Future.successful(nameOpt)
-        case Left(ex) => Future.failed(ex)
-      }
+      Future.successful(sessionData.get.addIncomeSourceData.get.businessName)
```
EdwardDennis, line 76: Again, there is no need to create a future here. The sessionData is already available.

PR #2013, opened 2023-12-04
MISUV-6723: QuarterType field added

SiddharthAtos:


app/models/incomeSourceDetails/QuarterTypeElection.scala
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.incomeSourceDetails
+
+import play.api.libs.json.{Format, Json}
+
+case class QuarterTypeElection(quarterReportingType: String, taxYearofElection: String) {
+  val isQuarterReportingTypeStandard: Option[Boolean] = {
```
EdwardDennis, line 22: Should fix:
I think 'isStandardQuarterlyReporting' is more readable

app/views/incomeSources/manage/ManageIncomeSourceDetails.scala.html
```diff
@@ -174,6 +177,46 @@
...
+            val content = HtmlContent(p1.toString() + p2.toString() + anchor.toString())
+
+            govukDetails(Details(id = Some("expandable-info"),
+                                summary = Text(messages(s"incomeSources.manage.quarterly-period.$quarterTypeMessageKey.summary")),
+                                content = content))
+        case _ =>
+    }
+
+
+}
```
EdwardDennis, line 218: Should fix: can be simplified by using Option.map in order to avoid having redundant None / default case

EdwardDennis: It returns Some when defined, otherwise it returns None. This can be tidied up to avoid empty default cases which may hide missed cases

app/models/incomeSourceDetails/QuarterTypeElection.scala
```diff
@@ -0,0 +1,45 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.incomeSourceDetails
+
+import play.api.libs.json.{Format, Json}
+
+trait QuarterReportingType {
```
EdwardDennis, line 21: Can you make this sealed? This gives us a bit more security so it cannot be extended in other files

app/models/incomeSourceDetails/QuarterTypeElection.scala
```diff
@@ -0,0 +1,45 @@
...
+  val value: String
+}
+
+object Standard extends QuarterReportingType {
+  val value: String = "standard"
+}
+
+object Calendar extends QuarterReportingType {
+  val value: String = "calendar"
```
EdwardDennis, line 30: You could have these values as uppercase (as per API) and then reference them directly in the factory pattern match below
e.g. case Standard.value => Some(Standard)

EdwardDennis (approved):

PR #2037, opened 2023-12-14
MISUV-6812 Remove duplicate Unit tests for Income Source Journeys - Add pt 2

danielmaltz1:


app/controllers/incomeSources/manage/CannotGoBackErrorController.scala
```diff
@@ -46,7 +46,7 @@ class CannotGoBackErrorController @Inject()(val checkSessionTimeout: SessionTime
                                             mcc: MessagesControllerComponents,
                                             val ec: ExecutionContext,
                                             val itvcErrorHandler: ItvcErrorHandler,
-                                            val itvcErrorHandlerAgent: AgentItvcErrorHandler) extends ClientConfirmedController with IncomeSourcesUtils {
+                                            val itvcErrorHandlerAgent: AgentItvcErrorHandler) extends ClientConfirmedController with IncomeSourcesUtils with JourneyChecker{
```
EdwardDennis, line 49: Should fix:
not formatted

app/controllers/incomeSources/add/AddIncomeSourceStartDateController.scala
```diff
@@ -91,7 +91,7 @@ class AddIncomeSourceStartDateController @Inject()(authenticate: AuthenticationP
 
     val messagesPrefix = incomeSourceType.startDateMessagesPrefix
 
-    withSessionData(JourneyType(Add, incomeSourceType)) { _ =>
+    withSessionData(JourneyType(Add, incomeSourceType), initialPage = {if (incomeSourceType != SelfEmployment) true else false}) { _ =>
```
EdwardDennis, line 94: Should fix:
You can simplify by passing the param as incomeSourceType != SelfEmployment, if/else is redundant

app/utils/JourneyChecker.scala
```diff
@@ -79,31 +71,47 @@ trait JourneyChecker extends IncomeSourcesUtils {
...
+
   private def useDefaultRedirect(data: UIJourneySessionData, journeyType: JourneyType, midwayFlag: Boolean): Boolean = {
     journeyType.operation match {
       case Add => !(midwayFlag && data.addIncomeSourceData.flatMap(_.incomeSourceAdded).getOrElse(false))
       case _ => true
     }
   }
 
-  def withSessionData(journeyType: JourneyType, midwayFlag: Boolean = true)(codeBlock: UIJourneySessionData => Future[Result])
+  def withSessionData(journeyType: JourneyType, midwayFlag: Boolean = true, initialPage: Boolean = false, nonLoopFlag: Boolean = false)(codeBlock: UIJourneySessionData => Future[Result])
```
avelx, line 94: midwayFlag and initialPage representing state where user at the moment within the actual workflow, these can be combined into a single enumeration:

```
   State | midwayFlag | initialPage | WorkflowState
   A     |   false    | false   | None
   B     |   true     | false   | MidWay
   C     |   false    | true    | InitialPage
   D     |   true     | true    | Not possible scenario
```

avelx: I would suggest to avoid using defaults in this method, its fine when we resolving compile time errors;

avelx (approved):
approved

PR #2048, opened 2023-12-20
MISUV-6753 View & Change | TxM events | Agent-Client UTR (success/failure)

tharunprasath:


app/audit/models/EnterClientUTRAuditModel.scala
```diff
@@ -0,0 +1,53 @@
...
+        "failureReason" -> "API returned error - unable to login agent"
+      )
+  )
+
+  private val outcome: JsObject = if (isSuccessful) successOutcome else failureOutcome
+
+  private val userDetailsJson: JsObject = Json.obj(
+    "nino" -> nino,
+    "mtditid" -> mtditid,
+    "agentReferenceNumber" -> arn,
```
simondrugan16, line 45: :thinking: could we use ++ syntax for Optional fields (arn and credId) so that they are removed from the audit if empty?

app/controllers/agent/EnterClientsUTRController.scala
```diff
@@ -108,4 +113,16 @@ class EnterClientsUTRController @Inject()(enterClientsUTR: EnterClientsUTR,
         }
       )
   }
+
+  private def addAudit(isSuccessful: Boolean, user: IncomeTaxAgentUser, validUTR: String, nino: String, mtdItId: String)(implicit request: Request[_]): Unit = {
```
simondrugan16, line 117: :pick: could we rename to createAudit/sendAudit/something else perhaps?

simondrugan16:
🔧 I think we should have some ITs covering these audits too, using
`AuditStub.verifyAuditEvent`
method we are using for other TxM events

app/controllers/agent/EnterClientsUTRController.scala
```diff
@@ -90,11 +93,13 @@ class EnterClientsUTRController @Inject()(enterClientsUTR: EnterClientsUTR,
                     SessionKeys.clientNino -> nino,
                     SessionKeys.clientUTR -> validUTR
                   ) ++ firstName.map(SessionKeys.clientFirstName -> _) ++ lastName.map(SessionKeys.clientLastName -> _)
```
avelx, line 95: :pushpin: - this might be not a good idea to rely on SessionKeys, overall tendency would be to remove/decommission these from the code base as a redundant concept;

simondrugan16: I think this data we are still storing in cookies. Maybe we could make a separate ticket to migrate this to mongo since this seems to be outside the scope of the ticket?

avelx: oh .. there must a ticket already for this but its not directly related:
https://jira.tools.tax.service.gov.uk/browse/MISUV-6408

PR #2049, opened 2023-12-20
MISUV-6199 Create Dual Auth Function as Trait/Util and implement in every controller file

danielmaltz1:


app/utils/Authenticator.scala
```diff
@@ -0,0 +1,58 @@
...
+                              val itvcErrorHandler: AgentItvcErrorHandler,
+                              val ec: ExecutionContext) extends ClientConfirmedController with I18nSupport {
+
+
+  def authenticatedAction(isAgent: Boolean)(authenticatedCodeBlock: MtdItUser[_] => Future[Result]): Action[AnyContent] = {
+    if (isAgent)
+      Authenticated.async {
+        implicit request =>
+          implicit user =>
+            getMtdItUserWithIncomeSources(incomeSourceDetailsService).flatMap { implicit mtdItUser =>
```
avelx, line 48: :seedling: - something we need to address in the future, passing service as a parameter might be not the best choice as it can be injected or it can actually implement this function itself;
ie. IncomeSourceDetailsService -> getMtdItUserWithIncomeSources (requiredParams) -> return required incomeSources

avelx (approved):

PR #2050, opened 2023-12-20
MISUV-6831: changed field mapping from effectiveDateOfPayment to docu…

SiddharthAtos:


app/models/financialDetails/FinancialDetailsResponseModel.scala
```diff
@@ -36,7 +35,7 @@ case class FinancialDetailsModel(balanceDetails: BalanceDetails,
       financialDetails.find { fd =>
         fd.transactionId.contains(documentDetail.transactionId) &&
           fd.taxYear.toInt == documentDetail.taxYear
-      }.flatMap(_ => documentDetail.effectiveDateOfPayment)
+      }.flatMap(_ => documentDetail.documentDueDate)
```
joesheridan, line 38: this function makes no sense - we don't need to query the financialDetails object for the documentDetails->documentDueDate
can we move this function to DocumentDetails model and remove the financialDetails references?

SiddharthAtos: what about transaction Id validation with Financial Details? not needed?

PR #2055, opened 2023-12-28
MISUV-6754 View & Change | TxM events | Confirm client details

tharunprasath:


avelx:
@tharunprasath  - please add required it:tests

app/audit/models/ConfirmClientDetailsAuditModel.scala
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package audit.models
+
+import play.api.libs.json.{JsObject, JsValue, Json}
+import utils.Utilities.JsonUtil
+
+case class ConfirmClientDetailsAuditModel(clientName: Option[String], nino: Option[String],
```
avelx, line 22: I would suspect that all these fields like: 
* clientName, nino, arn etc mandatory in this context, so if agent manage to login and select client UTR we should have accurate picture of associated client

app/controllers/agent/ConfirmClientUTRController.scala
```diff
@@ -49,6 +52,14 @@ class ConfirmClientUTRController @Inject()(confirmClient: confirmClient,
 
   def submit: Action[AnyContent] = Authenticated.async { implicit request =>
     implicit user =>
+      auditingService.extendedAudit(ConfirmClientDetailsAuditModel(
```
avelx, line 55: assuming that clients details after successful login will be in the context, we can change next code to something like:
```
for {
        clientName <- fetchClientName
        nino <- request.session.get(SessionKeys.clientNino)
        clientMTDID <- request.session.get(SessionKeys.clientMTDID)
        arn <- user.agentReferenceNumber
        saUtr <- request.session.get(SessionKeys.clientUTR)
      } yield
        auditingService.extendedAudit(ConfirmClientDetailsAuditModel(
          clientName = clientName,
          nino = nino,
          mtditid = clientMTDID,
          arn = arn,
          saUtr = saUtr,
          credId = user.credIdd
        )
```

avelx (approved):
approved

PR #2065, opened 2024-01-05
MISUV-6841 - Date Refactor (Concept)

JordHemsworth:


avelx (changes requested):
remove usage of LocalDate.now in the spec files

app/models/incomeSourceDetails/TaxYear.scala
```diff
@@ -49,5 +49,15 @@ object TaxYear {
       case _ => None
     }
   }
+
+  def isValidTaxYear(year: Option[TaxYear], latencyDetails: Option[LatencyDetails]): Boolean = {
```
avelx, line 53: function should look like:
`isValidTaxYear(year: TaxYear, latencyDetails: LatencyDetails): Boolean`

where these are Optional you can use flatMap or for-comprehension to use main function

PR #2066, opened 2024-01-05
MISUV-6851 - Fix duplicated code in IncomeSourcesAccountingMethodController

danielmaltz1:


avelx:
@danielmaltz1  - I dont see any code duplication addressed in this story, if you can point lines pls

PR #2068, opened 2024-01-05
MISUV-6841 - Validate taxYear and changeTo query params from Manage

JordHemsworth:


app/models/incomeSourceDetails/TaxYear.scala
```diff
@@ -49,5 +49,15 @@ object TaxYear {
       case _ => None
     }
   }
+
+  def isValidTaxYear(year: TaxYear, latencyDetails: LatencyDetails): Boolean = {
```
EdwardDennis, line 53: A better name would be isValidLatencyYear

app/models/incomeSourceDetails/TaxYear.scala
```diff
@@ -49,5 +49,15 @@ object TaxYear {
...
+
+  def isValidLatencyYear(year: TaxYear, latencyDetails: LatencyDetails): Boolean = {
+    (latencyDetails, year) match {
+      case (details, year) =>
+        val latencyYears = Set(details.taxYear1, details.taxYear2)
+        latencyYears.contains(year.endYear.toString)
+      case _ => false
+    }
+  }
+
```
EdwardDennis, line 61: I think this logic would be more appropriate in LatencyYear.scala, this file appears to have generic tax year logic which for some reason is within incomeSourceDetails (it shouldn't be)

avelx: @EdwardDennis  - this is really depends on domain modelling rule in place, we don't have any atm as far as I know;

EdwardDennis: @avelx we might not have explicit domain modelling rules at the moment, but it's a good idea to start considering them to improve the maintainability of our codebase. Keeping code modular and logically structured is important. The concept of a tax year is generic and applies to more than just income sources, so encapsulating this within incomeSourceDetails could lead to confusion. Similarly, logic specific to latency years is better placed in a more intuitive location. This will make it easier for developers to understand the structure and locate the necessary code and reduces the likelihood of duplicate code in the future.

avelx: If you have any specific approach, as ie. above how these primitives would need to located / or related to each other; thats fine, you can come up with the proposal how to do this in the future, as long as this is generic this would good starting point to start applying these

EdwardDennis (approved):

PR #2072, opened 2024-01-10
MISUV-6780-txm-response-data-too-long

simondrugan16:


app/audit/AuditingService.scala
```diff
@@ -70,6 +73,19 @@ class AuditingService @Inject()(appConfig: FrontendAppConfig, auditConnector: Au
     }
   }
 
+  private def eventCheckLength(extendedDataEvent: ExtendedDataEvent): Unit = {
+    if (extendedDataEvent.toString.length > 32665) {
```
avelx, line 77: might be a good idea to move this "magic number" into app config ?
so, this would look like:

`if (extendedDataEvent.toString.length > appConfig.maxAuditEventSizeBytes)`

PR #2076, opened 2024-01-12
MISUV-6932: Enable No Open Obligations on FE

EdwardDennis:


avelx:
Assuming we are changing logic in the Home controller, I would guess that next story would not be longer relevant:
https://jira.tools.tax.service.gov.uk/browse/MISUV-6776

EdwardDennis:
> Assuming we are changing logic in the Home controller, I would guess that next story would not be longer relevant: https://jira.tools.tax.service.gov.uk/browse/MISUV-6776

Prototype migration? Still looks relevant to me. This PR addresses where a user has signed up for future year or is no longer enrolled in MTD/SA so has no future obligations.

avelx:
> > Assuming we are changing logic in the Home controller, I would guess that next story would not be longer relevant: https://jira.tools.tax.service.gov.uk/browse/MISUV-6776
> 
> Prototype migration? Still looks relevant to me. This PR addresses where a user has signed up for future year or is no longer enrolled in MTD/SA so has no future obligations.

I was actually referring to https://jira.tools.tax.service.gov.uk/browse/MISUV-6907

PR #2078, opened 2024-01-16
MISUV-6877: SPP009b | Display 'available credit' on V&C homepage tile

ayubAz:


app/views/Home.scala.html
```diff
@@ -37,7 +37,9 @@
 	home_tile: home_tile
 )
 
-@(nextPaymentDueDate: Option[LocalDate],
+@(
```
avelx, line 40: I am fine with this approach, but would be nice to move all these parameters into a single view model

avelx (changes requested):
as per comment, few minor changes requested

avelx (approved):
approved

PR #2081, opened 2024-01-17
DO NOT MERGE / for ref only

avelx:
as per PR review


app/views/TaxYearSummary.scala.html
```diff
@@ -277,27 +277,34 @@ <h3 class="govuk-heading-m">
...
-                    <th scope="row" class="govuk-table__header govuk-!-font-weight-regular">
-                        <div>@updateTypeHelper(nextUpdateWithIncomeType.obligation.obligationType)</div>
-                    </th>
-                    <td class="govuk-table__cell">@obligationsTypeHelper@{messages(nextUpdateWithIncomeType.incomeType)}</td>
-                    <td class="govuk-table__cell">@nextUpdateWithIncomeType.obligation.dateReceived.map(_.toLongDateShort)</td>
-                </tr>
-            }
-            </tbody>
-        </table>
+        @dueObligations.map { case (period, obligations) =>
```
avelx, line 280: I think we can rename 'period' into quarterlyTypeName ?

PR #2083, opened 2024-01-18
MISUV-6706 fix for standard + calendar quarters

danielmaltz1:


app/views/helpers/injected/obligations/NextUpdatesHelper.scala.html
```diff
@@ -69,27 +70,45 @@
...
+                head = Some(Seq(
+                    HeadCell(
+                        content = Text(messages("nextUpdates.table.header.type"))
+                    ),
+                    HeadCell(
+                        content = Text(messages("nextUpdates.table.header.source"))
+                    )
+                )),
+                caption = {
+                    if(validObligations.head.obligation.obligationType == "Quarterly")
```
avelx, line 89: consider next change to make this code more safe with respect to data in the collection:

```
caption = {
                    validObligations.headOption.map(x => x.obligation.obligationType) match {
                        case Some("Quarterly") =>
                            validObligations.headOption.map(x => getTableHeadingMessage(x.obligation))
                        case _ => None
                    }
                },
```

app/views/helpers/injected/obligations/NextUpdatesHelper.scala.html
```diff
@@ -69,27 +70,45 @@
...
-        attributes = Map("id"->"next-updates-table")
-    )
-    )
+            )
+        case _ => HtmlFormat.empty
+    }
+
+}
+
+@bodyContent(collectedObligations: Seq[NextUpdateModelWithIncomeType]) = @{
```
avelx, line 100: might worth to move this within service layer so this can be tested for various scenarios;
there is some complexity involved in here and also worth to consider to change view model (add new view model if needed), depending how much time we can spend on this in this sprint

app/views/helpers/injected/obligations/NextUpdatesHelper.scala.html
```diff
@@ -69,27 +70,45 @@
...
+    if (collectedObligations.head.obligation.obligationType == "Quarterly") {
```
avelx, line 101: try avoid .head as much as possible, even its seems safe based on the caller logic, next change 
would work well I think:

`if (collectedObligations.headOption.map(_.obligation.obligationType) == Some("Quarterly") `

app/views/helpers/injected/obligations/NextUpdatesHelper.scala.html
```diff
@@ -69,27 +69,47 @@
...
+            )
+        case _ => HtmlFormat.empty
+    }
+}
+
+@bodyContent(collectedObligations: Seq[NextUpdateModelWithIncomeType]) = @{
+    if (collectedObligations.headOption.map(_.obligation.obligationType).contains("Quarterly")) {
+        val obligations = collectedObligations.partition(ob => ob.obligation.start == collectedObligations.head.obligation.start)
+        obligations match {
+            case (x: Seq[NextUpdateModelWithIncomeType], y: Seq[NextUpdateModelWithIncomeType]) if x.nonEmpty && y.nonEmpty =>
```
EdwardDennis, line 105: Should fix: reintroducing single character references

PR #2092, opened 2024-01-23
[DO NOT MERGE] implement logger util for AddBusinessName

ayubAz:


app/controllers/incomeSources/add/LoggerUtil.scala
```diff
@@ -0,0 +1,53 @@
...
+import play.api.mvc.Result
+
+trait LoggerUtil { self =>
+
+  val fileName: String = self.getClass.getSimpleName
+
+  lazy val methodName: String =
+    Thread.currentThread.getStackTrace()(14).getMethodName.split('$').toList.reverse match {
+      case List(_, methodName, _*) => methodName
+      case _ => throw new Exception("could not get method name")
```
avelx, line 31: might worth to use some default value instead, raising exception within logger logic is a bit risky;

app/controllers/incomeSources/add/LoggerUtil.scala
```diff
@@ -0,0 +1,53 @@
...
+package controllers.incomeSources.add
+
+import auth.MtdItUser
+import config.{AgentItvcErrorHandler, ItvcErrorHandler}
+import play.api.i18n.Lang.logger
+import play.api.mvc.Result
+
+trait LoggerUtil { self =>
+
+  val fileName: String = self.getClass.getSimpleName
```
avelx, line 26: just curios how this work with respect to of usage within Scala: object and trait's; ie. if we extended trait C with LoggerUtil and trait C is used by service A, what would be the name of the class ? or if we use this within an object ??

PR #2099, opened 2024-02-05
MISUV-7025: Implement Mongo Encryption of PII data for income sources - ADD Journey [in-scope]

ayubAz:


app/models/incomeSourceDetails/BusinessAddressModel.scala
```diff
@@ -18,16 +18,10 @@ package models.incomeSourceDetails
...
-}
-
-case class BusinessAddressModel(auditRef: String,
-                                address: Address)
-
-object Address {
-  implicit val format: OFormat[Address] = Json.format[Address]
-}
+case class BusinessAddressModel(
+                                 auditRef: String,
```
avelx, line 22: as I understand this field (auditRef), is used only in scope of the tests, so I would question if its really something required?

app/repositories/SensitiveUIJourneySessionDataRepository.scala
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  def keepAlive(data: SensitiveUIJourneySessionData): Future[Boolean] =
+    collection
+      .updateOne(
+        filter = dataFilter(data),
+        update = Updates.set("lastUpdated", Instant.now(clock))
+      )
+      .toFuture()
+      .map(_ => true)
```
avelx, line 81: worth to change signature to: Future[Unit], as we always returning value of true for some reason

app/repositories/SensitiveUIJourneySessionDataRepository.scala
```diff
@@ -0,0 +1,126 @@
...
+    val updatedAnswers = data copy (lastUpdated = Instant.now(clock))
+
+    collection
+      .replaceOne(
+        filter = dataFilter(data.encrypted),
+        replacement = updatedAnswers.encrypted,
+        options = ReplaceOptions().upsert(true)
+      )
+      .toFuture()
+      .map(_ => true)
```
avelx, line 105: same, worth to change signature to Future[Unit]

build.sbt
```diff
@@ -41,6 +41,7 @@ def test(scope: String = "test"): Seq[ModuleID] = Seq(
   "org.scalatestplus.play" %% "scalatestplus-play" % scalaTestPlusVersion % scope,
   "org.scalamock" %% "scalamock" % scalaMockVersion % scope,
   "org.pegdown" % "pegdown" % pegdownVersion % scope,
+  "uk.gov.hmrc" %% "crypto-json-play-28" % "7.6.0",
```
avelx, line 44: do we have something to support Play 3.0? as we migrated to higher version as the moment

conf/application.conf
```diff
@@ -294,3 +294,10 @@ bootstrap.http.headersAllowlist = [ "Gov-Test-Scenario", "Location", "X-Request-
 internalServiceHostPatterns = [ "localhost" ]
 
 accessibility-statement.service-path = "/income-tax-view-change"
+
+###################### DECIDE WHAT TO SET ENCRYPTION KEY TO ###############################
+useEncryption = false
```
avelx, line 299: it would make sense to put these together with the key, like:

```
encryption {
  useEncryption = false
  key = "QmFyMTIzNDVCYXIxMjM0NQ=="
}
```

app/controllers/incomeSources/add/AddBusinessNameController.scala
```diff
@@ -150,11 +148,30 @@ class AddBusinessNameController @Inject()(val authorisedFunctions: AuthorisedFun
...
+                    Some(
+                      AddIncomeSourceData(
+                        businessName = Some(formData.name)
+                      )
+                    )
+                )
+            }
+          ) flatMap {
+            case true  => Future.successful(Redirect(getRedirect(isAgent, isChange)))
+            case false => Future.failed(new Exception("Mongo update call was not acknowledged"))
```
avelx, line 174: false case will be always redundant, as setMongoData will never return this value;

PR #2103, opened 2024-02-05
MISUV-7063 'Payment and refund history' page heading, title and V&C homepage tile link update

EdwardDennis:


app/models/homePage/PaymentCreditAndRefundHistoryTileViewModel.scala
```diff
@@ -0,0 +1,45 @@
...
+      }.getOrElse(BigDecimal(0.00)))
+    } else None
+
+  def title()(implicit messages: Messages): String = if (paymentHistoryRefundsEnabled) {
+    messages("home.paymentHistoryRefund.heading")
+  } else {
+    messages("home.paymentHistory.heading")
+  }
+
+  def paymentCreditRefundHistoryMessageKey(): String = (creditsRefundsRepayEnabled, paymentHistoryRefundsEnabled) match {
```
SiddharthAtos, line 38: Why not use implicit `messages`  hear as well like you did in `PaymentCreditAndRefundHistoryViewModel.scala`

EdwardDennis: The link component takes the message key
See here:
https://github.com/hmrc/income-tax-view-change-frontend/blob/main/app/views/components/link.scala.html

PR #2104, opened 2024-02-05
Misuv 7025 no encryption boolean

ayubAz:


avelx (approved):
approved with comment

app/services/SessionService.scala
```diff
@@ -81,7 +94,10 @@ class SessionService @Inject()(uiJourneySessionDataRepository: UIJourneySessionD
 
   def setMongoData(uiJourneySessionData: UIJourneySessionData)
                   (implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Boolean] = {
-    uiJourneySessionDataRepository.set(uiJourneySessionData)
+    if (config.encryptionIsEnabled)
```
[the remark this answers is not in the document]
avelx: the problem is that we never return false from this method, so use of Boolean type redundant

conf/application.conf
```diff
@@ -294,3 +294,8 @@ bootstrap.http.headersAllowlist = [ "Gov-Test-Scenario", "Location", "X-Request-
 internalServiceHostPatterns = [ "localhost" ]
 
 accessibility-statement.service-path = "/income-tax-view-change"
+
+encryption {
+  key = "QmFyMTIzNDVCYXIxMjM0NQ=="
+  isEnabled = false
+}
```
EdwardDennis, line 301: I think this should be stored in app-config-base instead
See here, encryption not to be stored in application.conf
https://github.com/hmrc/app-config-base/blob/main/documentation/README.md#applicationconf

EdwardDennis: This key should be changed as well, given that this repo is public

PR #2108, opened 2024-02-07
MISUV-6969 - Prevent Submission Handoff for years Prior to 22/23 on Tax Years List Page

danielmaltz1:


app/views/TaxYears.scala.html
```diff
@@ -51,10 +51,12 @@
 }
 
 @updateReturnLink(year: Int) = {
-    <a class="govuk-link" id="updateReturn-link-@year" href="@appConfig.submissionFrontendTaxYearsPage(year)">
-        @messages("taxYears.updateReturn")
-        <span class="govuk-visually-hidden">@messages("taxYears.taxYears", s"${year - 1}", s"$year")</span>
-    </a>
+    @if(year > 2022) {
```
avelx, line 54: if we can be specific what type of year we have here: calendar year or tax year, this would help understand this local context;

avelx (approved):
approved

PR #2123, opened 2024-02-15
MISUV-7088-v2 - Design changes for credits list on 'Claim a refund' page

JordHemsworth:


owen-gilfellon (dismissed):
Looks good - I've a few comments about general style (.gets on Options should be avoided, Optional fields that look like they shouldn't be Optional), but they all pre-existed this ticket, so I'll bring it up at sprint retro 🙂

owen-gilfellon (approved):

PR #2126, opened 2024-02-16
MISUV-7124: Include 'Tax year' to credits on history page and sort on 'Date' in descending order DO NOT MERGE

danielmaltz1:


app/models/repaymentHistory/PaymentHistoryEntry.scala
```diff
@@ -17,10 +17,23 @@
...
 
 case class PaymentHistoryEntry(date: LocalDate,
                                description: String,
                                amount: Option[BigDecimal],
                                transactionId: Option[String] = None,
                                linkUrl: String,
-                               visuallyHiddenText: String)
+                               visuallyHiddenText: String) {
+
+  def getTaxYearEndYear: Int = {
```
avelx, line 29: I believe this is something already implemented as a part of DateServiceInterface or you can add new method within DateService to expose this  logic; essentially all operations with LocalDate we would expect to be localised within this DateService

app/views/PaymentHistory.scala.html
```diff
@@ -78,14 +78,18 @@
...
             <tr class="govuk-table__row">
                 <td class="govuk-table__cell">@payment.date.toLongDate</td>
                 <td class="govuk-table__cell">
                     <a class="govuk-link" href="@payment.linkUrl">
                         @messages(payment.description)
                         <span class="govuk-visually-hidden">@payment.visuallyHiddenText @if(yearPayments.size > 1){ Item @(index + 1)}</span>
                     </a>
+                    <br>
+                    <a class="govuk-text">
+                        @messages("paymentHistory.taxYear", (payment.getTaxYearEndYear-1).toString, payment.getTaxYearEndYear.toString)
```
avelx, line 91: We finally need to defined correct TaxYear type in order to remove usage of this pattern: line 91;
I would say its optional to apply this at this stage;

---------------------
PaymentHistory model should expose TaxYear property/model
TaxYear model will expose startYear: Int/String and endYear: Int/String

Here is example:
https://jira.tools.tax.service.gov.uk/browse/MISUV-6559

app/controllers/PaymentHistoryController.scala
```diff
@@ -55,7 +55,7 @@ class PaymentHistoryController @Inject()(val paymentHistoryView: PaymentHistory,
                                          val repaymentService: RepaymentService,
                                          val retrieveNinoWithIncomeSources: IncomeSourceDetailsPredicate)
                                         (implicit val appConfig: FrontendAppConfig,
-                                         dateService: DateServiceInterface,
+                                         dateService: DateService,
```
avelx, line 58: this can remain DateServiceInterface,
in case you extending, adding new method etc, you would need to apply changes in all dependencies

app/models/repaymentHistory/RepaymentHistoryUtils.scala
```diff
@@ -62,7 +63,7 @@ object RepaymentHistoryUtils {
 
   def getGroupedPaymentHistoryData(payments: List[Payment], repayments: List[RepaymentHistory], isAgent: Boolean,
                                    MFACreditsEnabled: Boolean, CutOverCreditsEnabled: Boolean, languageUtils: LanguageUtils
-                                  )(implicit messages: Messages): List[(Int, List[PaymentHistoryEntry])] = {
+                                  )(implicit messages: Messages, dateService: DateService): List[(Int, List[PaymentHistoryEntry])] = {
```
avelx, line 66: Please use DateServiceInterface instead, this would give us option to switch actual implementation under test context

avelx (approved):
approved

PR #2128, opened 2024-02-19
MISUV-7163: time machine v2 / DO NOT MERGE

avelx:


app/services/DateService.scala
```diff
@@ -19,42 +19,49 @@ package services
...
-  def getCurrentDate(isTimeMachineEnabled: Boolean = false): LocalDate = {
-    if (isTimeMachineEnabled) {
-      frontendAppConfig
-        .timeMachineAddYears.map(LocalDate.now().plusYears(_))
-        .getOrElse(LocalDate.now())
+  // TODO: instantiate clock depending on the configuration settings
+  private val timeMachineIsOn: Boolean = frontendAppConfig.timeMachineAddYears.nonEmpty
+
+  lazy val clock: Clock = if (timeMachineIsOn) {
+    Clock.offset(Clock.systemDefaultZone.withZone(ZoneOffset.UTC), Duration.ofDays(100))
```
simondrugan16, line 33: Would this be 365 days, not 100?

app/services/DateService.scala
```diff
@@ -19,42 +19,49 @@ package services
...
 
-  def getCurrentDate(isTimeMachineEnabled: Boolean = false): LocalDate = {
-    if (isTimeMachineEnabled) {
-      frontendAppConfig
-        .timeMachineAddYears.map(LocalDate.now().plusYears(_))
-        .getOrElse(LocalDate.now())
+  // TODO: instantiate clock depending on the configuration settings
+  private val timeMachineIsOn: Boolean = frontendAppConfig.timeMachineAddYears.nonEmpty
+
+  lazy val clock: Clock = if (timeMachineIsOn) {
```
simondrugan16, line 32: What is the benefit of using the clock here rather than keeping Localdate.now() as it is (and if time machine is on then add a year)?

avelx: this is yet another way to simulate fixed time or implement timeMachine, have a look at the method which Clock instance can provide, you will see what can be done, ie. for unit tests etc

simondrugan16: Yep, I think it's a good idea. Do we have a separate ticket for implementing this?

PR #2130, opened 2024-02-21
MISUV-7076 SCP025 | Update Income source tile

danielmaltz1:


avelx:
If we can update **HomeControllerISpec** as well / it:Tests

avelx (approved):

PR #2138, opened 2024-02-26
MISUV-5771: SPP009b | Update 'Claim a refund' page to show soft allocation with content change

ayubAz:
…accordingly

### DIFF
--- a/app/views/CreditAndRefunds.scala.html
+++ b/app/views/CreditAndRefunds.scala.html
@@ -113,9 +113,9 @@ balance.getOrElse(throw MissingFieldException("Balance"))
                         <span class="govuk-body" id="unallocated-payment">
                             <strong class="govuk-body govuk-!-font-weight-bold">@maybeAvailableCredit.toCurrencyString</strong>
                             @messages("credit-and-refund.payment")
-                            @(maybeCreditChargesHead match {
+                            @((maybeCreditChargesHead match {
                                 case (documentDetailWithDueDate, _) => documentDetailWithDueDate.dueDate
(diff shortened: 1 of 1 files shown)


app/models/financialDetails/BalanceDetails.scala
```diff
@@ -42,6 +43,10 @@ case class BalanceDetails(balanceDueWithin30Days: BigDecimal,
     availableCredit.map (credit => math.abs(credit.toDouble))
   }
 
+  def getAbsoluteAllocatedCreditAmount: Option[BigDecimal] = {
+    allocatedCredit.map(credit => math.abs(credit.toDouble))
```
avelx, line 47: we need to raise this with the BAs to make sure we are doing correct calculation

app/views/CreditAndRefunds.scala.html
```diff
@@ -77,36 +70,107 @@
...
+            </span>
+        </summary>
+        <div class="govuk-details__text">
+            @creditAndRefundType match {
+                case Some(UnallocatedCreditFromOnePayment) => {
+                    <h2 class="govuk-heading-s">
+                        <span class="govuk-body" id="unallocated-payment">
+                            <strong class="govuk-body govuk-!-font-weight-bold">@maybeAvailableCredit.toCurrencyString</strong>
+                            @messages("credit-and-refund.payment")
+                            @maybeCreditChargesHead._1.dueDate.getOrElse(throw MissingFieldException("Due Date")).toLongDate
```
avelx, line 116: if we can avoid using tuples in this way, (._1, _2), as this make code not readable

PR #2144, opened 2024-02-29
Misuv 7077 your businesses page

tharunprasath:


app/models/incomeSourceDetails/BusinessDetailsModel.scala
```diff
@@ -22,6 +22,7 @@ import play.api.libs.json.{Format, Json}
 import java.time.LocalDate
 
 case class BusinessDetailsModel(incomeSourceId: String,
+                                incomeSource: Option[String],
```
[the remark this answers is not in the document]
tharunprasath: No, It's a different field from the API business_data_details.incomeSource. Which holds the business type.

app/models/incomeSourceDetails/viewmodels/ViewIncomeSourcesViewModel.scala
```diff
@@ -26,6 +26,7 @@ case class ViewIncomeSourcesViewModel(viewSoleTraderBusinesses: List[ViewBusines
                                       viewCeasedBusinesses: List[CeasedBusinessDetailsViewModel])
 
 case class ViewBusinessDetailsViewModel(incomeSourceId: IncomeSourceId,
+                                        incomeSource: Option[String],
```
avelx, line 29: why not to use IncomeSourceId defined in line 28 ? you should be able to convert to string where required

tharunprasath: No, It's a different field from the API business_data_details.incomeSource. Which holds the business type.

I mapped the name based on the API field name. Do we need to use "businessType" for more readability?

avelx: in case incomeSource field independent to represent incomeSourceType, so we might need to rename it accordingly, + defined required enum, as we have a full control on the level of viewModel

tharunprasath: IncomeSource is the name of the income. 
Eg: IncomeSource: Fruit Company Ltd
IncomeSourceId: 0010100100101

IncomeSourceType is a different field which denotes the UK/Foreign. This property "IncomeSource" holds the user-defined name of the business which is slight different from the TradeName

avelx: I see, so it some sort of IncomeSourceName ? as incomeSource is a bit to abstract naming convention, but I guess we can leave it as is

avelx (approved):

PR #2146, opened 2024-03-01
MISUV-7208-investigate-qa-balancing-charge-tys

simondrugan16:


app/controllers/ChargeSummaryController.scala
```diff
@@ -88,8 +88,16 @@ class ChargeSummaryController @Inject()(val authenticate: AuthenticationPredicat
       }.foldLeft(FinancialDetailsModel(BalanceDetails(0.00, 0.00, 0.00, None, None, None, None), List(), List()))((merged, next) => merged.mergeLists(next))
 
       val matchingYear: List[FinancialDetailsResponseModel] = financialResponses.collect {
-        case (year, response) if year == taxYear => response
+        case (year, response) if year == taxYear =>
+          Logger("application").debug(s"[ChargeSummaryController][handleRequest] - ID:abcd3 year = $year")
```
avelx, line 92: if we can merge these under single log entry as this we should produce as less log entries as possible

avelx (changes requested):
Please address comment, and if you can remove not relevant markers in the log like: 
abcd5, abcd2, abcd1 etc

avelx (approved):

PR #2153, opened 2024-03-04
Sole trader add business name page - Apostrophe fix

danielmaltz1:


simondrugan16:
There was discussion around escaping characters rather than passing them in. This worked fine for the frontend but did not work well in the ViewSpec for this page as it was not rendering properly when escaping. Passing it in like in this ticket fixed this problem.

PR #2165, opened 2024-03-06
MISUV-5751 FE: code refactoring FallBackBackLinks.scala

ayubAz:


ian-hmrc:
There are no unit tests around this function. Can we get those in first and then check use them to check the refactoring preserves the behaviour?

ian-hmrc (approved):
LGTM

PR #2171, opened 2024-03-08
MISUV-7219 - SCP025 | Add property | Type of property business

danielmaltz1:


app/views/manageBusinesses/add/AddProperty.scala.html
```diff
@@ -0,0 +1,96 @@
...
+    pageTitle = getMessage("heading"),
+    isAgent = isAgent,
+    backUrl = backUrl,
+    btaNavPartial = user.btaNavPartial) {
+
+    @formWithCSRF(action = postAction, Symbol("id") -> "type-of-business") {
+
+        @errorSummary(form)
+
+        @if(pageHasErrors){
```
avelx, line 51: there are high chances to produce incorrect html with such approach, might worth to move this html generation into separate function

app/forms/manageBusinesses/add/AddProprertyForm.scala
```diff
@@ -0,0 +1,46 @@
...
+ * limitations under the License.
+ */
+
+package forms.manageBusinesses.add
+
+import play.api.data.Form
+import play.api.data.Forms._
+
+object AddProprertyForm {
+
```
avelx, line 23: I guess we would need a unit test for this form ?

PR #2177, opened 2024-03-12
MISUV-7258: transitional profit - improved code readability

SiddharthAtos:
MISUV-7258: transitional profit - improved code readability
- moved logic to view model

### DIFF
--- a/app/models/liabilitycalculation/viewmodels/TaxDueSummaryViewModel.scala
+++ b/app/models/liabilitycalculation/viewmodels/TaxDueSummaryViewModel.scala
@@ -137,12 +137,13 @@ object TaxDueSummaryViewModel {
 
 object TransitionProfitRow {
   def apply(incomeTaxCharged: Option[BigDecimal], totalTaxableProfit: Option[BigDecimal]): Option[TransitionProfitRow] = {
+
     (incomeTaxCharged, totalTaxableProfit) match {
       case (Some(tax), Some(profit)) => Some(TransitionProfitRow(incomeTaxCharged = tax, totalTaxableProfit = profit))
-      case (None, _) =>
+      case (None, Some(_)) =>
         Logger("application").warn(s"[TaxDueSummaryViewModel][TransitionProfitRow] missing incomeTaxChargedOnTransitionProfits")
         None
-      case (_, None) =>
+      case (Some(_), None) =>
         Logger("application").warn(s"[TaxDueSummaryViewModel][TransitionProfitRow] missing totalTaxableTransitionProfit")
         None
       case _ => None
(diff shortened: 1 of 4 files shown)


app/models/liabilitycalculation/viewmodels/TaxDueSummaryViewModel.scala
```diff
@@ -125,10 +125,29 @@ object TaxDueSummaryViewModel {
...
+          calc.transitionProfit.flatMap(_.totalTaxableTransitionProfit))
       )
       case None => TaxDueSummaryViewModel()
     }
   }
 }
+
+object TransitionProfitRow {
+  def apply(incomeTaxCharged: Option[BigDecimal], totalTaxableProfit: Option[BigDecimal]): Option[TransitionProfitRow] = {
+    (incomeTaxCharged, totalTaxableProfit) match {
```
ian-hmrc, line 140: Did you consider doing this with flatMap()? I think it could be a 1-liner.

app/models/liabilitycalculation/viewmodels/TaxDueSummaryViewModel.scala
```diff
@@ -125,10 +125,29 @@ object TaxDueSummaryViewModel {
...
+      case (Some(tax), Some(profit)) => Some(TransitionProfitRow(incomeTaxCharged = tax, totalTaxableProfit = profit))
+      case (None, _) =>
+        Logger("application").warn(s"[TaxDueSummaryViewModel][TransitionProfitRow] missing incomeTaxChargedOnTransitionProfits")
+        None
+      case (_, None) =>
+        Logger("application").warn(s"[TaxDueSummaryViewModel][TransitionProfitRow] missing totalTaxableTransitionProfit")
+        None
+      case _ => None
```
ian-hmrc, line 148: This case is not shown as being covered by the tests. I think the None, None case is matching earlier and will be logging inappropriately.

SiddharthAtos: Made the change.

app/models/liabilitycalculation/viewmodels/TaxDueSummaryViewModel.scala
SiddharthAtos: Keep logs will provide as a clue in case of missing fields

ian-hmrc: Yeah. As discussed in the Tech Deep Dive, I wish we didn't need to do this, but let's keep them.

app/models/liabilitycalculation/viewmodels/TaxDueSummaryViewModel.scala
ian-hmrc: Cool

PR #2179, opened 2024-03-12
[v2]: MISUV-6989: Refactor: HomeController.handleShowRequest method 

ayubAz:


app/controllers/HomeController.scala
```diff
@@ -106,10 +106,7 @@ class HomeController @Inject()(val homeView: views.html.Home,
       paymentsDue              = getDueDates(unpaidCharges)
       dunningLockExistsValue   = unpaidCharges.collectFirst { case fdm: FinancialDetailsModel if fdm.dunningLockExists => true }.getOrElse(false)
       outstandingChargesModel <- getOutstandingChargesModel(unpaidCharges)
-      outstandingChargesDueDate = outstandingChargesModel.flatMap {
-        case OutstandingChargeModel(_, Some(relevantDate), _, _) => List(relevantDate)
-        case _ => Nil
-      }
+      outstandingChargesDueDate  = outstandingChargesModel.collect { case OutstandingChargeModel(_, relevantDate, _, _) => relevantDate}.flatten
```
ian-hmrc, line 111: Would it be worth extracting a function for this like the other cases? Maybe for the dunningLockExists case too?

app/controllers/HomeController.scala
```diff
@@ -110,6 +105,8 @@ class HomeController @Inject()(val homeView: views.html.Home,
       paymentsDueMerged          = (paymentsDue ::: outstandingChargesDueDate).sortWith(_ isBefore _).headOption
     } yield {
 
+      lazy val nextUpdatesTileViewModel = NextUpdatesTileViewModel(nextUpdatesDueDates, dateService.getCurrentDate)
```
ian-hmrc, line 116: What's the benefit of using a lazy value here?

app/controllers/HomeController.scala
```diff
@@ -143,6 +143,12 @@ class HomeController @Inject()(val homeView: views.html.Home,
       case _ => Nil
     }
 
+  private def calculateOverduePaymentsCount(paymentsDue: List[LocalDate], outstandingChargesModel: List[OutstandingChargeModel]): Int = {
+    lazy val overduePaymentsCountFromDate = paymentsDue.count(_.isBefore(dateService.getCurrentDate))
```
ian-hmrc, line 155: What's the benefit of using lazy values here if we're going to use them immediately?

PR #2180, opened 2024-03-12
MISUV-5754: changed useFallbackBackLink default to true. amended the test cases

SiddharthAtos:
changed useFallbackBackLink default to true. Amended the test cases.


JordHemsworth (approved):
Looks good

PR #2183, opened 2024-03-13
MISUV-7285-remove-unused-javascript-files

simondrugan16:
moving testOnly assets to a separate place so they cannot be accessed in production

removing unused javascript files


avelx (changes requested):
as per comment in the PR

simondrugan16:
Closed since restarted PR with the two sections of the ticket separated into 2 PRs. One for removing unused JS files. One for obscuring testOnly assets (JS file on login page)

PR #2186, opened 2024-03-13
MISUV-6989: Refactor: HomeController.handleShowRequest method

ayubAz:


app/auth/MtdItUser.scala
```diff
@@ -65,4 +65,7 @@ case class MtdItUser[A](mtditid: String,
                         saUtr: Option[String],
                         credId: Option[String],
                         userType: Option[AffinityGroup],
-                        arn: Option[String])(implicit request: Request[A]) extends MtdItUserBase[A]
+                        arn: Option[String])(implicit request: Request[A]) extends MtdItUserBase[A] {
+
+  def isAgent: Boolean = userType contains Agent
```
ian-hmrc, line 70: Can we remove this now we've moved back to getting isAgent from the request?

app/models/outstandingCharges/OutstandingChargeModel.scala
```diff
@@ -23,7 +23,10 @@ case class OutstandingChargeModel(chargeName: String,
                                   relevantDueDate: Option[LocalDate],
                                   chargeAmount: BigDecimal,
                                   tieBreaker: Int
-                                 )
+                                 ) {
+
+  def hasRelevantDueDateWithBCDChargeName: Boolean = relevantDueDate.isDefined && chargeName == "BCD"
```
ian-hmrc, line 28: Could we find a better name for this method? It seems to be named after it's implementation. What does this method do in domain terms? Could we use that to get a better name?

app/controllers/HomeController.scala
```diff
@@ -60,107 +60,122 @@ class HomeController @Inject()(val homeView: views.html.Home,
...
+  private def getOutstandingChargesModel(unpaidCharges: List[FinancialDetailsResponseModel])
+                                        (implicit user: MtdItUser[_]): Future[List[OutstandingChargeModel]] =
+    whatYouOweService.getWhatYouOweChargesList(
+      unpaidCharges,
+      isEnabled(CodingOut),
+      isEnabled(MFACreditsAndDebits)
+    ) map {
+      case WhatYouOweChargesList(_, _, Some(OutstandingChargesModel(outstandingCharges)), _) =>
+        outstandingCharges.filter(_.isBalancingChargeDebit)
+                          .filter(_.relevantDueDate.isDefined)
```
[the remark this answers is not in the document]
ian-hmrc: Yes. That's likely to be the case. But if there's an efficiency argument to be made, I'd want to see performance data and know that once version is clearly much better than the other, and that it is in a time-critical portion of the code (which I strongly suspect this isn't as it isn't a complex processing area, or in multiple nested loops).

Moreover, I spotted (but didn't do) a further change that might improve this greatly. I think the outstandingChargesModel and the outstandingChargesDueDates values can be collapsed into by mapping to Due Dates instead of holding onto the OutstandingChargesModel. They are used around lines 112, 113, but these uses can be converted over and make much more sense when based on a list of due dates. This change would need to have those filters separate so one could be converted to map. I'll put a comment on the TD ticket for the splitting as tiles so that this can be looked at at then.

JordHemsworth (approved):

ian-hmrc (approved):
LGTM

PR #2190, opened 2024-03-14
MISUV-7220 SCP025 | Add SE, properties| Businesses that have ceased

tharunprasath:


app/controllers/manageBusinesses/cease/ViewAllCeasedBusinessesController.scala
```diff
@@ -16,34 +16,66 @@
...
+          sessionService.deleteSession(Manage).map { _ =>
+            Ok(viewAllCeasedBusinesses(
+              sources = viewModel,
+              isAgent = isAgent,
+              backUrl = backUrl
+            ))
+          } recover {
+            case ex: Exception =>
+              Logger("application").error(
+                s"[ViewAllCeasedBusinessesController][handleRequest] - Session Error: ${ex.getMessage} - ${ex.getCause}")
```
avelx, line 71: not necessarily a session error, we don't really know what error can be here

avelx (approved):

app/controllers/manageBusinesses/cease/ViewAllCeasedBusinessesController.scala
```diff
@@ -16,34 +16,66 @@
...
-                                                  val checkSessionTimeout: SessionTimeoutPredicate)
-                                                 (implicit val appConfig: FrontendAppConfig,
-                                             mcc: MessagesControllerComponents,
-                                             val ec: ExecutionContext,
-                                             val itvcErrorHandlerAgent: AgentItvcErrorHandler)
-  extends ClientConfirmedController with FeatureSwitching with I18nSupport {
-
-  def show(): Action[AnyContent] = Action {
-    Ok("")
+  def show(isAgent: Boolean): Action[AnyContent] = auth.authenticatedAction(isAgent) { implicit user =>
```
avelx, line 48: :100: - I think we should start using this approach right away (drop using showAgent method) and use joint method as removal of isAgent flag is on the horizon;

PR #2191, opened 2024-03-14
Misuv 7256 via actions v2: read FS from mongo

avelx:
**Main changes around:** 
** - In MtdItUser added field featureSwitches: List[FeatureSwitch] to keep list of featureSwitches, see relevant changes in FeatureSwitchPredicate and AuthenticatorPredicate;
** - Changes in FeatureSwitching.isEnabled to accommodate scenario when we read FS from mongoDb;
** - Changes around FeatureSwitch type;
** - Added: FeatureSwitchService / FeatureSwitchRepository - to read FS from the mongo; 

* **Known side effects**: when feature-switches.read-from-mongo set to false, no Fs available on http://localhost:9081/report-quarterly/income-and-expenses/view/test-only/allfeatureSwitches;
* not possible to read FS from mongo before authorisation is completed, see usage of isEnabledFromConfig in few edge case scenarios;

### DIFF
--- a/app/config/featureswitch/FeatureSwitching.scala
+++ b/app/config/featureswitch/FeatureSwitching.scala
@@ -28,7 +28,7 @@ trait FeatureSwitching {
   val FEATURE_SWITCH_OFF = "false"
 
   def isEnabledFromConfig(featureSwitch: FeatureSwitchName): Boolean = {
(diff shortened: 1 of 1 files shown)


simondrugan16:
Will the list of active feature switches updated/retrieved only at the beginning of the journey (since we are passing round the implicit user) or would it update upon new page load?

simondrugan16 (approved):

PR #2194, opened 2024-03-15
MISUV-7285-obscure-test-only-assets

simondrugan16:


app/testOnly/controllers/TestOnlyAssetsController.scala
```diff
@@ -0,0 +1,34 @@
...
+import java.io.File
+import javax.inject.Inject
+import play.api.mvc._
+
+import scala.concurrent.ExecutionContext
+
+class TestOnlyAssetsController @Inject()(cc: ControllerComponents)(implicit ec: ExecutionContext) extends AbstractController(cc) {
+
+  def at(filePath: String): Action[AnyContent] = Action {
+    val file = new File(s"publicTestOnly/$filePath")
```
avelx, line 28: I am just curios if this would work out of fat jar, as this very likely not available in the actual environment

avelx: you can try package it locally using sbt
https://www.scala-sbt.org/sbt-native-packager/gettingstarted.html#setup

sbt universal:packageZipTarball
and then run FE locally / extract content of zip file etc and run it from binaries

simondrugan16: Will give it a go

avelx (approved):

PR #2199, opened 2024-03-19
WIP: MISUV-7056: Create config and code for migration of API#1554 from DES to IF for R12

ayubAz:


it/test/controllers/ChargeSummaryControllerISpec.scala
```diff
@@ -44,7 +44,7 @@ class ChargeSummaryControllerISpec extends ComponentSpecBase {
     paymentsWithCharge("SA Payment on Account 1", NIC4_SCOTLAND, "2019-08-13", -9000.0, lotItem = "000001"),
     paymentsWithCharge("SA Payment on Account 2", NIC4_SCOTLAND, "2019-08-13", -8000.0, lotItem = "000001")
   )
-  val chargeHistories: List[ChargeHistoryModel] = List(ChargeHistoryModel("2019", "1040000124", LocalDate.of(2018, 2, 14), "ITSA- POA 1", 2500, LocalDate.of(2019, 2, 14), "Customer Request"))
+  val chargeHistories: List[ChargeHistoryModel] = List(ChargeHistoryModel("2019", "1040000124", LocalDate.of(2018, 2, 14), "ITSA- POA 1", 2500, LocalDate.of(2019, 2, 14), "Customer Request", None))
```
ian-hmrc, line 47: All these test cases are for where the poaAdjustmentReason is None. Should we be testing the case when it is present?

ian-hmrc: OK. I can see that some of those have been change. That give better confidence as to what it will be.

ian-hmrc: I this this latest change has broken the PR build.

ian-hmrc (approved):
LGTM

PR #2200, opened 2024-03-19
MISUV-7285-v3

simondrugan16:


simondrugan16:
Closing attempt at moving test only JS to different location since better solution was found by Pavel

simondrugan16:
Closing attempt at moving test only JS to different location since better solution was found by Pavel

PR #2203, opened 2024-03-21
Fix address lookup service redirecting back to new journey

danielmaltz1:


SiddharthAtos (approved):

app/connectors/AddressLookupConnector.scala
```diff
@@ -43,7 +43,12 @@ class AddressLookupConnector @Inject()(val appConfig: FrontendAppConfig,
     s"${baseUrl}/api/v2/confirmed?id=$id"
   }
 
-  def continueUrl(isAgent: Boolean, isChange: Boolean): String = if (isAgent) controllers.incomeSources.add.routes.AddBusinessAddressController.agentSubmit(None, isChange = isChange).url else controllers.incomeSources.add.routes.AddBusinessAddressController.submit(None, isChange = isChange).url
+  def continueUrl(isAgent: Boolean, isChange: Boolean): String = (isAgent, isEnabled(IncomeSourcesNewJourney)) match {
```
avelx, line 46: I guess this changes would imply changes in the tests?

danielmaltz1: We stub the API response with the correct url in the tests already

avelx: I see, we are not testing this method in any test; 
we might need to be consistent as in some connectors we are testing methods returning urls, but in some not

danielmaltz1: ATs should cover this, as they check whether the user has been redirected to the correct location

avelx (approved):

PR #2215, opened 2024-03-28
Fix journey checker redirects and session deletion

danielmaltz1:


app/controllers/manageBusinesses/ManageYourBusinessesController.scala
```diff
@@ -63,12 +63,16 @@ class ManageYourBusinessesController @Inject()(val manageYourBusinesses: ManageY
     withIncomeSourcesFS {
       incomeSourceDetailsService.getViewIncomeSourceViewModel(sources) match {
         case Right(viewModel) =>
-          sessionService.deleteSession(Manage).map { _ =>
-            Ok(manageYourBusinesses(
-              sources = viewModel,
-              isAgent = isAgent,
-              backUrl = backUrl
-            ))
+          sessionService.deleteSession(Add).flatMap { _ =>
```
avelx, line 66: I would suggest to create deleteAllSessions within service, as this logic really related to service

avelx: + maybe worth to apply changes on the repo level if this is in scope of this story, ie you can use deleteMany method

PR #2217, opened 2024-04-02
MISUV-7134 - Add repayment supplement credit from excess payment due to charge reduction to credits on View and Change

JordHemsworth:
A credit from refund interest field


app/audit/models/CreditSummaryAuditing.scala
```diff
@@ -32,7 +34,7 @@ object CreditSummaryAuditing {
   }
 
   private def toDescription(creditType: CreditType)
-                           (implicit messages: MessagesApi): String = messages(creditType.key)(Lang("en"))
+                           (implicit messages: MessagesApi): String = messages(s"credit.description.${creditType.key}")(Lang("en"))
```
avelx, line 37: :thinking: we might need to introduce concept of "default language" and apply it to audit only scope

test/views/CreditAndRefundsViewSpec.scala
```diff
@@ -376,6 +387,55 @@ class CreditAndRefundsViewSpec extends TestSupport with FeatureSwitching with Im
...
+          isCutOverCreditsEnabled = true,
+          welshLang = true
+        ) {
+          document.title() shouldBe "Hawlio ad-daliad - Rheoli’ch diweddariadau Treth Incwm - GOV.UK"
+          layoutContent.selectHead("h1").text shouldBe "Hawlio ad-daliad"
+          document.select("ul#credits-list li:nth-child(1)").text() shouldBe
+            "Credyd o £100.00 o flwyddyn dreth gynharach - blwyddyn dreth 2022 i 2023"
+          document.select("ul#credits-list li:nth-child(2)").text() shouldBe
+            "Credyd o £200.00 o ordaliad treth - blwyddyn dreth 2021 i 2022"
+          // TODO: Update welsh language
```
avelx, line 432: just wondering if this TODO is still required?

test/testConstants/PaymentAllocationsTestConstants.scala
```diff
@@ -283,6 +288,7 @@ object PaymentAllocationsTestConstants {
         "to" -> "2019-08-27",
         "chargeType" -> NIC4_WALES,
         "mainType" -> "SA Payment on Account 1",
+//        "mainTransaction" -> "4920",
```
avelx, line 291: commented line, can we either remove it or uncomment

test/services/helpers/CreditHistoryDataHelper.scala
```diff
@@ -112,22 +160,38 @@ trait CreditHistoryDataHelper {
...
         transactionDate = None, `type` = None, totalAmount = None, originalAmount = None, outstandingAmount = None,
+//<<<<<<< HEAD
         clearedAmount = None, chargeType = None, accruedInterest = None, items = Some(Seq(SubItem(Some(fixedDate.plusDays(3)))))),
-      FinancialDetail(taxYear = "2022", mainType = None, transactionId = Some("testTransactionId2"),
+      FinancialDetail(taxYear = "2022", mainType = None, mainTransaction = None, transactionId = Some("testTransactionId8"),
         transactionDate = None, `type` = None, totalAmount = None, originalAmount = None, outstandingAmount = None,
         clearedAmount = None, chargeType = None, accruedInterest = None, items = Some(Seq(SubItem(Some(fixedDate.plusDays(5)))))),
-      FinancialDetail(taxYear = "2022", mainType = Some("SA Balancing Charge Credit"), transactionId = Some("testTransactionId3"),
+      FinancialDetail(taxYear = "2022", mainType = Some("SA Balancing Charge Credit"), mainTransaction = Some("4905"), transactionId = Some(documentDetailsWhichIsBCCredit.transactionId),
+//=======
```
avelx, line 188: its seems there are some issues after merging changes from the main, if this merge conflict can be resolved

app/views/CreditAndRefunds.scala.html
```diff
@@ -41,29 +42,29 @@
     govukInsetText : GovukInsetText
 )
 @(viewModel: CreditAndRefundViewModel, balance: Option[BalanceDetails], creditAndRefundType: Option[UnallocatedCreditType],
-    isAgent: Boolean = false, backUrl: String, isMFACreditsAndDebitsEnabled: Boolean = false, isCutOverCreditsEnabled: Boolean = false)(implicit request: Request[_], user: MtdItUser[_], messages: Messages)
+isAgent: Boolean = false, backUrl: String, isMFACreditsAndDebitsEnabled: Boolean = false, isCutOverCreditsEnabled: Boolean = false)(implicit request: Request[_], user: MtdItUser[_], messages: Messages)
```
avelx, line 45: I think we would need a story to refactor next under viewModel definition
`isAgent: Boolean = false, backUrl: String, isMFACreditsAndDebitsEnabled: Boolean = false, isCutOverCreditsEnabled: Boolean = false`

app/utils/CreditAndRefundUtils.scala
```diff
@@ -36,6 +35,7 @@ object CreditAndRefundUtils {
       def validCredit(detail: FinancialDetail): Boolean = detail.getCreditType match{
         case Some(MfaCreditType) => isMFACreditsAndDebitsEnabled
         case Some(CutOverCreditType) => isCutOverCreditEnabled
+        case Some(RepaymentInterest) => true
```
avelx, line 38: case Some(RepaymentInterest) | Some(BalancingChargeCreditType) => true

app/models/repaymentHistory/PaymentHistoryEntry.scala
```diff
@@ -33,8 +34,8 @@ case class PaymentHistoryEntry(date: LocalDate,
     TaxYear(endYear-1, endYear)
   }
 
-  private val creditDescriptions = Seq("paymentHistory.mfaCredit",
-    "paymentHistory.balancingChargeCredit", "paymentHistory.paymentFromEarlierYear")
-
-  def isCredit: Boolean = creditDescriptions.contains(description)
+  def isCredit: Boolean = creditType match {
```
avelx, line 37:
```
 private val creditTypes: Seq[CreditType] = Seq(BalancingChargeCreditType,  CutOverCreditType,  MfaCreditType, RepaymentInterest)

  def isCredit: Boolean = creditTypes.contains(creditType)
```

app/models/financialDetails/PaymentsResponse.scala
```diff
@@ -45,7 +45,11 @@ case class Payment(reference: Option[String],
                    dueDate: Option[LocalDate],
                    documentDate: LocalDate,
                    transactionId: Option[String],
-                   mainType: Option[String] = None) {
+                   mainType: Option[String] = None,
+                   mainTransaction: Option[String] = None) {
```
avelx, line 49: we might need to introduce a new type here:
class TransactionId(val id: String) extends AnyVal
should be ok to move it into tech debt story as this is really not in scope of this story

app/models/financialDetails/CreditType.scala
```diff
@@ -0,0 +1,73 @@
...
+  // values come from EPID #1138
+  private val cutOver = "6110"
+  private val balancingCharge = "4905"
+  private val repaymentInterest = "6020"
+  private val mfaCredit = Range.inclusive(4004, 4025)
+    .filterNot(_ == 4010).filterNot(_ == 4020).map(_.toString)
+    .toList
+  private val payment = List("4920", "4930")
+
+  def apply(mainTransaction: String): Option[CreditType] = {
```
avelx, line 58: its looks like this is more "unapply" semantic rather than "appy"

app/models/creditsandrefunds/CreditAndRefundViewModel.scala
```diff
@@ -16,18 +16,35 @@
...
+case class CreditViewRow(amount: BigDecimal, creditType: CreditType, taxYear: String) extends CreditRow
+case class PaymentCreditRow(amount: BigDecimal, creditType: CreditType, date: LocalDate) extends CreditRow
 
-  private val balancingChargeCredit = "BCC"
-  private val mfaCredit = "MFA"
-  private val cutOverCredit = "CutOver"
-  private val payment = "Payment"
+case class CreditAndRefundViewModel(creditCharges: List[(DocumentDetailWithDueDate, FinancialDetail)]) {
 
+  def creditViewRows()(implicit messages: Messages): List[CreditRow] = {
```
avelx, line 33: generally we avoiding usage of tuple directly as its hard to read this code:
Something like this would improve code readability:

```
val sortedCreditRows: Seq[Option[CreditRow]] = sortCreditsByYear.map(cc =>
    {
      val (documentDetails, financialDetail) = cc
      val maybeCreditRow = for {
        creditType <- financialDetail.getCreditType
        amount <- documentDetails.documentDetail.paymentOrChargeCredit
      } yield {
        creditType match {
          case PaymentType => PaymentCreditRow(amount = amount, creditType = creditType, date =documentDetails.dueDate.get)
          case _ =>  CreditViewRow(amount = amount, creditType = creditType, taxYear = financialDetail.taxYear)
        }
      }
```

app/models/creditsandrefunds/CreditAndRefundViewModel.scala
```diff
@@ -16,18 +16,35 @@
...
 
+import java.time.LocalDate
 
-case class CreditAndRefundViewModel(creditCharges: List[(DocumentDetailWithDueDate, FinancialDetail)]) {
-
+sealed trait CreditRow {
+  val amount: BigDecimal
+  val creditType: CreditType
+}
+case class CreditViewRow(amount: BigDecimal, creditType: CreditType, taxYear: String) extends CreditRow
```
avelx, line 28: ideally if we can use TaxYear type instead of String type, but should be ok this time as there already lots of changes in scope of this story / feel free to raise Tech Debt

app/models/financialDetails/CreditType.scala
owen-gilfellon: I'd disagree - it operates more like a constructor? It takes the mainTransaction as an argument and returns a subtype of CreditType?

app/models/creditsandrefunds/CreditAndRefundViewModel.scala
JordHemsworth: Shall I raise the ticket for updating TaxYear type across the repo or just for these files?

avelx: I think it would be easy to approach this case by case, but yes, you can move this change into tech debt if you like, scope TBC

app/views/CreditAndRefunds.scala.html
JordHemsworth: Could you expand please

app/models/financialDetails/PaymentsResponse.scala
JordHemsworth: What is the benefit of this?

app/audit/models/CreditSummaryAuditing.scala
JordHemsworth: Lang was already there from messages API, could raise in deep dive

app/views/CreditAndRefunds.scala.html
avelx: I am not sure about agent and backUrl parameters, but 
isMFACreditsAndDebitsEnabled and isCutOverCreditsEnabled can be moved into viewModel, 
so list of parameters we are passing to the view will become "lighter", but I think Ian was thinking to split such views into view, on per each FS (?)

app/models/financialDetails/PaymentsResponse.scala
avelx: * this would improve code readability and less chances to make a mistake, 
there are many other reason to do this, 
* we already applying something similar ie: IncomeSourceId and Nino, TaxYear, these basic domain primitives
in general we should avoid using such types as String and Int, there is no meaning in these when you read code, variable name is not something we should rely on as this is not providing type safety

app/models/financialDetails/CreditType.scala
avelx: apply is for constructors returning plain type; here is for unapply:
```
The return type of an unapply should be chosen as follows:

If it is just a test, return a Boolean. For instance case even().
If it returns a single sub-value of type T, return an Option[T].
If you want to return several sub-values T1,...,Tn, group them in an optional tuple Option[(T1,...,Tn)].
```

avelx: https://docs.scala-lang.org/scala3/reference/other-new-features/creator-applications.html
apply is a shortcut to drop usage of new, nothing else

owen-gilfellon: The link is just saying that the automatically generated apply methods on case classes have been extended to concrete classes - it's not saying apply cannot be used in this context.

I think apply being used to implement a Factory pattern is valid, concise and readable 🙂

https://www.oreilly.com/library/view/scala-cookbook/9781449340292/ch06s10.html

avelx: method logic clearly contains validation logic, I really dont think this is something to do with how apply is intended to be used;

owen-gilfellon: Shall we just rename this method, then? Something like CreditType.parse(mainTransaction) or CreditType.fromCode(mainTransaction)?

avelx: * you can keep it as is, there is not harm in this, its just not advisable practice in my opinion;
* renaming is also a good option, as this seem to be like type conversion not sure about validation;
* also I was thinking next approach:

mainTransactionString => Transaction
def apply(mainTransaction: Transaction): CreditType = ???

owen-gilfellon: I think renaming is the best option at the moment. It would be good to look into using specific types and validation for a lot of fields throughout (using validators when parsing objects from the JSON would be the usual way), but removing the Option[CreditType] return type would require some thought - there are valid mainTransaction codes that we don't currently support on these pages.

vinay1yv (approved):

PR #2225, opened 2024-04-08
MISUV-7408-session-storage-test-only-page

simondrugan16:
Created a test only page that can be accessed to check whether the new income-tax-session-data service is working correctly


app/testOnly/controllers/OptOutTestDataController.scala
```diff
@@ -45,7 +43,8 @@ class OptOutTestDataController @Inject()(
                                           implicit val dateService: DateServiceInterface,
                                           implicit val mcc: MessagesControllerComponents,
                                           val itvcErrorHandler: ItvcErrorHandler,
-                                          implicit val itvcErrorHandlerAgent: AgentItvcErrorHandler
+                                          implicit val itvcErrorHandlerAgent: AgentItvcErrorHandler,
```
avelx, line 46: I don't think this file in anyway related to this change, if we can revert change in this file

app/testOnly/controllers/SessionStorageServiceController.scala
```diff
@@ -0,0 +1,91 @@
...
+        Future.successful(handleError(isAgent))
+    }
+  }
+
+  def post(isAgent: Boolean, sessionId: SessionId)(implicit hc: HeaderCarrier, ec: ExecutionContext, user: MtdItUser[_])
+  : Future[Either[Throwable, String]] =
+    sessionDataService.postSessionData(isAgent, sessionId)
+
+  private def handleError(isAgent: Boolean)(implicit request: Request[_]): Result = {
+    val errorHandler = if (isAgent) itvcErrorHandlerAgent else itvcErrorHandler
```
avelx, line 87: why not to show that there is no data ? we don't have to crash execution here

app/testOnly/controllers/SessionStorageServiceController.scala
```diff
@@ -0,0 +1,91 @@
...
+      handleShow(isAgent = false)
+  }
+
+  def showAgent: Action[AnyContent] = auth.authenticatedAction(isAgent = true) {
+    implicit mtdItUser =>
+      handleShow(isAgent = true)
+  }
+
+  private def handleShow(isAgent: Boolean)(implicit hc: HeaderCarrier, ec: ExecutionContext, user: MtdItUser[_]): Future[Result] = {
+    hc.sessionId match {
```
avelx, line 52: I would suspect its fine to read sessionID from query parameters, there is not strict need to be attached to headerCarrier sessionId

app/testOnly/controllers/OptOutTestDataController.scala
simondrugan16: the pr bot had an issue with the global execution context used here, so i thought it was slightly related

app/testOnly/controllers/SessionStorageServiceController.scala
simondrugan16: query params when accessing the page? if so you would need to know the session id when accessing the page, or create our own session id...

app/testOnly/controllers/SessionStorageServiceController.scala
simondrugan16: for which scenario do you mean?

app/testOnly/controllers/SessionStorageServiceController.scala
avelx: I mean url query parameter can be used here, there is not strict need to rely on existing session for this POC

app/testOnly/controllers/SessionStorageServiceController.scala
avelx: as I understand current logic: you crash execution if there is no data ? so instead you can just show text on the page that there is no data

app/testOnly/controllers/SessionStorageServiceController.scala
simondrugan16: then this would not be the actual sessionId though. it would be just an id that we call session id. i can do this though if you like

avelx: my understanding of sessionId concept in this context that this is customised session across services to pass data from service A (data producer) to service B(data consumer), essentially internal sessionId for service A and/or service B has no direction relation to this concept, but again, this something we can raise as a question

app/testOnly/controllers/SessionStorageServiceController.scala
simondrugan16: in the current logic we post the data in the show method, so if there is no data then this is a big problem. Maybe I could show text on the page explaining this? E.g: No data in database for this session. Something has gone wrong.

app/testOnly/controllers/SessionStorageServiceController.scala
simondrugan16: Who should we raise this with? We could call it sessionDataId perhaps

avelx: this is POC, we can change things later when actual integration work would start

app/testOnly/controllers/SessionStorageServiceController.scala
avelx: this is not decided yet how client would handle missing data, so I guess not critical at the moment

avelx (approved):

PR #2226, opened 2024-04-09
MISUV-7346 ITSAViewer_OptOut: Home page - for no Opt out & Opt out - Feature Switch Added

tharunprasath:


conf/messages
```diff
@@ -152,6 +152,9 @@ home.overdue.message.dunningLock.false                          = You have overd
 home.updates.overdue.payments                                   = {0} OVERDUE PAYMENTS
 home.updates.overdue.updates                                    = {0} OVERDUE UPDATES
 
+## Opt-Out ##
```
ian-hmrc, line 155: Will having the opt-out messages separate from the main messages cause a problem when we eventually remove the feature switch? Might we end up leaving some old messages lying around? Would it be better to put these next to the messages they're replacing?

test/views/HomePageViewSpec.scala
```diff
@@ -102,19 +102,20 @@ class HomePageViewSpec extends TestSupport with FeatureSwitching {
   val overdueMessage = s"! Warning ${messages("home.overdue.message.dunningLock.false")}"
   val overdueMessageForDunningLocks = s"! Warning ${messages("home.overdue.message.dunningLock.true")}"
   val currentDate = dateService.getCurrentDate
-  private val viewModelFuture: NextUpdatesTileViewModel = NextUpdatesTileViewModel(Seq(LocalDate.of(2100, 1, 1)), currentDate)
-  private val viewModelOneOverdue: NextUpdatesTileViewModel = NextUpdatesTileViewModel(Seq(LocalDate.of(2018, 1, 1)), currentDate)
+  private val viewModelFuture: NextUpdatesTileViewModel = NextUpdatesTileViewModel(Seq(LocalDate.of(2100, 1, 1)), currentDate, false)
```
ian-hmrc, line 105: Should we be adding unit tests for when the feature switch is true?

tharunprasath: Added viewModelOptOut to test when the feature switch is true.

conf/messages
tharunprasath: Done

test/views/HomePageViewSpec.scala
ian-hmrc: Cool :)

ian-hmrc (approved):
LGTM :)

PR #2243, opened 2024-04-19
MISUV-7475-claim-to-adjust-entry-point

simondrugan16:
Adding helper functions for checking POA tax year to eliminate duplication of effort


app/services/ClaimToAdjustService.scala
```diff
@@ -0,0 +1,62 @@
...
+          getPoAPayments(model.documentDetails)
+        })
+    }
+    a.map(_.head)
+  }
+
+  private def getPoAPayments(documentDetails: List[DocumentDetail]): Either[Throwable, Option[TaxYear]] = {
+
+    val poa1: Option[TaxYear] = documentDetails.filter(_.documentDescription.exists(_.equals("ITSA- POA 1")))
+      .sortBy(_.taxYear).reverse.headOption.map(_.toTaxYear)
```
avelx, line 48: I am not sure if we need to do any sorting at all, our expectation is that there will either only one record 
with documentDescrition equals to "ITSA- POA 1" or None, similar for "ITSA - POA 2"

app/models/financialDetails/DocumentDetail.scala
```diff
@@ -202,6 +204,10 @@ case class DocumentDetail(taxYear: Int,
     }
   }
 
+  def toTaxYear: TaxYear = {
```
avelx, line 207: not sure if we need this method, seem to be duplicate of makeTaxYearWithEndYear

app/services/ClaimToAdjustService.scala
```diff
@@ -0,0 +1,62 @@
...
+                                    (implicit ec: ExecutionContext) {
+
+  def getPoATaxYear(implicit hc: HeaderCarrier, user: MtdItUser[_]): Future[Either[Throwable, Option[TaxYear]]] = {
+    val a = financialDetailsService.getAllFinancialDetails map {
+      item =>
+        item.collect({ case (_, model: FinancialDetailsModel) =>
+          getPoAPayments(model.documentDetails)
+        })
+    }
+    a.map(_.head)
```
avelx, line 42: Please avoid using .head unless you are 100% sure there data in the relevant collection etc

app/services/ClaimToAdjustService.scala
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  private def getPoAPayments(documentDetails: List[DocumentDetail]): Either[Throwable, Option[TaxYear]] = {
```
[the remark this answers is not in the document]
simondrugan16: I tried but I think it's impossible because the for compresion returns Option[X] if either poa1 is empty or poa2 is empty (invalid scenario), _AND_ if they are both empty (valid scenario). Both poa 1 and poa 2 need to be wrapped in their own options witch I couldnt solve with a for comprehension

app/services/ClaimToAdjustService.scala
simondrugan16: I clarified this in the Lunar chat on Friday. When a tax year is crystallised new poas are generated but the old ones will still show up in 1553 AFAIK. Therefore we need grab the PoAs for the most recent TaxYear and not the old ones

app/services/ClaimToAdjustService.scala
simondrugan16: Good point i will write additional test for this

app/services/ClaimToAdjustService.scala
```diff
@@ -0,0 +1,62 @@
...
+import javax.inject.{Inject, Singleton}
+import scala.concurrent.{ExecutionContext, Future}
+
+@Singleton
+class ClaimToAdjustService @Inject()(val financialDetailsConnector: FinancialDetailsConnector,
+                                     val financialDetailsService: FinancialDetailsService)
+                                    (implicit ec: ExecutionContext) {
+
+  def getPoATaxYear(implicit hc: HeaderCarrier, user: MtdItUser[_]): Future[Either[Throwable, Option[TaxYear]]] = {
+    val a = financialDetailsService.getAllFinancialDetails map {
```
avelx, line 36: this might be not a good idea to create dependency on other service, like financialDetailsService;
if we can rely on the financialDetailsConnector;

app/models/financialDetails/DocumentDetail.scala
simondrugan16: let me check if makeTaxYearWithEndYear works with map

simondrugan16: It doesn't seem to work with map in the same way (.map(_.taxYear. makeTaxYearWithEndYear). Doing it like this makes it quite concise too I think

app/services/ClaimToAdjustService.scala
simondrugan16: getAllFinancialDetails does pretty much everything we want though. I could duplicate the code in this service though..?

avelx: you can duplicate this code in this service, but if you would look at the implementation, we should not do so many API calls;

app/models/financialDetails/DocumentDetail.scala
avelx: I would say if we can keep all functions returning same time under relevant singleton ? + if we can avoid creating new function by using existing one, why we need to create something new?

app/services/ClaimToAdjustService.scala
avelx: I suspect that after you would drop usage of FinancialDetailsService, but use connector instead, this function would significantly change

avelx (approved):

PR #2244, opened 2024-04-22
Misuv 7493 exceptions thrown in view

james-king-hmrc:
- code refactor to target change: tech-debt task to remove code throwing exceptions from the view template file


app/controllers/ChargeSummaryController.scala
```diff
@@ -40,6 +40,20 @@ import views.html.errorPages.CustomNotFoundError
...
 import javax.inject.Inject
 import scala.concurrent.{ExecutionContext, Future}
+import scala.language.implicitConversions
+import controllers.ChargeSummaryController._
+
+object ChargeSummaryController {
+  case class ErrorCode(message: String, code: Int = 0, showInternalServerError: Boolean = true) extends RuntimeException(message)
+  type ItvcResponse[T]  = Either[ErrorCode, T]
+  type ItvcResult  = Future[Result]
+  implicit def toEither[T](t: T): ItvcResponse[T] = Right(t)
```
avelx, line 50: I would suspect this is something we can cover by using cats library which is in scope of this project

james-king-hmrc: indeed, already investigated using Cats and we can use EitherT transformer for this type of conversion but didn't want to spend to much time perfecting it before getting some initial feedback, will do

james-king-hmrc: tried to use Cats, it didn't compile for some reason happy to pair with someone on this, but i don't favour invisible implicits that aren't on the screen for me to see, I prefer implicit type extensions that get called explicitly and are visible on the screen to give a hint to what's going on in the code, happy to discuss :)

app/utils/ImplicitsWithFunUtils.scala
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package utils
+
+import play.api.mvc.Result
+
+import scala.concurrent.Future
+import scala.language.implicitConversions
+
+object ImplicitsWithFunUtils {
```
avelx, line 24: this should come via proposal stage as this fundamentals different from the way we do in the rest of the code base; the short answer, not in the scope of this PR pls

james-king-hmrc: okay no problem, will move code to controller

ian-hmrc:
I'm struggling to review this PR with so much change and large commits. I like the final result, but I can't be sure that it is a refactoring that provides equivalent behaviour to the original code. Would it be possible to break this down into smaller, cleaner commits to show the connection from the original code to the final result, and are easy to review?

PR #2246, opened 2024-04-23
MISUV-7475-fe-work-claim-to-adjust-entry-point

simondrugan16:
[MISUV-7388](https://jira.tools.tax.service.gov.uk/browse/MISUV-7388) working on the visible aspects of this ticket, aka including the links to the new page in the correct places

### DIFF
--- a/app/controllers/WhatYouOweController.scala
+++ b/app/controllers/WhatYouOweController.scala
@@ -92,10 +92,13 @@ class WhatYouOweController @Inject()(val whatYouOweService: WhatYouOweService,
     }
   }
 
-  private def claimToAdjustViewModel(nino: Nino)(implicit hc: HeaderCarrier, user: MtdItUser[_]): Future[WYOClaimToAdjustViewModel] = {
+  private def claimToAdjustViewModel(nino: Nino)(implicit hc: HeaderCarrier): Future[WYOClaimToAdjustViewModel] = {
+    if (isEnabled(AdjustPaymentsOnAccount) ){
     claimToAdjustService.getPoaTaxYearForEntryPoint(nino).flatMap {
       case Right(value) => Future.successful(whatYouOweService.getWYOClaimToAdjustViewModel(isEnabled(AdjustPaymentsOnAccount), value))
       case Left(ex: Throwable) => Future.failed(ex)
+    }} else {
+      Future.successful(WYOClaimToAdjustViewModel(isEnabled(AdjustPaymentsOnAccount), None))
     }
   }
 
--- a/app/controllers/WhatYouOweController.scala
+++ b/app/controllers/WhatYouOweController.scala
@@ -24,10 +24,12 @@ import config.featureswitch._
 import controllers.agent.predicates.ClientConfirmedController
 import enums.GatewayPage.WhatYouOwePage
 import forms.utils.SessionKeys.gatewayPage
+import models.core.Nino
+import models.nextPayments.viewmodels.WYOClaimToAdjustViewModel
 import play.api.Logger
 import play.api.i18n.{I18nSupport, Messages}
 import play.api.mvc.{Action, AnyContent, MessagesControllerComponents, Result}
-import services.{DateServiceInterface, WhatYouOweService}
+import services.{ClaimToAdjustService, DateServiceInterface, WhatYouOweService}
 import uk.gov.hmrc.http.HeaderCarrier
 import utils.AuthenticatorPredicate
 import views.html.WhatYouOwe
@@ -36,6 +38,7 @@ import javax.inject.Inject
 import scala.concurrent.{ExecutionContext, Future}
 
 class WhatYouOweController @Inject()(val whatYouOweService: WhatYouOweService,
+                                     val claimToAdjustService: ClaimToAdjustService,
                                      val itvcErrorHandler: ItvcErrorHandler,
                                      implicit val itvcErrorHandlerAgent: AgentItvcErrorHandler,
                                      val authorisedFunctions: FrontendAuthorisedFunctions,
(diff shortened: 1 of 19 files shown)


app/controllers/WhatYouOweController.scala
```diff
@@ -86,6 +92,14 @@ class WhatYouOweController @Inject()(val whatYouOweService: WhatYouOweService,
     }
   }
 
+  private def claimToAdjustViewModel(nino: Nino)(implicit hc: HeaderCarrier, user: MtdItUser[_]): Future[WYOClaimToAdjustViewModel] = {
+    claimToAdjustService.getPoaTaxYearForEntryPoint(nino).flatMap {
+      case Right(value) => Future.successful(whatYouOweService.getWYOClaimToAdjustViewModel(isEnabled(AdjustPaymentsOnAccount), value))
+      case Left(ex: Throwable) => Future.failed(ex)
```
avelx, line 98: always log error as this would easy up investigation

app/models/nextPayments/viewmodels/WYOClaimToAdjustViewModel.scala
```diff
@@ -0,0 +1,32 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.nextPayments.viewmodels
+
+import models.incomeSourceDetails.TaxYear
+
+case class WYOClaimToAdjustViewModel(adjustPaymentsOnAccountFSEnabled: Boolean,
```
avelx, line 21: I dont think we need to pass this FS as a part of this view model, we should not be able to reach this part of evaluation/logic in the first place within a controller

simondrugan16: Should I pass the ViewModel (without the FS) to the View as an Option then?

avelx: please ignore my comment, it should be fine in this case

app/services/ClaimToAdjustService.scala
```diff
@@ -86,4 +89,67 @@ class ClaimToAdjustService @Inject()(val financialDetailsConnector: FinancialDet
     }).map(_.flatten)
   }
 
+  def getPoaTaxYearForEntryPoint(nino: Nino)(implicit hc: HeaderCarrier, user: MtdItUser[_]): Future[Either[Throwable, Option[TaxYear]]] = {
+    checkCrystallisation(nino, getPoaAdjustableTaxYears).flatMap {
```
avelx, line 93: I would suspect that for-comprehension is a good option to use here instead of pattern matching result evertime

app/services/ClaimToAdjustService.scala
```diff
@@ -86,4 +89,67 @@ class ClaimToAdjustService @Inject()(val financialDetailsConnector: FinancialDet
...
+      case Some(taxYear: TaxYear) => financialDetailsConnector.getFinancialDetails(taxYear.endYear, nino.value).map {
+        case financialDetails: FinancialDetailsModel => Right(arePoAPaymentsPresent(financialDetails.documentDetails))
+        case error: FinancialDetailsErrorModel if error.code != NOT_FOUND => Left(new Exception("There was an error whilst fetching financial details data"))
+        case _ => Right(None)
+      }
+    }
+  }
+
+  private def arePoAPaymentsPresent(documentDetails: List[DocumentDetail]): Option[TaxYear] = {
+    documentDetails.filter(_.documentDescription.exists(description => description.equals("ITSA- POA 1") || description.equals("ITSA - POA ")))
```
avelx, line 104: I would assume you mean: "ITSA - POA 2" here ?

avelx: you can use next more concise approach: 

```
  private val poas: List[String] = List("ITSA- POA 1", "ITSA - POA 2")
  
  private def arePoAPaymentsPresent(documentDetails: List[DocumentDetail]): Option[TaxYear] = {
    documentDetails.filter(_.documentDescription.exists(description => poas.contains(description) ) )
      .sortBy(_.taxYear).reverse.headOption.map(doc => makeTaxYearWithEndYear(doc.taxYear))
  }
```

app/services/ClaimToAdjustService.scala
```diff
@@ -86,4 +89,67 @@ class ClaimToAdjustService @Inject()(val financialDetailsConnector: FinancialDet
...
+      )
+    } else {
+      List(
+        TaxYear.makeTaxYearWithEndYear(dateService.getCurrentTaxYearEnd).addYears(-1),
+        TaxYear.makeTaxYearWithEndYear(dateService.getCurrentTaxYearEnd)
+      ).sortBy(_.endYear)
+    }
+  }
+
+  private def checkCrystallisation(nino: Nino, taxYearList: List[TaxYear])(implicit hc: HeaderCarrier): Future[Option[TaxYear]] = {
```
avelx, line 121: if we can use foldLeft or similar to avoid recursive function usage

app/services/ClaimToAdjustService.scala
```diff
@@ -86,4 +89,67 @@ class ClaimToAdjustService @Inject()(val financialDetailsConnector: FinancialDet
...
+    val futureTaxYear = taxYear.endYear >= currentTaxYearEnd
+    if (futureTaxYear) {
+      Future.successful(false)
+    } else {
+      getTYSCrystallisationResult(nino, taxYear.endYear)
+    }
+  }
+
+  private def getTYSCrystallisationResult(nino: Nino, taxYear: Int)(implicit hc: HeaderCarrier): Future[Boolean] = {
+    val taxYearRange = s"${(taxYear - 1).toString.substring(2)}-${taxYear.toString.substring(2)}"
```
avelx, line 144: I guess this taxYear format can be provided by taxYear if we would change it to be type of TaxYear, but we need to check if we have exact format on the level of TaxYear type

test/views/NextUpdatesViewSpec.scala
```diff
@@ -106,5 +108,10 @@ class NextUpdatesViewSpec extends TestSupport {
       table.select(".govuk-table__cell:nth-of-type(1)").text() shouldBe messages("nextUpdates.quarterly")
       table.select(".govuk-table__cell:nth-of-type(2)").text() shouldBe messages(testTradeName)
     }
+
+//    s"have the correct Claim To Adjust PoA Link and Text" in new Setup(obligationsModel) {
```
avelx, line 112: we would need to re-enable this UT at some point

test/views/TaxYearSummaryViewSpec.scala
```diff
@@ -415,6 +421,11 @@ class TaxYearSummaryViewSpec extends ViewSpec with FeatureSwitching {
         layoutContent.getOptionalSelector("p.govuk-body") shouldBe None
       }
 
+//      "display relevant paragraph and link relating to claim to adjust PoA" in new Setup(poaView()) {
```
avelx, line 424: re-enable this UT

app/models/nextPayments/viewmodels/WYOClaimToAdjustViewModel.scala
```diff
@@ -0,0 +1,36 @@
...
+
+  val claimToAdjustTaxYear: Option[TaxYear] = {
+    if (adjustPaymentsOnAccountFSEnabled) {
+      poaTaxYear
+    } else {
+      None
+    }
+  }
+
+  def ctaLink = {
```
avelx, line 32: not sure why we need this method at all

app/services/ClaimToAdjustService.scala
```diff
@@ -86,4 +89,70 @@ class ClaimToAdjustService @Inject()(val financialDetailsConnector: FinancialDet
...
+
+  // Next 2 methods taken from Calculation List Service (with small changes), but I don't like the code duplication. Is there any solution to this?
+  private def isTaxYearNonCrystallised(taxYear: TaxYear, nino: Nino)(implicit hc: HeaderCarrier): Future[Boolean] = {
+
+    val currentTaxYearEnd = dateService.getCurrentTaxYearEnd
+    val futureTaxYear = taxYear.endYear >= currentTaxYearEnd
+    if (futureTaxYear) {
+      Future.successful(true)
+    } else {
+      isTYSCrystallised(nino, taxYear.endYear).map {
```
avelx, line 139: I guess there is no need to mapping, you just inverse result in the .map ?

avelx (approved):

PR #2247, opened 2024-04-24
MISUV-7527-v2

simondrugan16:
hotfix for shared method

### DIFF
--- a/app/services/ClaimToAdjustService.scala
+++ b/app/services/ClaimToAdjustService.scala
@@ -18,6 +18,7 @@ package services
 
 import auth.MtdItUser
 import connectors.FinancialDetailsConnector
+import exceptions.MissingFieldException
 import models.financialDetails.{DocumentDetail, FinancialDetailsErrorModel, FinancialDetailsModel, FinancialDetailsResponseModel}
 import models.incomeSourceDetails.TaxYear
 import models.incomeSourceDetails.TaxYear.makeTaxYearWithEndYear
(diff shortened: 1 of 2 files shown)


app/services/ClaimToAdjustService.scala
```diff
@@ -48,24 +49,25 @@ class ClaimToAdjustService @Inject()(val financialDetailsConnector: FinancialDet
   private def getPoAPayments(documentDetails: List[DocumentDetail]): Either[Throwable, Option[TaxYear]] = {
     {
       for {
-        poa1 <- documentDetails.filter(_.documentDescription.exists(_.equals("ITSA- POA 1")))
-          .sortBy(_.taxYear).reverse.headOption.map(doc => makeTaxYearWithEndYear(doc.taxYear))
-        poa2 <- documentDetails.filter(_.documentDescription.exists(_.equals("ITSA - POA 2")))
-          .sortBy(_.taxYear).reverse.headOption.map(doc => makeTaxYearWithEndYear(doc.taxYear))
+        poa1 <- Option(documentDetails.filter(_.documentDescription.exists(_.equals("ITSA- POA 1")))
```
avelx, line 52: we don't need Option wrapper here;
the whole reason we use for-comprehension is to extract poa1 and poa2

avelx: + I thin filter should be by "ITSA - POA 1", you missed one space after "ITSA"

simondrugan16: API spec says that ITSA- POA 1 is correct
https://confluence.tools.tax.service.gov.uk/pages/viewpage.action?pageId=367723014

simondrugan16: If we do not use the option wrapper here then the method will return Left(Exception) when the user has no PoA data, for example if they have just signed up to MTD. This means we would only be able to use it after the entry page - which is fine so long as people are aware of this

avelx: this is not correct approach, we extract value from Option and wrap it back, why ?; we are in control what we return in each case, so we can always change this part of the logic

PR #2249, opened 2024-04-25
[WIP]: MISUV-7389 SCP023 | Claim to Adjust POA | Build - Display amendable POA page

ayubAz:


app/controllers/AmendPOAController.scala
```diff
@@ -0,0 +1,64 @@
...
+import services.ClaimToAdjustService
+import uk.gov.hmrc.auth.core.AuthorisedFunctions
+import utils.AuthenticatorPredicate
+import views.html.AmendPaymentOnAccount
+
+import javax.inject.{Inject, Singleton}
+import scala.concurrent.ExecutionContext
+
+@Singleton
+class AmendPOAController @Inject()(val authorisedFunctions: AuthorisedFunctions,
```
avelx, line 36: I think something like:  AmendablePOAController or DisplayAmendablePOAController so to add a bit more context based on the naming

app/controllers/AmendPOAController.scala
```diff
@@ -0,0 +1,64 @@
...
+  def show(isAgent: Boolean): Action[AnyContent] =
+    auth.authenticatedAction(isAgent) {
+      implicit user =>
+        claimToAdjustService.getPaymentsOnAccount map {
+          case Right(poa: PaymentOnAccount) =>
+            Ok(view(
+              isAgent = isAgent,
+              taxYearModel = poa.taxYear,
+              documentId = poa.transactionId,
+              poaOneFullAmount = poa.paymentOnAccountOne.toCurrencyString,
```
avelx, line 56: if we can align line 56 / 57 with existing practice to format currency and date's on the level of the view

app/models/paymentOnAccount/PaymentOnAccount.scala
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.paymentOnAccount
+
+import models.incomeSourceDetails.TaxYear
+
+case class PaymentOnAccount(
```
avelx, line 21: assuming we made a recently decision to move most of the 1553 processing to BE; see: https://jira.tools.tax.service.gov.uk/browse/MISUV-3817
worth to create a techDebt story to move this model to BE as well

app/services/ClaimToAdjustService.scala
```diff
@@ -33,6 +34,42 @@ class ClaimToAdjustService @Inject()(val financialDetailsConnector: FinancialDet
...
+        poaTwoDocDetail   <- documentDetails.filter(_.isPoATwo).sortBy(_.taxYear).reverse.headOption
+        poaTwoTaxYear      = makeTaxYearWithEndYear(poaTwoDocDetail.taxYear)
+        poaTwoTotalAmount  = poaTwoDocDetail.originalAmount
+        if poaOneTaxYear  == poaTwoTaxYear
+      } yield {
+        // TODO: .get has to be used here, DocumentDetail.originalAmount is not an optional field according to #API1553 - this needs to be fixed in BE/FE
+        PaymentOnAccount(
+          transactionId,
+          poaOneTaxYear,
+          poaOneTotalAmount.get,
```
avelx, line 54: might be worth to set relevant fields in the model as optional; but yes, we should avoid calling .get

app/services/ClaimToAdjustService.scala
```diff
@@ -33,6 +34,42 @@ class ClaimToAdjustService @Inject()(val financialDetailsConnector: FinancialDet
...
+          poaTwoTotalAmount.get
+        )
+      }
+    } toRight {
+      Logger("application").error(s"[ClaimToAdjustService][getPoAPayments] Unable to find required POA records")
```
avelx, line 59: I think this syntax is not required any more, ie you don't have to add something like "[ClaimToAdjustService][getPoAPayments]" to log message

app/views/AmendPaymentOnAccount.scala.html
```diff
@@ -0,0 +1,161 @@
...
+    govukButton : GovukButton,
+    mainTemplate: unifiedLayout,
+    continueButton: ContinueButton,
+    govukInsetText : GovukInsetText,
+    hmrcPageHeading: HmrcPageHeading,
+    govukSummaryList : GovukSummaryList
+)
+
+@(
+    documentId: String,
```
avelx, line 37: single view model for the view pls

app/models/financialDetails/DocumentDetail.scala
```diff
@@ -45,6 +45,10 @@ case class DocumentDetail(taxYear: Int,
                           documentDueDate: Option[LocalDate] = None
                          ) {
 
+  def isPoAOne: Boolean = documentDescription.exists(_.equals("ITSA- POA 1"))
```
avelx, line 48: this would make sense to move these to methods to ClaimToAdjustService, as this is the only place where these are used

app/services/ClaimToAdjustService.scala
```diff
@@ -89,20 +88,47 @@ class ClaimToAdjustService @Inject()(val financialDetailsConnector: FinancialDet
     }).map(_.flatten)
   }
 
-  def getPoaTaxYearForEntryPoint(nino: Nino)(implicit hc: HeaderCarrier): Future[Either[Throwable, Option[TaxYear]]] = {
+  def getPoaTaxYearForEntryPoint(nino: Nino)(implicit hc: HeaderCarrier): Future[Either[Throwable, Option[PaymentOnAccount]]] = {
```
avelx, line 91: if we can keep original function and have a second function with the name:
getPoaTaxYearForEntryPointAsPaymentOnAccount or similar

app/controllers/AmendablePOAController.scala
```diff
@@ -0,0 +1,85 @@
...
+                  poaTwoTransactionId = poa.poaTwoTransactionId,
+                  poaOneFullAmount = poa.paymentOnAccountOne,
+                  poaTwoFullAmount = poa.paymentOnAccountTwo
+                ))
+              )
+            case Right(None) =>
+              Logger("application").error(s"Failed to create PaymentOnAccount model")
+              Future.successful(showInternalServerError(isAgent))
+            case Left(ex) =>
+              Logger("application").error(s"Exception: ${ex.getMessage} - ${ex.getCause}")
```
avelx, line 69: I guess you can just throw Exception here instead of logging and failing future

avelx (approved):
IT to be implemented

PR #2251, opened 2024-04-26
MISUV-7475-fe-v2

simondrugan16:


app/models/nextPayments/viewmodels/WYOClaimToAdjustViewModel.scala
```diff
@@ -29,8 +30,16 @@ case class WYOClaimToAdjustViewModel(adjustPaymentsOnAccountFSEnabled: Boolean,
...
 
-  def ctaLink = {
-    claimToAdjustTaxYear
+}
+
+object WYOClaimToAdjustViewModel {
+
+  def ctaLink(isAgent: Boolean): String = {
+    if (isAgent) {
+      "/report-quarterly/income-and-expenses/view/agents/adjust-poa/start"
```
SiddharthAtos, line 39: could fix: why not get this url from routes something like this  - `controllers.routes.CreditsSummaryController.showAgentCreditsSummary(calendarYear).url`?

simondrugan16: The page it's redirecting to is not created yet and I didn't want to make dummy controllers/routes to keep the compiler happy but your suggestion would be an improvement for future or if the other stuff is merged

SiddharthAtos: cool 😎

app/controllers/TaxYearSummaryController.scala
```diff
@@ -228,20 +234,29 @@ class TaxYearSummaryController @Inject()(taxYearSummaryView: TaxYearSummary,
...
+
     withTaxYearFinancials(taxYear, isAgent) { charges =>
       withObligationsModel(taxYear, isAgent) { obligationsModel =>
         val codingOutEnabled: Boolean = isEnabled(CodingOut)
         val mtdItId: String = if (isAgent) getClientMtditid else user.mtditid
         val nino: String = if (isAgent) getClientNino else user.nino
-        calculationService.getLiabilityCalculationDetail(mtdItId, nino, taxYear).map { liabilityCalcResponse =>
-          view(liabilityCalcResponse, charges, taxYear, obligationsModel, codingOutEnabled,
+        for {
+          viewModel <- claimToAdjustViewModel(Nino(value = user.nino), taxYear)
```
avelx, line 244: not sure if this is a correct order, there are quite a few cases where we don't need to call claimToAdjustViewModel(Nino(value = user.nino), taxYear) based on liabilityCalcResponse

app/controllers/TaxYearSummaryController.scala
```diff
@@ -290,4 +305,18 @@ class TaxYearSummaryController @Inject()(taxYearSummaryView: TaxYearSummary,
...
   }
+
+  private def claimToAdjustViewModel(nino: Nino, taxYear: Int)(implicit hc: HeaderCarrier): Future[TYSClaimToAdjustViewModel] = {
+    if (isEnabled(AdjustPaymentsOnAccount)) {
+      claimToAdjustService.getPoaTaxYearForEntryPoint(nino).flatMap {
+        case Right(value) => value match {
+          case Some(value) if value.endYear == taxYear => Future.successful(TYSClaimToAdjustViewModel(isEnabled(AdjustPaymentsOnAccount), Option(value)))
+          case _ => Future.successful(TYSClaimToAdjustViewModel(isEnabled(AdjustPaymentsOnAccount), None))
+        }
+        case Left(ex: Throwable) => Future.failed(ex)
```
avelx, line 316: Please log each error , this would simplify Kibana log analysis

app/models/liabilitycalculation/viewmodels/TaxYearSummaryViewModel.scala
```diff
@@ -40,4 +42,27 @@ case class TaxYearSummaryViewModel(calculationSummary: Option[CalculationSummary
...
+    }
+  }
+
+}
+
+object TYSClaimToAdjustViewModel {
+
+  def ctaLink(isAgent: Boolean): String = {
+    if (isAgent) {
+      "/report-quarterly/income-and-expenses/view/agents/adjust-poa/start"
```
avelx, line 62: Feel free to create required controller and routes to resolve this URL in the code

app/models/nextPayments/viewmodels/WYOClaimToAdjustViewModel.scala
avelx: Please create required controller and routes to point to this page etc

app/services/ClaimToAdjustService.scala
```diff
@@ -92,42 +98,37 @@ class ClaimToAdjustService @Inject()(val financialDetailsConnector: FinancialDet
...
+        }
       }
-      case Nil => Future.successful(None)
     }
   }
 
-  // Next 2 methods taken from Calculation List Service (with small changes), but I don't like the code duplication. Is there any solution to this?
-  private def isTaxYearNonCrystallised(taxYear: TaxYear, nino: Nino)(implicit hc: HeaderCarrier): Future[Boolean] = {
-
+  private def isFutureTaxYear(taxYear: TaxYear): Boolean = {
```
avelx, line 112: this method is good candidate to be moved into TaxYear case class;
something like that would work well:

` def isFutureTaxYear(implicit dateService: DateService): Boolean = ???`

app/services/ClaimToAdjustService.scala
```diff
@@ -92,42 +98,37 @@ class ClaimToAdjustService @Inject()(val financialDetailsConnector: FinancialDet
...
       Future.successful(true)
     } else {
-      isTYSCrystallised(nino, taxYear.endYear).map {
-        case true => false
-        case false => true
+      calculationListConnector.getCalculationList(nino, taxYear.formatTaxYearRange).flatMap {
+        case res: CalculationListModel => Future.successful(res.crystallised)
+        case err: CalculationListErrorModel if err.code == 204 => Future.successful(Some(false))
+        case err: CalculationListErrorModel => Future.failed(new InternalServerException(err.message))
+      } map {
```
avelx, line 125: you can change logic in lines 121-124 in the next way, so this code would be easy to read:

```
 calculationListConnector.getCalculationList(nino, taxYear.formatTaxYearRange).flatMap {
        case res: CalculationListModel => Future.successful(res.crystallised.getOrElse(false))
        case err: CalculationListErrorModel if err.code == 204 => Future.successful(false)
        case err: CalculationListErrorModel => Future.failed(new InternalServerException(err.message))
      }
```

app/services/ClaimToAdjustService.scala
```diff
@@ -92,42 +98,37 @@ class ClaimToAdjustService @Inject()(val financialDetailsConnector: FinancialDet
...
+  private def isTaxYearNonCrystallised(taxYear: TaxYear, nino: Nino)(implicit hc: HeaderCarrier): Future[Boolean] = {
+    if (isFutureTaxYear(taxYear)) {
       Future.successful(true)
     } else {
-      isTYSCrystallised(nino, taxYear.endYear).map {
-        case true => false
-        case false => true
+      calculationListConnector.getCalculationList(nino, taxYear.formatTaxYearRange).flatMap {
+        case res: CalculationListModel => Future.successful(res.crystallised)
+        case err: CalculationListErrorModel if err.code == 204 => Future.successful(Some(false))
```
avelx, line 123: Please use standard Http error codes like:
 case err: CalculationListErrorModel if err.code == NO_CONTENT => Future.successful(false)

app/services/ClaimToAdjustService.scala
```diff
@@ -92,42 +98,37 @@ class ClaimToAdjustService @Inject()(val financialDetailsConnector: FinancialDet
   }
 
   private def checkCrystallisation(nino: Nino, taxYearList: List[TaxYear])(implicit hc: HeaderCarrier): Future[Option[TaxYear]] = {
-    taxYearList match {
-      case ::(head, next) => isTaxYearNonCrystallised(head, nino).flatMap {
-        case true => Future.successful(Some(head))
-        case false => checkCrystallisation(nino, next)
+    taxYearList.foldLeft(Future.successful(Option.empty[TaxYear])) { (acc, item) =>
```
avelx, line 101: you can try next code as well:

```
 Future.find {
      taxYearList
        .map(taxYear => isTaxYearNonCrystallised(taxYear, nino)
          .map(crystallised => if (crystallised) Some(taxYear) else None))
    }(p => p.isDefined)
      .flatMap(x => Future.successful(x.flatten))
```

avelx: You can try next Cats specific implementation:

```
private def checkCrystallisation(nino: Nino, taxYearList: List[TaxYear])(implicit hc: HeaderCarrier): Future[Option[TaxYear]] = {
  taxYearList
    .foldLeft(Future.successful(Option.empty[TaxYear])) { (acc, item) =>
        OptionT.liftF(acc)
          .getOrElseF {
            isTaxYearNonCrystallised(item, nino)
              .flatMap(crystallised => if (crystallised) Future.successful(Some(item)) else Future.successful(None))
          }
  }
}
```

app/controllers/TaxYearSummaryController.scala
simondrugan16: I'm not really familiar with liabilityCalcResponse. For what conditions would we not need to check?

avelx: have a look at the current implementation for the next method:
TaxYearSummaryController =>  private def view( ...
in some cases we really do not need to claimToAdjustViewModel

simondrugan16: I didn't want to make claimToAdjustViewModel optional as this moves more logic to the view. So when we don't have anything to show on the page we just put in the viewmodel with nothing in it, like 
`      Future.successful(TYSClaimToAdjustViewModel(isEnabled(AdjustPaymentsOnAccount), None))
`
Maybe this is not the best approach?

avelx (approved):

PR #2256, opened 2024-04-29
remove exceptions from view, validate values in controller

james-king-hmrc:
MISUV-7493: remove exceptions from view, validate values in controller

### DIFF
--- a/app/controllers/ChargeSummaryController.scala
+++ b/app/controllers/ChargeSummaryController.scala
@@ -22,6 +22,7 @@ import auth.{FrontendAuthorisedFunctions, MtdItUser}
 import config.featureswitch._
 import config.{AgentItvcErrorHandler, FrontendAppConfig, ItvcErrorHandler}
 import connectors.FinancialDetailsConnector
+import controllers.ChargeSummaryController.ErrorCode
 import controllers.agent.predicates.ClientConfirmedController
 import controllers.predicates._
 import enums.GatewayPage.GatewayPage
@@ -40,6 +41,10 @@ import views.html.errorPages.CustomNotFoundError
 
 import javax.inject.Inject
 import scala.concurrent.{ExecutionContext, Future}
+
+object ChargeSummaryController {
+  case class ErrorCode(message: String, code: Int = 0, showInternalServerError: Boolean = true) extends RuntimeException(message)
+}
 
 class ChargeSummaryController @Inject()(val authenticate: AuthenticationPredicate,
                                         val checkSessionTimeout: SessionTimeoutPredicate,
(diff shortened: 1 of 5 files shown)


ian-hmrc (approved):
LGTM

app/controllers/ChargeSummaryController.scala
```diff
@@ -42,6 +43,10 @@ import views.html.errorPages.CustomNotFoundError
 import javax.inject.Inject
 import scala.concurrent.{ExecutionContext, Future}
 
+object ChargeSummaryController {
+  case class ErrorCode(message: String, code: Int = 0, showInternalServerError: Boolean = true) extends RuntimeException(message)
```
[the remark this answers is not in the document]
james-king-hmrc: will change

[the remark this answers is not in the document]
ian-hmrc: OK, cool.

Looking even more closely, do we need the code and showInternalServerError parameters? These appear to be unused.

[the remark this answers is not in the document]
james-king-hmrc: done

ian-hmrc (approved):
LGTM

app/views/ChargeSummary.scala.html
```diff
@@ -312,24 +312,24 @@ <h2 class="govuk-heading-l">@messages("chargeSummary.chargeHistory.heading")</h2
                         }
                         </td>
                         <td class="govuk-table__cell numeric govuk-!-text-align-right">@{
-                            documentDetail.originalAmount.getOrElse(throw MissingFieldException("Original Amount")).toCurrency
+                            documentDetail.originalAmount.get.toCurrency
```
avelx, line 315: I think this is a way to hide problem rather than resolve issue;
If this can be address with the right view model pls

app/views/ChargeSummary.scala.html
```diff
@@ -312,24 +312,24 @@ <h2 class="govuk-heading-l">@messages("chargeSummary.chargeHistory.heading")</h2
...
                         }</td>
                     </tr>
                 }
 
                 @if(latePaymentInterestCharge) {
                     <tr>
-                        <td class="govuk-table__cell govuk-!-width-one-quarter">@{documentDetail.interestEndDate.getOrElse(throw MissingFieldException("Interest End Date")).toLongDateShort}</td>
+                        <td class="govuk-table__cell govuk-!-width-one-quarter">@{documentDetail.interestEndDate.get.toLongDateShort}</td>
```
avelx, line 322: again, if we can avoid usage of .get over Option field

avelx:
@james-king-hmrc  - if you can address comments in this PR

app/views/ChargeSummary.scala.html
james-king-hmrc: agree, a view model is the right solution as it has different constraints to the domain model for example

james-king-hmrc: having said that it may require a wider architecture standards review so that we have a consistent approach in development and a consistent code review process

avelx: I was under strong impression that this in scope of this story, as we already done investigation how to approach this type of stories; but yes, really up to you how this would be delivered

james-king-hmrc:
Would this approach work? https://github.com/hmrc/income-tax-view-change-frontend/pull/2263/files

PR #2260, opened 2024-05-03
MISUV-7587: opt-out-user-request api call

james-king-hmrc:
MISUV-7587: opt-out-user-request api call


app/models/optOut/OptOutMessageResponse.scala
```diff
@@ -18,8 +18,28 @@ package models.optOut
...
 import models.itsaStatus.StatusDetail
+import play.api.libs.json.{Format, Json}
 
 case class YearStatusDetail(taxYear: TaxYear, statusDetail: StatusDetail)
 case class OptOutMessageResponse(oneYearOptOut: Boolean = false, taxYears: Array[TaxYear] = Array()) {
   def oneYearOptOutTaxYear: TaxYear = taxYears(0)
 }
+
+
+case class OptOutUpdateReason(code: Int) {
```
avelx, line 29: good candidate to use value class instead of plain case class

james-king-hmrc: turns out I don't need it, but will do if I introduce it again

test/connectors/OptOutConnectorTest.scala
```diff
@@ -0,0 +1,125 @@
...
+          List(ErrorItem("INVALID_TAXABLE_ENTITY_ID",
+          "Submission has not passed validation. Invalid parameter taxableEntityId."))
+        )
+        val httpResponse = HttpResponse(BAD_REQUEST, Json.toJson(apiFailResponse), Map(CorrelationIdHeader -> Seq("123")))
+
+        setupHttpClientMock[OptOutUpdateRequest](connector.getUrl(taxableEntityId))(apiRequest, httpResponse)
+
+        val result: Future[OptOutUpdateResponse] = connector.requestOptOutForTaxYear(taxYear, taxableEntityId)
+
+        Await.result(result, 10.seconds)
```
avelx, line 111: Please avoid using Await.result anywhere in the code base
https://www.scalatest.org/user_guide/async_testing

test/mocks/MockHttp.scala
```diff
@@ -33,6 +33,7 @@ import scala.concurrent.{ExecutionContext, Future}
 
 trait MockHttp extends UnitSpec with BeforeAndAfterEach {
 
+  //todo: refactor: rename mockHttpGet to mockHttp
```
ian-hmrc, line 36: Was there a reason for not doing this?

app/connectors/OptOutConnector.scala
```diff
@@ -0,0 +1,77 @@
...
+  def getUrl(taxableEntityId: String): String =
+    s"${appConfig.itvcProtectedService}/income-tax/itsa-status/update/$taxableEntityId"
+
+  def requestOptOutForTaxYear(taxYear: TaxYear, taxableEntityId: String)
+                             (implicit headerCarrier: HeaderCarrier): Future[OptOutUpdateResponse] = {
+
+    val body = OptOutUpdateRequest(taxYear = taxYear.toString)
+
+    http.PUT[OptOutUpdateRequest, HttpResponse](
+      getUrl(taxableEntityId), body, Seq[(String, String)]()
```
ian-hmrc, line 50: I found getUrl() a little confusing in the context of a PUT... Maybe rename to just url()?

app/models/optOut/OptOutUpdateRequestModel.scala
```diff
@@ -0,0 +1,40 @@
...
+  private val defaultUpdateReason: Int = 10
+  private val successfulStatusCode: Int = NO_CONTENT
+
+  case class OptOutUpdateRequest(taxYear: String, updateReason: Int = defaultUpdateReason)
+  sealed trait OptOutUpdateResponse {
+    val statusCode: Int
+  }
+  case class OptOutUpdateResponseSuccess(correlationId: String, statusCode: Int = successfulStatusCode) extends OptOutUpdateResponse
+  case class ErrorItem(code: String, reason: String)
+  case class OptOutUpdateResponseFailure(statusCode: Int, failures: List[ErrorItem]) extends OptOutUpdateResponse
```
ian-hmrc, line 33: I'm not really a fan of using polymorphic response objects like this. I think we might be more idiomatic with an Either or Future. However, I know this is a pattern in the app. Maybe one to bring up at Tech Deep Dive sometime?

app/connectors/OptOutConnector.scala
```diff
@@ -0,0 +1,77 @@
...
+    ).map { response =>
+      response.status match {
+        case Status.NO_CONTENT =>
+          response.json.validate[OptOutUpdateResponseSuccess].fold(
```
ian-hmrc, line 54: Hmm... I'm confused by this. In the case of NO_CONTENT, there shouldn't be any payload, so why are we validating it?

app/connectors/OptOutConnector.scala
```diff
@@ -0,0 +1,77 @@
...
+          valid => {
+            valid.copy(
+              statusCode = response.status,
+              correlationId = response.headers(CorrelationIdHeader).headOption.getOrElse(s"Unknown_$CorrelationIdHeader")
+            )
+          }
+        )
+        case _ =>
+          response.json.validate[OptOutUpdateResponseFailure].fold(
+            invalid => {
```
ian-hmrc, line 68: According to the API spec, it looks like the error will contain `code` and `reason`, yet `OptOutUpdateResponseFailure(statusCode: Int, failures: List[ErrorItem])` does not match that. So I assume we're always going to fail validation and always use the invalid block? Is that intended?

james-king-hmrc: I see that failures is an array of (code and reason) items, agree?
Screenshot 2024-05-09 at 09 36 33

app/models/optOut/OptOutUpdateRequestModel.scala
james-king-hmrc: I guess we still need the OptOutUpdateResponseFailure to parse the response but I can/prefer Either[ErrorCode, OptOutUpdateResponseSuccess], happy to change, shall I?

test/mocks/MockHttp.scala
james-king-hmrc: I was planning to do any refactoring needed as a separate PR at the end as it's not specifically related to this ticket, including may be the isCrystalised Boolean ticket

app/models/optOut/OptOutUpdateRequestModel.scala
avelx: We have to map response in the way @james-king-hmrc  implemented this, its a standard approach across FE and BE, so should be fine, Future[Either[_,_]] type is used in the up stream logic close to the core domain

test/mocks/MockHttp.scala
ian-hmrc: Cool

app/connectors/OptOutConnector.scala
ian-hmrc: Ah, OK. I think I mis-read that spec. Looks good with the `ErrorItem(code: String, reason: String)`.

app/models/optOut/OptOutUpdateRequestModel.scala
james-king-hmrc: yes indeed i was gonna keep OptOutUpdateResponseFailure but return Either[ErrorCode, OptOutUpdateResponseSuccess]

ian-hmrc: Yeah. I know this is a standard pattern, but it still feels less than optimal. I'm happy to stick with it for now, but maybe this is one to discuss in a Tech Deep Dive? Feels like we should be moving away from this?

james-king-hmrc: i think returning Either helps in exiting (when in error) the monad bind loop in a more standard way without the extra step of checking if the response was a failure or not

app/connectors/OptOutConnector.scala
ian-hmrc: OK. I can see that has been updated.

app/connectors/OptOutConnector.scala
ian-hmrc: OK. I can see that has been addressed. :)

ian-hmrc (approved):
LGTM

PR #2263, opened 2024-05-07
What do you think of this approach?

james-king-hmrc:
- implicitly convert or log missing field and default


avelx:
@james-king-hmrc  - removing exception implies changes in the relevant view model or/and introduce one with the mapping between REST API response and view model if required, so exception / data validation logic would be pushed down to service or controller level

PR #2268, opened 2024-05-08
MISUV-7528 - Fix backUrl on AddBusinessNameController

danielmaltz1:


avelx:
@danielmaltz1  - do we have IT test to cover this specific scenario ?

avelx (approved):

avelx (approved):

PR #2276, opened 2024-05-09
MISUV-7609

omirostr1:


app/controllers/ChargeSummaryController.scala
```diff
@@ -197,15 +196,16 @@ class ChargeSummaryController @Inject()(val authenticate: AuthenticationPredicat
     val viewSection3 = isEnabled(ChargeHistory) && (isEnabled(CodingOut) && documentDetailWithDueDate.documentDetail.isPayeSelfAssessment)
 
     val values = List(
-      (viewSection1, documentDetailWithDueDate.documentDetail.originalAmount.isDefined, "Original Amount"),
+      (viewSection1, documentDetailWithDueDate.documentDetail.originalAmount, "Original Amount"),
```
avelx, line 199: this change looks a bit strange, as .isDefined would return Boolean, but .originalAmount would return BigDecimal, 
my assumption is that we can just put "true" instead, as this field will be always with the value

omirostr1: I put "true" for a similar case at line 202

app/models/financialDetails/DocumentDetail.scala
```diff
@@ -47,23 +47,21 @@ case class DocumentDetail(taxYear: Int,
                          ) {
 
   def credit: Option[BigDecimal] = originalAmount match {
-    case None => None
     case _ if (paymentLotItem.isDefined && paymentLot.isDefined) => None
-    case Some(_) if (originalAmount.get == 0) => None
-    case Some(_) if (originalAmount.get > 0) => None
-    case Some(credit) => Some(credit * -1)
+    case _ if (originalAmount == 0) => None
```
avelx, line 51: we can merge these two lines like:
`case _ if originalAmount >= 0 => None`

app/models/financialDetails/DocumentDetail.scala
```diff
@@ -47,23 +47,21 @@ case class DocumentDetail(taxYear: Int,
...
+    case _ if (originalAmount > 0) => None
+    case credit => Some(credit * -1)
   }
 
   def paymentOrChargeCredit: Option[BigDecimal] = outstandingAmount match {
-    case None => None
-    case Some(_) if (outstandingAmount.get == 0) => None
-    case Some(_) if (outstandingAmount.get > 0) => None
-    case Some(credit) => Some(credit * -1)
+    case _ if (outstandingAmount == 0) => None
```
avelx, line 57: same here, lines 57 and 58 can be merged

test/models/financialDetails/DocumentDetailSpec.scala
```diff
@@ -96,34 +96,31 @@ class DocumentDetailSpec extends UnitSpec {
...
         "original amount is negative" in {
-          fullDocumentDetailModel.copy(originalAmount = Some(-20)).originalAmountIsNotZeroOrNegative shouldBe false
+          fullDocumentDetailModel.copy(originalAmount = -20).originalAmountIsNotZeroOrNegative shouldBe false
         }
       }
       "return true" when {
-        "original amount is not present" in {
-          fullDocumentDetailModel.copy(originalAmount = None).originalAmountIsNotZeroOrNegative shouldBe true
-        }
+//        "original amount is not present" in {
```
avelx, line 106: I would assume this code is not applicable anymore and we can drop it? or this is something still need to be fixed?

omirostr1: It can be dropped, I commented it out and forgot to delete it entirely from there after the discussion we had

app/services/ClaimToAdjustHelper.scala
```diff
@@ -43,11 +43,11 @@ trait ClaimToAdjustHelper {
 
   private val isUnpaidPoAOne: DocumentDetail => Boolean = documentDetail =>
     documentDetail.documentDescription.contains(POA1) &&
-      !documentDetail.outstandingAmount.contains(BigDecimal(0))
+      !(documentDetail.outstandingAmount == (0))
```
avelx, line 46: we can make this easier to read if we can apply next change:
(documentDetail.outstandingAmount != 0)

app/services/ClaimToAdjustHelper.scala
```diff
@@ -43,11 +43,11 @@ trait ClaimToAdjustHelper {
...
 
   private val isUnpaidPoATwo: DocumentDetail => Boolean = documentDetail =>
     documentDetail.documentDescription.contains(POA2) &&
-      !documentDetail.outstandingAmount.contains(BigDecimal(0))
+      !(documentDetail.outstandingAmount == (0))
```
avelx, line 50: same here:
`(documentDetail.outstandingAmount != 0)`

avelx:
there are quite a few instance of .head left around change we applied, might worth to create yet another tech debt story to address these, unless there is relevant story to remove exception from the views which can cover these

omirostr1:
From what I can see there is a Service Improvement ticket currently going on for exceptions thrown in ChargeSummaryView but that's the only one, so let me know if you want me to create a new tech debt ticket for it

avelx (approved):

avelx (approved):
with comments

PR #2288, opened 2024-05-17
MISUV-7393: SCP023 | Claim to Adjust POA | Build - Check your answers for adjusting POA page

ayubAz:


app/controllers/claimToAdjustPoa/ChangePoaAmount.scala
```diff
@@ -0,0 +1,48 @@
...
+import play.api.mvc.{Action, AnyContent, MessagesControllerComponents}
+import uk.gov.hmrc.auth.core.AuthorisedFunctions
+import utils.AuthenticatorPredicate
+import views.html.claimToAdjustPoa.CheckYourAnswers
+
+import javax.inject.{Inject, Singleton}
+import scala.concurrent.{ExecutionContext, Future}
+
+@Singleton
+class ChangePoaAmount @Inject()(val authorisedFunctions: AuthorisedFunctions,
```
avelx, line 30: **ChangePoaAmountController**: current naming convention for controllers is to include suffix "Controller"

app/controllers/claimToAdjustPoa/ConfirmationController.scala
```diff
@@ -0,0 +1,47 @@
...
+                                           implicit val itvcErrorHandlerAgent: AgentItvcErrorHandler)
+                                          (implicit val appConfig: FrontendAppConfig,
+                                           implicit override val mcc: MessagesControllerComponents,
+                                           val ec: ExecutionContext)
+  extends ClientConfirmedController {
+
+  def show(isAgent: Boolean): Action[AnyContent] =
+    auth.authenticatedAction(isAgent) {
+      implicit user =>
+        Future successful Ok(
```
avelx, line 41: we don't have agreement regarding usage of "infix notion", so this is something we need to raise during technical deep dive; my preference would be to use "dot notion" across our code base

it/test/controllers/agent/claimToAdjustPoa/CheckYourAnswersControllerISpec.scala
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package controllers.agent.claimToAdjustPoa
+
+class CheckYourAnswersControllerISpec extends controllers.claimToAdjustPoa.CheckYourAnswersControllerISpec {
+
```
avelx, line 20: I would think we need a ISpec for agent as well

avelx: ok, I see, I am not sure why we need this file in this case?

avelx (approved):

PR #2289, opened 2024-05-17
MISUV-7646 | Claim To Adjust POA Connector

owen-gilfellon:
- Created models for POA adjustment request and response (round BigDecimal to two decimal places as per schema)
- Errors reduced to a single message for the FE, with intention of logging in full in BE
- Uses HttpClientV2 - HttpClient has been deprecated [(http-verbs)](https://github.com/hmrc/http-verbs?tab=readme-ov-file#usage)
- CorrelationId generated if one is not provided in HeaderCarrier


app/models/claimToAdjustPoa/ClaimToAdjustPoaRequest.scala
```diff
@@ -0,0 +1,44 @@
...
+  nino: String,
+  taxYear: String,
+  amount: BigDecimal,
+  poaAdjustmentReason: SelectYourReason)
+
+object ClaimToAdjustPoaRequest {
+
+  private val currencyWrites: Writes[BigDecimal] = new Writes[BigDecimal] {
+    def writes(o: BigDecimal): JsValue = {
+      JsNumber(o.setScale(2, RoundingMode.HALF_UP))
```
avelx, line 34: I am not sure that we have to apply rounding here, I think BigDecimal value would be already in the required format

owen-gilfellon: Will it be guaranteed? API requires max of two decimal places, BigDecimal (I think) by default uses 34

app/models/core/CorrelationId.scala
```diff
@@ -0,0 +1,37 @@
...
+
+package models.core
+
+import models.core.CorrelationId.correlationId
+import uk.gov.hmrc.http.HeaderCarrier
+
+import java.util.UUID
+import java.util.UUID.randomUUID;
+
+case class CorrelationId(id: UUID = randomUUID()) {
```
avelx, line 25: should be fine to add more more case class to avoid usage of tuples, they are quite hard to read;

app/models/claimToAdjustPoa/ClaimToAdjustPoaRequest.scala
avelx: I think service or controller would pass amount as a string and it will be guaranteed that there is no more then 2 decimal places; so BigDecimal is just an intermediate type which we use to validate that string in a valid money format; but yes, I happy to see use case where this chain of conventions can break

avelx (approved):
approved with comments

PR #2290, opened 2024-05-20
MISUV-TEMP-temp-reduce-minimum-scoverage

simondrugan16:
reduced minimum scoverage % to 85


avelx:
We assume we would raise it back to 90% in the next tidy up PR

PR #2291, opened 2024-05-20
MISUV-7461: checkpoint page where following year is annual

SiddharthAtos:


test/services/optout/OptOutServiceSpec.scala
```diff
@@ -353,4 +353,52 @@ class OptOutServiceSpec extends UnitSpec
       }
     }
   }
+  "OptOutService.optOutCheckPointPageViewModel" when {
+    val CY = TaxYear.forYearEnd(2024)
+    val PY = CY.addYears(-1)
```
james-king-hmrc, line 358: why not use CY.**previousYear** method? optional comment since this is test code

SiddharthAtos: didn't knew that existed. will implement.

tharunprasath (approved):

PR #2294, opened 2024-05-21
MISUV-5195: NavBarPredicate patch

avelx:
* patch NavBarFromNinoPredicate: load FS state from mongo fix;
  * mark isDisabled method as deprecated;

### DIFF
--- a/app/config/featureswitch/FeatureSwitching.scala
+++ b/app/config/featureswitch/FeatureSwitching.scala
@@ -40,10 +40,13 @@ trait FeatureSwitching {
     }
   }
 
-  // TODO replace this function, or all uses of it, with the inverse of isEnabled
-  // It is currently used in NavBarFromNinoPredicate.scala, ForecastIncomeSummaryController.scala, ForecastTaxCalcSummaryController.scala
-  def isDisabled(featureSwitch: FeatureSwitchName): Boolean = {
-    sys.props.get(featureSwitch.name) orElse appConfig.config.getOptional[String](featureSwitch.name) contains FEATURE_SWITCH_OFF
+  // TODO: replace this function, or all uses of it, with the inverse of isEnabled
+  def isDisabled(featureSwitch: FeatureSwitchName, fs: List[FeatureSwitch]): Boolean = {
+    if (appConfig.readFeatureSwitchesFromMongo) {
+     !fs.exists(x => x.name.name == featureSwitch.name && x.isEnabled)
+    } else {
+      sys.props.get(featureSwitch.name) orElse appConfig.config.getOptional[String](featureSwitch.name) contains FEATURE_SWITCH_OFF
+    }
   }
 
   def enable(featureSwitch: FeatureSwitchName): Unit =
--- a/app/config/featureswitch/FeatureSwitching.scala
+++ b/app/config/featureswitch/FeatureSwitching.scala
@@ -40,7 +40,7 @@ trait FeatureSwitching {
     }
   }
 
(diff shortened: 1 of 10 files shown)


app/config/featureswitch/FeatureSwitching.scala
```diff
@@ -40,10 +40,13 @@ trait FeatureSwitching {
...
-  // TODO replace this function, or all uses of it, with the inverse of isEnabled
-  // It is currently used in NavBarFromNinoPredicate.scala, ForecastIncomeSummaryController.scala, ForecastTaxCalcSummaryController.scala
-  def isDisabled(featureSwitch: FeatureSwitchName): Boolean = {
-    sys.props.get(featureSwitch.name) orElse appConfig.config.getOptional[String](featureSwitch.name) contains FEATURE_SWITCH_OFF
+  // TODO: replace this function, or all uses of it, with the inverse of isEnabled
+  def isDisabled(featureSwitch: FeatureSwitchName, fs: List[FeatureSwitch]): Boolean = {
+    if (appConfig.readFeatureSwitchesFromMongo) {
+     !fs.exists(x => x.name.name == featureSwitch.name && x.isEnabled)
+    } else {
+      sys.props.get(featureSwitch.name) orElse appConfig.config.getOptional[String](featureSwitch.name) contains FEATURE_SWITCH_OFF
```
simondrugan16, line 48: Should be:
sys.props.get(featureSwitch.name) orElse appConfig.config.getOptional[String]("feature-switch.enable-"+ featureSwitch.name) contains FEATURE_SWITCH_OFF

because the names have a prefix in application.conf that isn't in FeatureSwitchName.name

simondrugan16: Need the same thing in the method: isEnabledFromConfig

app/controllers/ForecastIncomeSummaryController.scala
```diff
@@ -67,26 +69,27 @@ class ForecastIncomeSummaryController @Inject()(val forecastIncomeSummaryView: F
...
-              Ok(forecastIncomeSummaryView(model, taxYear, backUrl(taxYear, origin, isAgent), isAgent,
-                user.btaNavPartial))
-            case _ =>
-              onError("No income data could be retrieved. Not found", isAgent, taxYear)
-          }
-        case error: LiabilityCalculationError if error.status == NO_CONTENT =>
-          onError("No income data found.", isAgent, taxYear)
-        case _: LiabilityCalculationError =>
-          onError("No new calc income data error found. Downstream error", isAgent, taxYear)
+    featureSwitchService.getAll.flatMap { fs =>
```
simondrugan16, line 72: Is there a benefit to passing in the FS list (fs: List[FeatureSwitch]) to isDisabled, rather than isDisabled fetching it separately? This would have a bit less complexity

avelx: * marked isDisabled as deprecated, this would discourage it usage anywhere else in the code base;
* controller method need to be rewritten in the way we can use isEnabled instead
Tech debt story created: https://jira.tools.tax.service.gov.uk/browse/MISUV-7712

app/config/featureswitch/FeatureSwitching.scala
avelx: added required tests coverage and fix above;
remove not used FS keys as well

simondrugan16 (approved):

simondrugan16 (approved):

PR #2308, opened 2024-05-28
MISUV-7719-intercept-poa-post

simondrugan16:
added itvc prefix to url, and added support for post request


avelx (approved):

avelx (changes requested):
please fix broken IT tests

avelx (approved):

PR #2314, opened 2024-05-30
MISUV-7490 update connectors

danielmaltz1:


app/connectors/ChargeHistoryConnector.scala
```diff
@@ -0,0 +1,78 @@
...
+import models.chargeHistory.{ChargeHistoryResponseModel, ChargesHistoryErrorModel, ChargesHistoryModel}
+import play.api.Logger
+import play.api.http.Status
+import play.api.http.Status.OK
+import uk.gov.hmrc.http.{HeaderCarrier, HttpClient, HttpResponse}
+
+import javax.inject.Inject
+import scala.concurrent.{ExecutionContext, Future}
+
+class ChargeHistoryConnector @Inject()(val http: HttpClient,
```
avelx, line 29: We already have a connector which returns Future[ChargeHistoryResponseModel], see FinancialDetailsConnector.getChargeHistory, so might worth to move method from FinancialDetailsConnector into this new connector for consistency or we can create a tech debt for this ?

app/connectors/ChargeHistoryConnector.scala
```diff
@@ -0,0 +1,78 @@
...
+              Logger("application").debug(s"Status: ${response.status}, json: ${response.json}")
+              response.json.validate[ChargesHistoryModel].fold(
+                invalid => {
+                  Logger("application").error(s"Json Validation Error: $invalid")
+                  ChargesHistoryErrorModel(Status.INTERNAL_SERVER_ERROR, "Json Validation Error. Parsing ChargeHistory Data Response")
+                },
+                valid => valid
+              )
+            case status =>
+              if (status == 404 || status == 403) {
```
avelx, line 55: Please use HttpStatus code like
` if (status == NOT_FOUND || status == FORBIDDEN) {`

app/connectors/ChargeHistoryConnector.scala
danielmaltz1: Hi Pavel, this is that same code, I've just moved it into a new file for clarity

avelx (approved):

PR #2315, opened 2024-05-30
MISUV-7733-submit-data-to-etmp-on-cya-page-when-user-is-decreasing-poa-amount

simondrugan16:
On CTA POA CYA page. When user is increasing the amount we submit ETMP data here rather than putting them through to the acknowledge page


app/services/claimToAdjustPoa/SubmitPoaHelper.scala
```diff
@@ -0,0 +1,91 @@
...
+    poaSessionService.getMongo(hc, ec).flatMap {
+      case Right(Some(newPoaData: PoAAmendmentData)) =>
+        Future.successful(newPoaData)
+      case _ =>
+        Future.failed(new Exception(s"Failed to retrieve session data"))
+    }
+  }
+
+  protected def handlePoaAndOtherData(poa: PaymentOnAccountViewModel,
+                                      otherData: PoAAmendmentData, isAgent: Boolean, nino: String, ctaCalculationService: ClaimToAdjustPoaCalculationService)
```
avelx, line 47: :seedling: - in general if we can use Nino value class instead of String type, but should be fine to use String this time;

test/views/claimToAdjustPoa/CheckYourAnswersControllerViewSpec.scala
```diff
@@ -49,6 +49,7 @@ class CheckYourAnswersControllerViewSpec extends TestSupport {
       )
   }
 
+  //noinspection ScalaStyle
```
avelx, line 52: Just wondering what was ScalaStyle error here?

app/services/claimToAdjustPoa/SubmitPoaHelper.scala
```diff
@@ -0,0 +1,91 @@
...
+import models.claimToAdjustPoa.{PaymentOnAccountViewModel, PoAAmendmentData}
+import models.core.Nino
+import play.api.Logger
+import play.api.mvc.Result
+import services.{ClaimToAdjustPoaCalculationService, ClaimToAdjustService, PaymentOnAccountSessionService}
+import uk.gov.hmrc.http.HeaderCarrier
+
+import scala.concurrent.{ExecutionContext, Future}
+
+trait SubmitPoaHelper extends ClientConfirmedController with FeatureSwitching {
```
avelx, line 34: :wrench: - RecalculatePoaHelper or similar, I am not sure if "Submit" is something related to this domain, we need to be domain specific so to easy read this code in the future;

avelx (approved):

test/views/claimToAdjustPoa/CheckYourAnswersControllerViewSpec.scala
simondrugan16: It was complaining that the method was too long

app/services/claimToAdjustPoa/SubmitPoaHelper.scala
simondrugan16: I was just refactoring the code but happy to add this

avelx (approved):

PR #2317, opened 2024-05-30
Hot fix nav bar bug staging

avelx:
Change predicate order while forming Play action; NavBar predicate requires FS data to be loaded first by FeatureSwitchPredicate;


simondrugan16:
We use the old way of retrieving nav bar in DeductionsSummaryController line 83, ForecastIncomeSummaryController and ForecastTaxCalcSummaryController. Do these need the FSPredicate in them too?

avelx:
> We use the old way of retrieving nav bar in DeductionsSummaryController line 83, ForecastIncomeSummaryController and ForecastTaxCalcSummaryController. Do these need the FSPredicate in them too?

NavBarFromNinoPredicate and NavBarPredicate not compatible by the type the return, so they way how NavBarFromNinoPredicate is changes is fine we can change this approach later if needed

simondrugan16 (approved):

PR #2332, opened 2024-06-07
MISUV-7536-choose-to-opt-out-from-multiple-eligible-years

james-king-hmrc:
choose-to-opt-out-from-multiple-eligible-years


app/services/optout/OptOutService.scala
```diff
@@ -113,6 +116,18 @@ class OptOutService @Inject()(itsaStatusUpdateConnector: ITSAStatusUpdateConnect
   def makeOptOutUpdateRequest(optOutProposition: OptOutProposition, intent: OptOutTaxYear)(implicit user: MtdItUser[_],
                                                                                            hc: HeaderCarrier, ec: ExecutionContext): Future[OptOutUpdateResponse] = {
 
+    def combineByReturningAnyFailureFirstOrAnySuccess(responses: Seq[Future[OptOutUpdateResponse]]): Future[OptOutUpdateResponse] = {
```
avelx, line 119: again, this is bad practice to deal with Future input parameter, if this can be addressed

PR #2336, opened 2024-06-11
MISUV-7394 Dymanic data for poa amount

danielmaltz1:


app/connectors/FinancialDetailsConnector.scala
```diff
@@ -91,15 +92,17 @@ class FinancialDetailsConnector @Inject()(val http: HttpClient,
...
+                         (implicit headerCarrier: HeaderCarrier, mtdItUser: MtdItUser[_]): Future[FinancialDetailsResponseModel] = {
 
     val dateFrom: String = (taxYear - 1).toString + "-04-06"
     val dateTo: String = taxYear.toString + "-04-05"
 
     val url = getChargesUrl(nino, dateFrom, dateTo)
     Logger("application").debug(s"GET $url")
 
-    http.GET[HttpResponse](url)(httpReads, headerCarrier, implicitly) map { response =>
+    val hc = checkAndAddTestHeader(mtdItUser.path, headerCarrier, appConfig.poaAdjustmentOverrides(), "afterPoaAmountAdjusted")
```
avelx, line 103: * we might need to raise a tech debt to move all this kind of transformation to a different level, like => HeaderExtractor etc;

danielmaltz1: I think we only want this to be called in these 2 specific scenarios (at least, at the moment), so I'm not sure if moving it to the HeaderExtractor is the best place

avelx: I am going to create a tech debt story for you as I think connectors should not include this type of logic

avelx (approved):
approved with comments

PR #2340, opened 2024-06-11
MISUV-7598: SCP023 | Claim to Adjust POA | Update - Display adjustable POA page

ayubAz:


app/models/chargeHistory/ChargeHistoryModel.scala
```diff
@@ -37,6 +37,8 @@ case class ChargeHistoryModel(taxYear: String,
       Logger("application").error(s"Missing or non-matching history reason: $error found")
       "unrecognisedReason"
   }
+
+  val isPoA: Boolean = documentDescription == "POA1" || documentDescription == "POA2"
```
avelx, line 41: you can changed it to the next code + POAx vals can be moved to constants

```
  val isPoA: Boolean = Seq(POA1, POA2).contains(documentDescription)
```

app/services/ClaimToAdjustHelper.scala
```diff
@@ -134,4 +141,39 @@ trait ClaimToAdjustHelper {
...
+    if poasAreBeforeDeadline
+  } yield {
+    AmendablePoaViewModel(
+      poaOneTransactionId            = poaOneDocDetail.transactionId,
+      poaTwoTransactionId            = poaTwoDocDetail.transactionId,
+      taxYear                        = makeTaxYearWithEndYear(latestDocumentDetail.taxYear),
+      paymentOnAccountOne            = poaOneDocDetail.originalAmount,
+      paymentOnAccountTwo            = poaTwoDocDetail.originalAmount,
+      poARelevantAmountOne           = poaOneDocDetail.poaRelevantAmount.getOrElse(throw MissingFieldException("DocumentDetail.poaRelevantAmount")),
+      poARelevantAmountTwo           = poaTwoDocDetail.poaRelevantAmount.getOrElse(throw MissingFieldException("DocumentDetail.poaRelevantAmount")),
```
avelx, line 161: you can just return error in the Left, not sure why we throwing an error

app/services/ClaimToAdjustService.scala
```diff
@@ -120,6 +120,21 @@ class ClaimToAdjustService @Inject()(val financialDetailsConnector: FinancialDet
       }
     }
   }
+
+  def getAdjustPaymentOnAccountViewModel(nino: Nino)
```
avelx, line 124: if you can rename this method to be aligned with the viewModel name, ie something like
getAmendablePoaViewModel or change viewModel name as it a bit confusing to read

app/services/ClaimToAdjustHelper.scala
```diff
@@ -134,4 +141,39 @@ trait ClaimToAdjustHelper {
       .sortBy(_.taxYear).reverse.headOption.map(doc => makeTaxYearWithEndYear(doc.taxYear))
   }
 
+  def getAmendablePoaViewModel(documentDetails: List[DocumentDetail],
```
avelx, line 144: its the same method name as in actual ClaimToAdjustService but with a different set of parameters, maybe we can change method name to createAmendablePoaViewModel ? or similar to have to differentiate between these two;

app/views/claimToAdjustPoa/AmendablePaymentOnAccount.scala.html
```diff
@@ -87,67 +93,131 @@
...
-                content = Text(getMessage("table-heading-2"))
-            )),
-            actions = Some(Actions(
-                items = Seq(
-                    ActionItem(
-                        href = getChargeSummaryUrl(paymentOnAccount.poaTwoTransactionId),
-                        content = Text(getMessage("table-heading.link")),
-                        attributes = Map("id" -> "poa2-more-details-date-link")
+        rows =
+            if(viewModel.poasHaveBeenAdjustedPreviously) {
```
avelx, line 160: not required in scope of this story, but we can refactor this into separate partial functions

avelx (dismissed):

avelx (approved):

PR #2344, opened 2024-06-13
MISUV-7677 Definition of payments on account to be added on POA drill down page

danielmaltz1:


avelx:
* this change would imply relevant changes where we testing documentDescription as well: have a look at next function validChargeTypeCondition  in FinancialDetailsModel: worth to create a tech debt I think ?

avelx (dismissed):

avelx (approved):

PR #2345, opened 2024-06-13
MISUV-7766 | Claim to Adjust POA | Update - Option (entry point) for adjusting POAs on WYO page

owen-gilfellon:
When both POA 1 and POA 2 are paid i.e. no payments on account due on 'What you owe' (WYO) page, we display the option to adjust POAs with some content.

### DIFF
--- a/app/models/financialDetails/DocumentDetail.scala
+++ b/app/models/financialDetails/DocumentDetail.scala
@@ -160,6 +160,11 @@ case class DocumentDetail(taxYear: Int,
     case _ => false
   }
 
+  val isPOA: Boolean = {
+    val poaDescriptions = Seq("ITSA- POA 1", "ITSA - POA 2")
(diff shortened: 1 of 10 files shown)


app/models/financialDetails/DocumentDetail.scala
```diff
@@ -160,6 +160,11 @@ case class DocumentDetail(taxYear: Int,
     case _ => false
   }
 
+  val isPOA: Boolean = {
+    val poaDescriptions = Seq("ITSA- POA 1", "ITSA - POA 2")
```
avelx, line 164: :question: - not sure how, but I thinking if we can move 
```
```
in some sort of constants ie. we can have ClaimToAdjustConstants trait ? not sure,

avelx (approved):

PR #2350, opened 2024-06-14
MISUV-7775 Content updates on POA drill down page

danielmaltz1:


app/views/ChargeSummary.scala.html
```diff
@@ -32,6 +32,8 @@
 @import enums.DocumentType
 @import enums.Poa1Charge
 @import enums.Poa2Charge
+@import enums.{BalancingCharge, OtherCharge, TRMAmmendCharge, TRMNewCharge}
```
avelx, line 35: this import is really not required

avelx (approved):

PR #2353, opened 2024-06-18
MISUV-7676 What you owe | Update Payments on Account (POA 1/POA 2) labels

danielmaltz1:


app/audit/Utilities.scala
```diff
@@ -48,8 +48,8 @@ object Utilities {
     (docDetail.getDocType, docDetail.documentText) match {
     case (_, Some(documentText)) if (documentText.contains("Class 2 National Insurance")) => Some("Class 2 National Insurance")
     case(_, Some(documentDescription)) if (documentDescription.contains("Cancelled PAYE Self Assessment")) => Some("Cancelled PAYE Self Assessment (through your PAYE tax code)")
-    case (Poa1Charge, _) => if (latePaymentCharge) Some("Late payment interest for payment on account 1 of 2") else Some("Payment on account 1 of 2")
-    case (Poa2Charge,_) => if (latePaymentCharge) Some("Late payment interest for payment on account 2 of 2") else Some("Payment on account 2 of 2")
+    case (Poa1Charge, _) => if (latePaymentCharge) Some("Late payment interest for payment on account 1 of 2") else Some("First payment on account")
```
simondrugan16, line 51: Could you use messages here? For if we need to change them in future?

test/views/ChargeSummaryViewSpec.scala
```diff
@@ -110,7 +110,9 @@ class ChargeSummaryViewSpec extends ViewSpec with FeatureSwitching {
     val poaTextBullets = messages("chargeSummary.paymentsOnAccount.bullet1") + " " + messages("chargeSummary.paymentsOnAccount.bullet2")
     val poaTextP2 = messages("chargeSummary.paymentsOnAccount.p2")
 
-    def poaHeading(year: Int, number: Int) = s"$taxYearHeading 6 April ${year - 1} to 5 April $year Payment on account $number of 2"
+    def poaHeading(year: Int, number: Int) = s"$taxYearHeading 6 April ${year - 1} to 5 April $year ${getFirstOrSecond(number)} payment on account"
+
+    def getFirstOrSecond(number: Int): String = if (number == 1) "First" else "Second"
```
avelx, line 115: this is a bit confusing, number can be only 1 or 2, so maybe something like:

```
def getFirstOrSecond(number: Int): String = {
      require(number > 0, "Number must be greater than zero")
      number match {
        case 1 => "First"
        case 2 => "Second"
        case _=> throw new Error(s"Number must be 1 or 2 but got: $number")
      }
    }
```

app/audit/Utilities.scala
danielmaltz1: Investigated, messages -> charge types doesn't match 1-to-1 so would need to define these strings locally anyway

avelx (approved):

PR #2354, opened 2024-06-18
MISUV-7834 Claim to adjust journey: add missing Future.recover block

tioadesola1:


app/controllers/claimToAdjustPoa/CheckYourAnswersController.scala
```diff
@@ -73,6 +73,9 @@ class CheckYourAnswersController @Inject()(val authorisedFunctions: AuthorisedFu
             logAndShowErrorPage(isAgent),
             view => view
           )
+        } recover {
+          case ex: Throwable => Logger("application").error(s"Unexpected error: ${ex.getMessage} - ${ex.getCause}")
```
simondrugan16, line 77: The logAndShowErrorPage method here already logs so we can remove the logger here

avelx (dismissed):

app/controllers/claimToAdjustPoa/SelectYourReasonController.scala
```diff
@@ -71,6 +71,9 @@ class SelectYourReasonController @Inject()(
           logAndShowErrorPage(isAgent),
           view => view
         )
+      } recover {
+        case ex: Throwable => Logger("application").error(s"Unexpected error: ${ex.getMessage} - ${ex.getCause}")
```
simondrugan16, line 75: Same here as above

tioadesola1: Same change case ex: Throwable => logAndShowErrorPage(isAgent)(ex)

danielmaltz1 (approved):

PR #2357, opened 2024-06-19
MISUV-7470: header-html-class, reset optout intent

james-king-hmrc:
header-html-class, reset optout intent

### DIFF
--- a/app/controllers/NextUpdatesController.scala
+++ b/app/controllers/NextUpdatesController.scala
@@ -75,11 +75,14 @@ class NextUpdatesController @Inject()(NoNextUpdatesView: NoNextUpdates,
             Future.successful(errorHandler.showInternalServerError())
           case (_, true) =>
             auditNextUpdates(user, isAgent, origin)
-            optOutService.getNextUpdatesQuarterlyReportingContentChecks.flatMap { checks =>
-              optOutService.nextUpdatesPageOptOutViewModel().map { optOutOneYearViewModel =>
-                Ok(nextUpdatesOptOutView(viewModel, optOutOneYearViewModel, checks, backUrl.url, isAgent, origin))
-              }
-            } recover {
+
+            val optOutSetup = for {
+              checks <- optOutService.getNextUpdatesQuarterlyReportingContentChecks
+              optOutOneYearViewModel <- optOutService.nextUpdatesPageOptOutViewModel()
+              _ <- optOutService.resetSavedIntent()
+            } yield Ok(nextUpdatesOptOutView(viewModel, optOutOneYearViewModel, checks, backUrl.url, isAgent, origin))
+
+            optOutSetup recover {
               case ex =>
                 Logger("application").error(s"Unexpected future failed error, ${ex.getMessage}")
                 errorHandler.showInternalServerError()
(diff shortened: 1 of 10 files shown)


app/controllers/NextUpdatesController.scala
```diff
@@ -75,11 +75,14 @@ class NextUpdatesController @Inject()(NoNextUpdatesView: NoNextUpdates,
...
-            optOutService.getNextUpdatesQuarterlyReportingContentChecks.flatMap { checks =>
-              optOutService.nextUpdatesPageOptOutViewModel().map { optOutOneYearViewModel =>
-                Ok(nextUpdatesOptOutView(viewModel, optOutOneYearViewModel, checks, backUrl.url, isAgent, origin))
-              }
-            } recover {
+
+            val optOutSetup = for {
+              checks <- optOutService.getNextUpdatesQuarterlyReportingContentChecks
+              optOutOneYearViewModel <- optOutService.nextUpdatesPageOptOutViewModel()
+              _ <- optOutService.resetSavedIntent()
```
avelx, line 82: you ignoring out come of this operation (_resetSavedIntent_()) and assuming it will be always success;
it is not guaranteed to be success all the time you have to handle it and raise error when not successful;

james-king-hmrc: happy to change but I don't think it's a good idea to show an error page just because the resetSavedIntent returns false, in any case there is ticket to gracefully degrade the opt-out service if any aspect if the feature fails, that should deal with opt-out error scenarios including this one, please reply to this comment

james-king-hmrc: will discuss with the team, thanks

ian-hmrc: Hey @avelx, I think this is a bit more subtle that just write success failure. I've just done some digging and the [Write Concern](https://www.mongodb.com/docs/v3.2/reference/write-concern/) in Mongo. It seems like we get a false value when we haven't written to enough (or specific) nodes within a timeout. It doesn't mean the write didn't happen (nodes could be temporarily down, etc.).

Since the consequence of this happening is not much (some radio-buttons may be pre-selected), plus the chances of it happening is pretty low, I'm definitely in favour of just leaving this as it is for now, as we need to get this out and it's not quite clear what we should do regarding this failure. We'll follow this up as James says in the ticket we have about graceful degradation. If we find anything interesting, we'll present our findings in a tech deep dive too.

avelx: @ian-hmrc  - I would suggest to crosscheck with CentalTech leads regarding best practice here as we are making quite fragile assumptions here; this scenario could happen, so unless we 100% sure that this would not cause issues , worth to cross check if we should handle this or not;

JordHemsworth (approved):
Approved on basis we are happy with discussion from Pavel

app/controllers/NextUpdatesController.scala
ian-hmrc: OK, in that case let's go ahead with what we have. I'm not concerned about the consequence of this as explained above (interestingly, there are several other similar bits of code already in the app that have the same approach). I'll check up on this with the CTL and make any changes necessary in upcoming tickets. Thanks.

PR #2361, opened 2024-06-20
MISUV-7470: fix agents path

james-king-hmrc:
fix agents path

### DIFF
--- a/app/views/optOut/OptOutError.scala.html
+++ b/app/views/optOut/OptOutError.scala.html
@@ -35,7 +35,11 @@
 }
 
 @getHomepageUrl = @{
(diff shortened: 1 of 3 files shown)


app/views/optOut/OptOutError.scala.html
```diff
@@ -35,7 +35,11 @@
 }
 
 @getHomepageUrl = @{
-    controllers.routes.HomeController.show().url
+    if(isAgent) {
```
SiddharthAtos, line 38: should fix: need to update view spec

SiddharthAtos (approved):

PR #2364, opened 2024-06-24
MISUV-7444 | Fixed labels on 'Credits and Refunds'

owen-gilfellon:
Updated Credits and Refund to display the correct message for credit type in all scenarios (inc. when there is a single credit or refund).

Tried to fix the payment mainTransaction (MISUV-7559), but needs AT updates so left commented out


app/models/financialDetails/CreditType.scala
```diff
@@ -53,8 +53,9 @@ object CreditType {
   private val mfaCredit = Range.inclusive(4004, 4025)
     .filterNot(_ == 4010).filterNot(_ == 4020).map(_.toString)
     .toList
-  // TODO: This should be "0060", not "4920" or "4930"
+  // TODO: MISUV-7759 - This should be "0060", not "4920" or "4930"
   private val payment = List("4920", "4930")
+  //  private val payment = List("0060")
```
avelx, line 58: I would assume we can tidy up commented code here?

test/views/CreditAndRefundsViewSpec.scala
```diff
@@ -90,313 +83,217 @@ class CreditAndRefundsViewSpec extends TestSupport with FeatureSwitching with Im
...
-              secondPendingAmountRequested = None,
-            )
-          )) {
+          creditAndRefundViewModel = ANewCreditAndRefundViewModel()
+            .withAvailableCredit(500.0)
+            .withAllocatedCredit(600.0)
+            .get()
+        ) {
           document.title() shouldBe creditAndRefundHeadingWithTitleServiceNameGovUk
+          println(document.text())
```
avelx, line 200: if we can drop this debug code pls

app/models/financialDetails/CreditType.scala
owen-gilfellon: It's for MISUV-7559 - I thought it better to get this ticket over the line, rather than carry it over because of the AT changes required

app/models/creditsandrefunds/CreditAndRefundViewModel.scala
```diff
@@ -26,33 +25,47 @@ sealed trait CreditRow {
...
+
+  private val creditsAndPayments: Seq[CreditRow] = sortCreditsByYear.flatMap(cc => {
     val (documentDetails, financialDetail) = cc
     val maybeCreditRow = for {
       creditType <- financialDetail.getCreditType
       amount <- documentDetails.documentDetail.paymentOrChargeCredit
     } yield {
       creditType match {
-        case PaymentType => PaymentCreditRow(amount = amount, creditType = creditType, date = documentDetails.dueDate.get)
+        case PaymentType => PaymentCreditRow(amount = amount, date = documentDetails.dueDate.get)
```
avelx, line 52: just wondering if dueDate is guarantied to be present in this context ? ( we are using .get call in line 52)

test/views/CreditAndRefundsViewSpec.scala
owen-gilfellon: Done

app/models/creditsandrefunds/CreditAndRefundViewModel.scala
owen-gilfellon: I think we need a refactoring ticket(s) to go through and remove .gets / exceptions on missing fields. Before the refactoring that introduced the credit rows, we were assuming dueDate would be present in this context - an exception was explicitly thrown in the view if it wasn't - https://github.com/hmrc/income-tax-view-change-frontend/blob/d9fc28a20ec474a25546ba7f69b040ff4674b0e5/app/views/CreditAndRefunds.scala.html#L235

avelx: we already have a story for this work: https://jira.tools.tax.service.gov.uk/browse/MISUV-7424
I hope we are going to continue work on this logic in the next sprint

avelx (approved):
approved with comments/ need ATs to be updated

PR #2365, opened 2024-06-25
MISUV-7713 SCP023 | Claim to Adjust POA | Update - Enter new POA amount page

ayubAz:


app/views/claimToAdjustPoa/EnterPoAAmountView.scala.html
```diff
@@ -53,6 +53,14 @@
     else controllers.routes.HomeController.show().url
 }
 
+@partiallyOrFullyPaidPoaExists = @{
```
avelx, line 56: I think these view level function is not required as this can be expressed on the level of view model ? + we can cover this with required UT for the view model ?

avelx (approved):

PR #2366, opened 2024-06-26
MISUV-7895 | Changed payment mainTransaction to 0060

owen-gilfellon:


avelx:
@owen-gilfellon  - ideally if we can add ITs test as well, to have more robust coverage around this issue, but really up to you;

avelx (dismissed):
approved with comment;

it/test/testConstants/IncomeSourceIntegrationTestConstants.scala
```diff
@@ -348,6 +351,50 @@ object IncomeSourceIntegrationTestConstants {
...
+      "paymentLotItem" -> "000001",
+    )
+  }
+
+  private def financialDetailsPayment( transactionId: String,
+                                       amount: BigDecimal,
+                                       clearingDate: LocalDate,
+                                       clearingSapDocument: String): JsObject = {
+    Json.obj(
+      "taxYear" -> s"$testTaxYear",
```
avelx, line 378: I don't think we need to use string interpolation here

avelx (approved):

it/test/testConstants/IncomeSourceIntegrationTestConstants.scala
owen-gilfellon: In FinancialDetail, taxYear is is String (Int in DocumentDetail)

PR #2367, opened 2024-06-26
MISUV-7770 POA Adjustment History is displayed on ITSA Viewer

danielmaltz1:


app/services/ChargeHistoryService.scala
```diff
@@ -0,0 +1,72 @@
...
+        val poaAdjustmentHistory: List[AdjustmentModel] = adjustments(chargeHistory.filter(_.poaAdjustmentReason.isDefined), documentDetail.originalAmount)
+        val otherAdjustmentHistory: List[AdjustmentModel] = chargeHistory.filter(_.poaAdjustmentReason.isEmpty).map(
+          event => AdjustmentModel(event.totalAmount, Some(event.reversalDate), event.reasonCode)
+        )
+        val fullAdjustmentHistory: List[AdjustmentModel] = poaAdjustmentHistory ++ otherAdjustmentHistory
+        AdjustmentHistoryModel(creation, fullAdjustmentHistory.sortBy(_.adjustmentDate))
+    }
+  }
+
+  private def adjustments(chargeHistory: List[ChargeHistoryModel], finalAmount: BigDecimal): List[AdjustmentModel] = {
```
avelx, line 59: if we can avoid recursive logic pls ? 
I am sure this can be simplified by use of .foldLeft / or fold

app/services/ChargeHistoryService.scala
```diff
@@ -0,0 +1,72 @@
...
+
+import javax.inject.Inject
+import scala.concurrent.{ExecutionContext, Future}
+
+class ChargeHistoryService @Inject()(chargeHistoryConnector: ChargeHistoryConnector) {
+
+  def chargeHistoryResponse(isLatePaymentCharge: Boolean, isPayeSelfAssessment: Boolean, chargeReference: Option[String],
+                                    isChargeHistoryEnabled: Boolean, isCodingOutEnabled: Boolean)
+                                   (implicit user: MtdItUser[_], hc: HeaderCarrier, ec: ExecutionContext): Future[Either[ChargeHistoryResponseModel, List[ChargeHistoryModel]]] = {
+    if (!isLatePaymentCharge && isChargeHistoryEnabled && !(isCodingOutEnabled && isPayeSelfAssessment)) {
```
avelx, line 33: I would suspect you were intended to defined this method in this way ?
with the next signature, to return first error if any or list of values in the List 
 
_Future[Either[ChargesHistoryErrorModel, List[ChargeHistoryModel]]]_

```
 def chargeHistoryResponse(isLatePaymentCharge: Boolean, isPayeSelfAssessment: Boolean, chargeReference: Option[String],
                                    isChargeHistoryEnabled: Boolean, isCodingOutEnabled: Boolean)
                                   (implicit user: MtdItUser[_], hc: HeaderCarrier, ec: ExecutionContext): Future[Either[ChargesHistoryErrorModel, List[ChargeHistoryModel]]] = {
    if (!isLatePaymentCharge && isChargeHistoryEnabled && !(isCodingOutEnabled && isPayeSelfAssessment)) {
      chargeHistoryConnector.getChargeHistory(user.nino, chargeReference).map {
        case chargesHistory: ChargesHistoryModel => Right(chargesHistory.chargeHistoryDetails.getOrElse(Nil))
        case errorResponse : ChargesHistoryErrorModel => Left(errorResponse)
      }
    } else {
      Future.successful(Right(Nil))
    }
  }
```

avelx: + conditions in the if statement is hard to read / and understand, we can refactor this later, but any condition were we have more than 2 values in the predicate we need to simplify in possible

app/models/chargeHistory/AdjustmentHistoryModel.scala
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.chargeHistory
+
+import java.time.LocalDate
+
+case class AdjustmentHistoryModel(creationEvent: AdjustmentModel,
```
avelx, line 21: Looks like we can use NonEmptyList with head representing creation event and history as a tail;
https://typelevel.org/cats/datatypes/nel.html;
but its really up to you to decided if you keen to use this type, current modelling approach is also ok;

app/views/ChargeSummary.scala.html
```diff
@@ -48,10 +50,10 @@
 )
 @(currentDate: LocalDate,
     documentDetailWithDueDate: DocumentDetailWithDueDate, backUrl: String,
-    paymentBreakdown: List[FinancialDetail], chargeHistory: List[ChargeHistoryModel], paymentAllocations: List[PaymentHistoryAllocations],
+    paymentBreakdown: List[FinancialDetail], paymentAllocations: List[PaymentHistoryAllocations],
```
avelx, line 53: next story created to refactor this view:
[MISUV-7901](https://jira.tools.tax.service.gov.uk/browse/MISUV-7901)

it/test/controllers/agent/ChargeSummaryControllerISpec.scala
```diff
@@ -490,7 +490,8 @@ class ChargeSummaryControllerISpec extends ComponentSpecBase with FeatureSwitchi
           "documentDescription" -> "ITSA- POA 1",
           "totalAmount" -> 123456789012345.67,
           "reversalDate" -> "2020-02-24",
-          "reversalReason" -> "amended return"
+          "reversalReason" -> "amended return",
+          "poaAdjustmentReason" -> "002"
```
avelx, line 494: if we can avoid "magic numbers" where possible =>
` "poaAdjustmentReason" -> OtherIncomeLower.code`

test/services/ChargeHistoryServiceSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+    mockChargeHistoryConnector
+  )
+
+  val testChargesHistoryModel: ChargesHistoryModel = ChargesHistoryModel("NINO", "AB123456C", "ITSA", None)
+
+  val testChargeHistoryErrorModel: ChargesHistoryErrorModel = ChargesHistoryErrorModel(INTERNAL_SERVER_ERROR, "Failure")
+
+  val testChargeHistory: List[ChargeHistoryModel] = List(ChargeHistoryModel(
+    taxYear = taxYear.toString, documentId = docNumber, documentDate = LocalDate.of(2021, 1, 1), documentDescription = "desc", totalAmount = 1000,
+    reversalDate = LocalDate.of(2021, 1, 1), reversalReason = "", poaAdjustmentReason = Some("001")
```
avelx, line 40: +
`  poaAdjustmentReason = Some(MainIncomeLower.code)`

app/models/chargeHistory/AdjustmentHistoryModel.scala
danielmaltz1: I think I prefer the current implementation, since we want to handle things differently based on whether the adjustments list is empty, I think this is clearer

avelx (approved):

PR #2368, opened 2024-06-26
 MISUV-7861-filter-fulfilled

james-king-hmrc:
include querterly and fulfilled, calculate counts, add missing param


app/models/nextUpdates/NextUpdatesResponseModel.scala
```diff
@@ -19,10 +19,16 @@ package models.nextUpdates
...
 import models.incomeSourceDetails.QuarterTypeElection.orderingByTypeName
 import models.incomeSourceDetails.{PropertyDetailsModel, QuarterReportingType, QuarterTypeCalendar, QuarterTypeStandard}
+import models.nextUpdates.ObligationStatus.Fulfilled
 import play.api.libs.json._
 
 import java.time.LocalDate
 
+case class ObligationStatus(code: String, name: String)
+object ObligationStatus {
+  val Open: ObligationStatus = ObligationStatus("O", "Open")
```
ian-hmrc, line 29: Would it be possible to use a sealed trait here, rather than holding a bunch of strings? We can then transform from the "O"/"F" representation when reading-in the json and just have that knowledge in that outer layer?

james-king-hmrc: please share an example of what you're thinking because I don't think it's possible to avoid the strings used here, thank you

app/controllers/optOut/OptOutChooseTaxYearController.scala
```diff
@@ -48,41 +47,36 @@ class OptOutChooseTaxYearController @Inject()(val optOutChooseTaxYear: OptOutCho
   def show(isAgent: Boolean = false): Action[AnyContent] = auth.authenticatedAction(isAgent) {
     implicit user =>
       for {
-        availableOptOutTaxYear <- optOutService.getTaxYearsAvailableForOptOut()
-        submissionCountForTaxYear <- optOutService.getQuarterlyUpdatesCountForTaxYear(availableOptOutTaxYear)
+        optOutProposition <- optOutService.fetchOptOutProposition()
```
ian-hmrc, line 50: With this more direct use of optOutProposition, optOutService.getTaxYearsAvailableForOptOut() is now only called from tests. Can this be removed?

james-king-hmrc: done, removed optOutService.getTaxYearsAvailableForOptOut()

ian-hmrc: Cool.

app/services/NextUpdatesService.scala
```diff
@@ -91,30 +93,60 @@ class NextUpdatesService @Inject()(val obligationsConnector: ObligationsConnecto
...
+
+    val isIncludeCountsForCurrentYear = isQueryTaxYearPreviousTaxYear && isCurrentOptOutTaxYearOffered
+
+    if (isIncludeCountsForCurrentYear) {
+
+      val sortedOffer = optOutProposition.availableTaxYearsForOptOut.sortBy(_.startYear)
+      val isDropLast = sortedOffer.size == 3
+      val yearsToProcess = if(isDropLast) sortedOffer.dropRight(LastItem) else sortedOffer
+
+      val responses = yearsToProcess.map(getQuarterlyUpdatesCountForTaxYear)
```
ian-hmrc, line 132: This code may call the API twice for CY, when it is needed to be included as part of the count for CY-1 and the count for CY. Could this be re-written to just get the quarterly updates count for each year, and then have a separate function that processes that to generate the cumulative counts from that data?

james-king-hmrc: isn't it doing that already? i can see it already "get the quarterly updates count for each year, and then have a separate function that processes that to generate the cumulative counts from that data", happy to get on a call to discuss this may be I'm not getting the comment here, thanks

app/models/nextUpdates/NextUpdatesResponseModel.scala
james-king-hmrc: sure will

app/services/NextUpdatesService.scala
james-king-hmrc: "yearsToProcess.map(getQuarterlyUpdatesCountForTaxYear)" is only called if the queried tax year is PY and CY is offered, not "for each of those outer tax years", not for CY it won't be called and not for NY it won't be called, and it won't even be called for PY if the CY is not offered, will call to discuss

ian-hmrc: OK, this is now resolved. :)

app/models/nextUpdates/NextUpdatesResponseModel.scala
ian-hmrc: Cool. :)

ian-hmrc (approved):
LGTM

PR #2369, opened 2024-06-27
MISUV-7889 opt out refactor updated years to tax year

ian-hmrc:
Refactors the logic for Opt Out to use the simpler TaxYear rather than the OptOutTaxYear


JordHemsworth (approved):
lgtm

PR #2371, opened 2024-06-27
Misuv 7898 Create view model for WhatYouNeedToKnow view

tioadesola1:


test/utils/claimToAdjust/JourneyCheckerClaimToAdjustSpec.scala
```diff
@@ -58,11 +58,15 @@ class JourneyCheckerClaimToAdjustSpec extends TestSupport with MockPaymentOnAcco
 
   val whatYouNeedToKnowView: WhatYouNeedToKnow = app.injector.instanceOf[WhatYouNeedToKnow]
 
-  def successfulFutureOk: PoAAmendmentData => Future[Result] = _ =>
-    Future.successful(Ok(whatYouNeedToKnowView(isAgent = false, TaxYear(2023, 2024), showIncreaseAfterPaymentContent = false, "")))
+  def successfulFutureOk: PoAAmendmentData => Future[Result] = _ => {
+    val viewModel = WhatYouNeedToKnowViewModel(poaTaxYear = TaxYear(2023, 2024), showIncreaseAfterPaymentContent = false, "")
```
simondrugan16, line 62: `val viewModel = WhatYouNeedToKnowViewModel(poaTaxYear = TaxYear(2023, 2024), showIncreaseAfterPaymentContent = false, "")`

is used quite a few times. It would be good to move this to the testConstants file
`testConstants/claimToAdjustPOA/ClaimToAdjustPOATestConstants.scala`
and use this val in all places for code maintainability

It would be nice to have an actual redirect location in here too rather than an empty string

simondrugan16 (approved):

PR #2373, opened 2024-07-01
MISUV-7889 - Add parameterised tests for the Opt Out logic based on the v15 spreadsheet

JordHemsworth:


test/services/optout/OptOutPropositionSpec.scala
```diff
@@ -0,0 +1,541 @@
...
+      /*  The programme has specified all required Opt Out scenarios in a large spreadsheet.
+          This code can ingest the data and convert to parameters for table based test found below.
+          This generator is currently ignored but can be reactivated to regenerate the test data if needed.
+            - To reactivate generator, replace "ignore" with "in"
+
+        Selected first 9 columns of Opt Out Scenarios Spreadsheet (currently v15).
+          - to avoid cells with carriage returns in to make parsing easier.
+        Paste into a fresh spreadsheet to save as OptOutScenarios.tsv in project root directory.
+          - csv has issues with commas within cells, therefore use tsv to allow simple parsing
+        Running test generates formatted test data in console, based on the required scenarios.
```
ian-hmrc, line 38: Be more assertive with the instructions. Maybe change 'Running the test generates...' to 'Run the test to generate the...'?

test/services/optout/OptOutPropositionSpec.scala
```diff
@@ -0,0 +1,541 @@
...
+          - Used to update 'scenarios' in table based test below.
+     */
+
+    val tsvOptOut = Source.fromFile("OptOutScenarios.tsv")
+
+    val lines = tsvOptOut.getLines().drop(2)
```
ian-hmrc, line 44: Could we move these 2 lines (the opening of the file and the dropping of the 2 header lines) to just above the loop over the lines? I feel that would improve cohesion and save having all the helper functions get in the way.

test/services/optout/OptOutPropositionSpec.scala
```diff
@@ -0,0 +1,541 @@
...
+    def parseTaxYears(taxYear: String): Seq[String] = {
+      taxYear match {
+        case "Nothing displayed" => Seq()
+        case "N/A" => Seq()
+        case _ => taxYear.replaceAll("\"", "").split("( [aA]nd |, )").map(_.trim).toSeq
+      }
+    }
+
+    def formatTaxYears(taxYears: Seq[String]) = {
+      "Seq(" ++ taxYears.map(option => s"\"$option\"").mkString(", ") ++ ")"
```
ian-hmrc, line 55: The ++ operator can be replaced with + operator in all cases in this file. More detail in [this stack overflow article](https://stackoverflow.com/questions/33074000/scala-string-vs) if needed.

test/services/optout/OptOutPropositionSpec.scala
```diff
@@ -0,0 +1,541 @@
...
+
+    def parseTaxYears(taxYear: String): Seq[String] = {
+      taxYear match {
+        case "Nothing displayed" => Seq()
+        case "N/A" => Seq()
+        case _ => taxYear.replaceAll("\"", "").split("( [aA]nd |, )").map(_.trim).toSeq
+      }
+    }
+
+    def formatTaxYears(taxYears: Seq[String]) = {
```
ian-hmrc, line 54: We omitted to give the return type for this function, could it be added back in to address the IDE warning?

test/services/optout/OptOutPropositionSpec.scala
```diff
@@ -0,0 +1,541 @@
...
+          quote(cells(3)),
+          formatTaxYears(parseTaxYears(cells(4))),
+          quote(parseCustomerIntent(cells(5))),
+          parseIsValid(cells(6)).toString,
+          formatTaxYears(parseTaxYears(cells(8)))
+        ).mkString(", ")
+      }),")
+    })
+
+    // Close the tsv file
```
ian-hmrc, line 92: This comment can be removed.

test/services/optout/OptOutPropositionSpec.scala
```diff
@@ -0,0 +1,541 @@
...
+
+    def quote(input: String): String = {
+      "\"" ++ input ++ "\""
+    }
+
+    def replaceEmptyWithSpace(input: String): String = {
+      if (input.isEmpty) " " else input
+    }
+
+    // Print the extracted data
```
ian-hmrc, line 74: This comment can be removed.

test/services/optout/OptOutPropositionSpec.scala
```diff
@@ -0,0 +1,541 @@
...
+        case "N/A" => Seq()
+        case _ => taxYear.replaceAll("\"", "").split("( [aA]nd |, )").map(_.trim).toSeq
+      }
+    }
+
+    def formatTaxYears(taxYears: Seq[String]) = {
+      "Seq(" ++ taxYears.map(option => s"\"$option\"").mkString(", ") ++ ")"
+    }
+
+    def parseIsValid(valid: String): Boolean = {
```
ian-hmrc, line 58: I think this could be more robust regarding the input. The following would be preferable.
```scala
    def parseIsValid(valid: String): Boolean =
      valid match {
      case "Allowed" => true
      case "N/A"     => false
      case _         => throw new RuntimeException("Unexpected entry in Valid column")
    }
```

test/services/optout/OptOutPropositionSpec.scala
```diff
@@ -0,0 +1,541 @@
...
+
+    def parseCustomerIntent(intent: String): String = {
+      intent.replaceAll("Customer wants to opt out from ", "")
+    }
+
+    def quote(input: String): String = {
+      "\"" ++ input ++ "\""
+    }
+
+    def replaceEmptyWithSpace(input: String): String = {
```
ian-hmrc, line 70: `replaceEmptyWithSpace()` is applied to all columns, whereas it's really only for the unknown ITSA statuses. I think we should remove it and replace it with specific parsing of the Crystallisation and ITSA statuses for the appropriate columns. This will make it clearer about what's being done and why, and to be more robust in parsing.

```scala

    def parseCrystallisedStatus(input: String): String =
      input match {
      case "Y" => "Y"
      case "N" => "N"
      case _   => throw new RuntimeException(s"Unexpected Crystallised status data ${input}")
    }

    def parseItsaStatus(input: String): String =
      input match {
      case "A" => "A"
      case "V" => "V"
      case "M" => "M"
      case ""  => " "
      case _   => throw new RuntimeException(s"Unexpected ITSA status data ${input}")
    }
```

test/services/optout/OptOutPropositionSpec.scala
```diff
@@ -0,0 +1,541 @@
...
+          This code can ingest the data and convert to parameters for table based test found below.
+          This generator is currently ignored but can be reactivated to regenerate the test data if needed.
+            - To reactivate generator, replace "ignore" with "in"
+
+        Selected first 9 columns of Opt Out Scenarios Spreadsheet (currently v15).
+          - to avoid cells with carriage returns in to make parsing easier.
+        Paste into a fresh spreadsheet to save as OptOutScenarios.tsv in project root directory.
+          - csv has issues with commas within cells, therefore use tsv to allow simple parsing
+        Running test generates formatted test data in console, based on the required scenarios.
+          - Used to update 'scenarios' in table based test below.
```
ian-hmrc, line 39: Make the instructions clearer... 'Copy the scenarios output on the console and paste them into the data table to update the following test in this file.'

ian-hmrc (approved):
LGTM :)

ian-hmrc (approved):
LGTM

vinay1yv (approved):

PR #2377, opened 2024-07-02
MISUV-7848 Graceful Degradation of Opt Out functionality on Next Updates page

tharunprasath:


app/controllers/NextUpdatesController.scala
```diff
@@ -76,17 +76,18 @@ class NextUpdatesController @Inject()(NoNextUpdatesView: NoNextUpdates,
...
-              checks <- optOutService.getNextUpdatesQuarterlyReportingContentChecks
-              optOutOneYearViewModel <- optOutService.nextUpdatesPageOptOutViewModel()
-              _ <- optOutService.resetSavedIntent()
-            } yield Ok(nextUpdatesOptOutView(viewModel, optOutOneYearViewModel, checks, backUrl.url, isAgent, origin))
-
-            optOutSetup recover {
-              case ex =>
-                Logger("application").error(s"Unexpected future failed error, ${ex.getMessage}")
-                errorHandler.showInternalServerError()
+            val optOutSetup = optOutService.getNextUpdatesQuarterlyReportingContentChecks.flatMap { checks =>
```
ian-hmrc, line 79: Can we just keep the call for the `checks`  inside the for comprehension? I think that as long as the whole comprehension is within braces, then the `recoverWith()` can apply to that.

tharunprasath: Done

it/test/controllers/NextUpdatesControllerISpec.scala
```diff
@@ -354,10 +354,72 @@ class NextUpdatesControllerISpec extends ComponentSpecBase {
...
+          val previousYear = currentTaxYear.addYears(-1)
+
+
+          IncomeTaxViewChangeStub.stubGetIncomeSourceDetailsResponse(testMtditid)(OK, propertyOnlyResponse)
+
+          IncomeTaxViewChangeStub.stubGetNextUpdates(testNino, ObligationsModel(Seq(singleObligationQuarterlyModel(testPropertyIncomeId))))
+
+          IncomeTaxViewChangeStub.stubGetFulfilledObligationsNotFound(testNino)
+          ITSAStatusDetailsStub.stubGetITSAStatusDetails(previousYear.formatTaxYearRange)
+          CalculationListStub.stubGetLegacyCalculationListError(testNino, previousYear.endYear.toString)
```
ian-hmrc, line 377: Is it possible to re-phrase this in terms of the Crystallisation Status that we're interested in? Maybe simply wrap this call in something that expresses that. Also, the test name "Calculation API Failure" could change too?

app/controllers/NextUpdatesController.scala
ian-hmrc: Cool.

ian-hmrc (approved):
LGTM

PR #2378, opened 2024-07-02
MISUV-7874 R12 TxM event | Transitional profit to be sent in TxM event for 'Tax Calculation Details'

danielmaltz1:


avelx:
@danielmaltz1  - can we update ITs tests as well? thanks
>it/test/controllers/agent/TaxDueSummaryControllerISpec.scala
>it/test/controllers/TaxDueSummaryControllerISpec.scala

avelx (approved):

PR #2382, opened 2024-07-03
MISUV-7847: Reduce API calls for Opt Out in NextUpdatesController

SiddharthAtos:


app/services/optout/OptOutService.scala
```diff
@@ -157,43 +135,54 @@ class OptOutService @Inject()(itsaStatusUpdateConnector: ITSAStatusUpdateConnect
...
-      responsesSeqOfFutures  <- OptionT(Future.successful(Option(makeUpdateCalls(yearsToUpdate))))
+      responsesSeqOfFutures <- OptionT(Future.successful(Option(makeUpdateCalls(yearsToUpdate))))
       responsesSeq <- OptionT(Future.sequence(responsesSeqOfFutures).map(v => Option(v)))
       finalResponse <- OptionT(Future.successful(Option(findAnyFailOrFirstSuccess(responsesSeq))))
     } yield finalResponse
 
     result.getOrElse(OptOutUpdateResponseFailure.defaultFailure())
   }
 
+  def nextUpdatesPageOptOutViewModel(proposition: OptOutProposition)(implicit user: MtdItUser[_], hc: HeaderCarrier, ec: ExecutionContext): Option[OptOutViewModel] = {
```
james-king-hmrc, line 154: unused params, can be removed: (implicit user: MtdItUser[_], hc: HeaderCarrier, ec: ExecutionContext)

app/controllers/NextUpdatesController.scala
```diff
@@ -77,8 +77,7 @@ class NextUpdatesController @Inject()(NoNextUpdatesView: NoNextUpdates,
             auditNextUpdates(user, isAgent, origin)
 
             val optOutSetup = for {
-              checks <- optOutService.getNextUpdatesQuarterlyReportingContentChecks
-              optOutOneYearViewModel <- optOutService.nextUpdatesPageOptOutViewModel()
+              (optOutOneYearViewModel, checks) <- optOutService.nextUpdatesPageOptOutWithChecks()
```
james-king-hmrc, line 80: not covered by unit test
Screenshot 2024-07-04 at 11 28 32

james-king-hmrc, line 80: controller argument 'incomeSourceDetailsService: IncomeSourceDetailsService,' not used can be removed

app/services/optout/OptOutService.scala
```diff
@@ -157,43 +135,54 @@ class OptOutService @Inject()(itsaStatusUpdateConnector: ITSAStatusUpdateConnect
...
+
+  def nextUpdatesPageOptOutWithChecks()(implicit user: MtdItUser[_], hc: HeaderCarrier, ec: ExecutionContext)
+  : Future[(Option[OptOutViewModel], NextUpdatesQuarterlyReportingContentChecks)] = {
+
     fetchOptOutProposition().map { proposition =>
-      proposition.optOutPropositionType.flatMap {
-        case p: OneYearOptOutProposition => Some(OptOutOneYearViewModel(oneYearOptOutTaxYear = p.intent.taxYear, state = p.state()))
-        case _: MultiYearOptOutProposition => Some(OptOutMultiYearViewModel())
-      }
+      (nextUpdatesPageOptOutViewModel(proposition),
```
james-king-hmrc, line 169: not covered by unit test
Screenshot 2024-07-04 at 11 42 31

app/models/optout/NextUpdatesQuarterlyReportingContentChecks.scala
```diff
@@ -16,11 +16,16 @@
 
 package models.optout
 
-case class NextUpdatesQuarterlyReportingContentChecks(currentYearItsaStatus: Boolean, previousYearItsaStatus: Boolean, previousYearCrystallisedStatus: Option[Boolean]) {
+import models.itsaStatus.ITSAStatus.{Mandated, Voluntary}
+import services.optout.OptOutProposition
 
-  private def showOptOutContent: Option[Unit] = if (currentYearItsaStatus || (previousYearItsaStatus && !previousYearCrystallisedStatus.getOrElse(false))) Some({}) else None
+case class NextUpdatesQuarterlyReportingContentChecks(proposition: OptOutProposition) {
```
james-king-hmrc, line 22: this class can be named something like OptOutStatesViewModel, the important point is 'xxxViewModel' to indicate this will be passed to view page

app/models/optout/NextUpdatesQuarterlyReportingContentChecks.scala
```diff
@@ -16,11 +16,16 @@
...
 
-  def showUpdateTypeDetailsSection: Option[Unit] = showOptOutContent
+  private val isCurrentYearMandatedOrVoluntary = proposition.currentTaxYear.status == Mandated || proposition.currentTaxYear.status == Voluntary
+  private val isPreviousYearMandatedOrVoluntary = proposition.previousTaxYear.status == Mandated || proposition.previousTaxYear.status == Voluntary
+  private val previousYearCrystallisedStatus = proposition.previousTaxYear.crystallised
+  private val showOptOutContent: Option[Unit] = if (isCurrentYearMandatedOrVoluntary || (isPreviousYearMandatedOrVoluntary && !previousYearCrystallisedStatus)) Some({}) else None
```
james-king-hmrc, line 27: 'Option[Unit]' lacks expressiveness, why not use Boolean?

app/controllers/NextUpdatesController.scala
james-king-hmrc, line 80: the tuple '(optOutOneYearViewModel, checks)' can be turned to a case class named NextUpdatesOptOutViewModel, more readable and reduces arguments passed to view

app/models/optout/NextUpdatesQuarterlyReportingContentChecks.scala
SiddharthAtos: It makes easier when using  in `scala.html` with the help of `.map`

SiddharthAtos: something like this - https://github.com/hmrc/income-tax-view-change-frontend/blob/main/app/views/nextUpdates/NextUpdatesOptOut.scala.html#L107

james-king-hmrc: I don't think that's a good reason to write code in that form, @ian-hmrc  what do you think?

SiddharthAtos: it makes templates look cleaner, in contract you will have to add `@if  else`  in templates

ian-hmrc: I think there's a bit of subtlety to using Options with the views. I think it reads well when the whole view-model inside the option may or may not be present.

The classic example of this was the early version of the single-year opt-out, when, if you're in a single-year scenario, there is a model with all the data to support building the required elements on the Next Updates page, and when you're not a single-year scenario, you don't have that view model at all. Mapping that Option in the view really makes sense here, as if that optional view model is present, all of it's elements can be used.

It's much better to do things this way and make the whole sub-section of the view model optional so you have it or you don't, rather than having each field in the view model optional and having to wrap the whole section of the view in an `if (someRandomFieldFromTheViewModel.isDefined) {...}` block, or relying on the view correctly ignoring individual Options when they are None. It expresses what you do or don't have much more clearly.

However, in this case, we don't use a view model. If the option is present, the Unit inside it can't be used as a view-model as it does not contain a value, which is also reinforced by the use of the underscore for the lambda variable to throw it away in the view...

```scala
	@nextUpdatesOptOutViewModel.showUpdateTypeDetailsSection.map { _ =>
	    @detailsDropdown("nextUpdates.dropdown.info", detailsContent)
	}
``` 

This shows we're not getting all the value out of the Option with the code as it stands.

One way to get more value out of this would be to keep the Option grow the Unit to a more fully-fledged view model. However, in this particular scenario, we're just displaying static content, so there's nothing to add to the Unit to make it a fuller view-model.

The other way forward here is to remove the Option and convert this to a simple Boolean. And since we have no need of the Option as mentioned above, I think this is the most pragmatic solution in this case.

SiddharthAtos: Right! understood. removing the options.

app/controllers/NextUpdatesController.scala
```diff
@@ -75,21 +74,15 @@ class NextUpdatesController @Inject()(NoNextUpdatesView: NoNextUpdates,
...
-
-            val optOutSetup = {
-              for {
-                checks <- optOutService.getNextUpdatesQuarterlyReportingContentChecks
-                optOutOneYearViewModel <- optOutService.nextUpdatesPageOptOutViewModel()
-                _ <- optOutService.resetSavedIntent()
-              } yield Ok(nextUpdatesOptOutView(viewModel, optOutOneYearViewModel, checks, backUrl.url, isAgent, origin))
-            }.recoverWith {
+            (for {
+              nextUpdatesOptOutViewModel <- optOutService.getNextUpdatesOptOutViewModel()
```
james-king-hmrc, line 78: not covered by unit test
Screenshot 2024-07-08 at 16 07 07

app/views/nextUpdates/NextUpdatesOptOut.scala.html
```diff
@@ -54,14 +53,14 @@
 
 	@h1(msg = "nextUpdates.heading")
 
-	@checks.showUpdateTypeDetailsSection.map { _ =>
+	@if(nextUpdatesOptOutViewModel.showUpdateTypeDetailsSection){
```
james-king-hmrc, line 56: nextUpdatesOptOutViewModel can be named 'optOutModel' to make it shorter

app/views/nextUpdates/NextUpdatesOptOut.scala.html
```diff
@@ -37,8 +37,7 @@
 	nextUpdatesHelper: NextUpdatesHelper
 )
 
-@(currentObligations: NextUpdatesViewModel, optOutViewModel: Option[OptOutViewModel],
-  checks:NextUpdatesQuarterlyReportingContentChecks,
+@(currentObligations: NextUpdatesViewModel, nextUpdatesOptOutViewModel: NextUpdatesOptOutViewModel,
```
james-king-hmrc, line 40: currentObligations can be named 'obligationsModel' to follow same format as 'optOutModel' below

test/controllers/NextUpdatesControllerSpec.scala
```diff
@@ -93,24 +94,24 @@ class NextUpdatesControllerSpec extends MockAuthenticationPredicate with MockInc
     NextUpdatesModel(BaseTestConstants.testPropertyIncomeId, List(NextUpdateModel(fixedDate, fixedDate, fixedDate, "EOPS", Some(fixedDate), "EOPS", Fulfilled)))
   ))
 
+  private def getQuarterType(string: String) = {
```
james-king-hmrc, line 97: 'string' can be 'typeName'

james-king-hmrc (approved):

PR #2387, opened 2024-07-04
MISUV-7628-add-time-machine-via-config

simondrugan16:
Re enabling the time machine feature switch which will be evaluated from config


app/config/FrontendAppConfig.scala
```diff
@@ -181,6 +179,14 @@ class FrontendAppConfig @Inject()(val servicesConfig: ServicesConfig, val config
 
   val encryptionIsEnabled: Boolean = config.get[Boolean]("encryption.isEnabled")
 
-  lazy val readFeatureSwitchesFromMongo = servicesConfig.getBoolean("feature-switches.read-from-mongo")
+  lazy val readFeatureSwitchesFromMongo: Boolean = servicesConfig.getBoolean("feature-switches.read-from-mongo")
+
+  lazy val isTimeMachineEnabled: Boolean = servicesConfig.getBoolean("time-machine.enable-time-machine")
+
+  lazy val isTimeMachineAddYearEnabled: Boolean = servicesConfig.getBoolean("time-machine.add-year.enable-time-machine-add-year")
```
avelx, line 186: I don't think we need to have so many config entries: for days, months and years; should be fine to have delta for day ie

simondrugan16: I think a delta would be more confusing. To go from today to Jan 31st you would have to work out the Delta which is annoying to do. Plus, if you wanted to end up on the same specific date when testing over a few days, you would have to change the config once it hits midnight

app/models/admin/TimeMachine.scala
```diff
@@ -0,0 +1,48 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.admin
+
+case class TimeMachine(isTimeMachineEnabled: Boolean,
```
avelx, line 19: assuming we don't need to so many config entries, this type would become redundant; 
+ if we really need one, this can be private type within DateService itself

test/services/DateServiceSpec.scala
```diff
@@ -38,35 +43,126 @@ class DateServiceSpec extends TestSupport with FeatureSwitching {
...
 
-//        if (timeMachineIsOn) {
-//          LocalDate.parse(date).plusYears(1)
-//        } else {
-          LocalDate.parse(date)
-//        }
+        //        if (timeMachineIsOn) {
+        //          LocalDate.parse(date).plusYears(1)
+        //        } else {
+        LocalDate.parse(date)
```
avelx, line 74: as its been mention quite a while ago, this can fail for some dates on the OS where dates got a different format say: YYYYMMDD => YYYYDDMM, so success of this operation is not guarantied

test/services/DateServiceSpec.scala
```diff
@@ -38,35 +43,126 @@ class DateServiceSpec extends TestSupport with FeatureSwitching {
...
+
+        TestDateService.getCurrentDate shouldBe expectedDate
+      }
+    }
+  }
+  "return a LocalDate, two calendar years in the future time on the 9th of April" when {
+    "the TimeMachine is on, and we have set the TimeMachine to place us on a specific date with additional years" in {
+      setupGetCurrentTestMocks(timeMachineFS = true, timeMachineAddYearFS = true, addYears = 2, timeMachineSetDateFS = true, setDay = 9, setMonth = 4)
+
+      val expectedDate = LocalDate.of(LocalDate.now.plusYears(2).getYear, 4, 9)
```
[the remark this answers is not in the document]
avelx: + try added/implement next method to the DateServiceInterface
++ call it as a single source of truth while evaluating date, you can override it where needed in the UTs
```
trait DateServiceInterface {
  protected def now(): LocalDate
```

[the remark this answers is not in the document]
simondrugan16: You said it was OK to use this if it won't break in the future, which these won't. For testing the other methods in the class (aka ones that aren't getCurrentDate) then I would not use this approach

app/services/admin/FeatureSwitchService.scala
```diff
@@ -65,15 +65,26 @@ class FeatureSwitchService @Inject()(
...
 
-  def setAll(featureSwitches: Map[FeatureSwitchName, Boolean]): Future[Unit] =
-    featureSwitchRepository.setFeatureSwitches(featureSwitches)
+  def setAll(featureSwitches: Map[FeatureSwitchName, Boolean]): Future[Unit] = {
+    Logger("application").info(s"Setting all feature switches. FS values: $featureSwitches")
+    if (appConfig.readFeatureSwitchesFromMongo) {
+      featureSwitchRepository.setFeatureSwitches(featureSwitches)
+    } else {
+      featureSwitches.foreach(featureSwitch =>
+        setFS(featureSwitch._1, featureSwitch._2)
```
avelx, line 84: resolving tuples in this fashion is generally discouraged as its hard to read/understand these lines

conf/application.conf
```diff
@@ -304,7 +303,9 @@ auth {
 }
 
 time-machine {
-    add-years = 1
+    enable-time-machine = false
```
avelx, line 306: not sure why we moved enable-time-machine under time-machine section?
this is the same FS as the rest of the FSs we have so far, no need to introduce this complexity

app/config/TimeMachine.scala
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+package config
+
+trait TimeMachine {
+
+  val appConfig: FrontendAppConfig
+
+  def timeMachine: TimeMachineModel = {
```
avelx, line 23: => getTimeMachineConfig

app/config/TimeMachine.scala
```diff
@@ -0,0 +1,37 @@
...
+    TimeMachineModel(
+      isTimeMachineEnabled = appConfig.isTimeMachineEnabled,
+      addYears = appConfig.timeMachineAddYears,
+      addDays = appConfig.timeMachineAddDays
+    )
+  }
+
+  case class TimeMachineModel(
```
avelx, line 31: TimeMachineModel => TimeMachineSettings

conf/application.conf
simondrugan16: Because it's not going to be on the FS list on the FS page I thought to make it a bit separate. Do you thinks it's better to move it to where all the other FS are?

test/services/DateServiceSpec.scala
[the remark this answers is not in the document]
avelx: I would really suggest to avoid use of dynamic data/generated data, again, this is not fit in the way we use UT or ITs;
its is expected to use static input, its hard to understand how it works and support this code in the futurel 
Dynamic data / generated input is fit under category of property based testing, and this is not property base test;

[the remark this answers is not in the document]
simondrugan16: How am I supposed to unit test this then? I can't mock a response from LocalDate.now() so I don't know how I can test/property based test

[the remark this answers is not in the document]
avelx: you don't have to mock LocalDate.now(), you can apply change proposed above, like:
and override => protected def now(): LocalDate in the fake you are going to use in the tests
```
trait DateServiceInterface {
  protected def now(): LocalDate

  def getCurrentDate: LocalDate = now()
```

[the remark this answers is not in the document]
simondrugan16: Ah ok I'll give it a go

app/services/DateService.scala
```diff
@@ -16,28 +16,28 @@
...
-//    } else {
+  def now(): LocalDate = LocalDate.now()
+
+  override def getCurrentDate: LocalDate = {
+    if (getTimeMachineConfig.isTimeMachineEnabled) {
+      now()
+        .plusYears(getTimeMachineConfig.addYears)
+        .plusDays(getTimeMachineConfig.addDays)
+    } else {
       LocalDate.now()
```
avelx, line 39: you would have to call now() in line 39, but I guess this would not fail in UT and ITs as we are using fake

it/test/helpers/ComponentSpecBase.scala
```diff
@@ -162,7 +162,10 @@ trait ComponentSpecBase extends TestSuite with CustomMatchers
     "encryption.isEnabled" -> "false",
     "microservice.services.contact-frontend.host" -> mockHost,
     "microservice.services.contact-frontend.port" -> mockPort,
-    "feature-switches.read-from-mongo" -> "false"
+    "feature-switches.read-from-mongo" -> "false",
+    "time-machine.enable-time-machine" -> "false",
```
avelx, line 166: I think you dropped this entry ? and its not required / both ComponentBaseSpecs

danielmaltz1 (approved):

PR #2388, opened 2024-07-05
MISUV-7909-use-sealed-trait

james-king-hmrc:
use-sealed-trait


app/models/nextUpdates/NextUpdatesResponseModel.scala
```diff
@@ -136,3 +131,27 @@ object NextUpdatesModel {
 object NextUpdatesErrorModel {
   implicit val format: Format[NextUpdatesErrorModel] = Json.format[NextUpdatesErrorModel]
 }
+
+sealed trait ObligationStatus
```
joesheridan, line 135: looks good thanks James.
Can we copy these changes to the BE repo? We should have similar models FE/BE.

james-king-hmrc: yes of course Joe, will discuss with @ian-hmrc and create tickets appropriately

joesheridan (approved):

PR #2390, opened 2024-07-08
MISUV-7850: Claim to Adjust: refactor view models

danielmaltz1:


app/models/claimToAdjustPoa/PaymentOnAccountViewModel.scala
```diff
@@ -22,15 +22,17 @@ case class PaymentOnAccountViewModel(
...
-                             paymentOnAccountOne: BigDecimal,
-                             paymentOnAccountTwo: BigDecimal,
-                             poARelevantAmountOne: BigDecimal,
-                             poARelevantAmountTwo: BigDecimal,
-                             poAPartiallyPaid: Boolean
+                             totalAmountOne: BigDecimal,
+                             totalAmountTwo: BigDecimal,
+                             relevantAmountOne: BigDecimal,
+                             relevantAmountTwo: BigDecimal,
+                             poaPreviouslyAdjusted: Option[Boolean],
```
avelx, line 29: naming convention, I guess we don't need "poa" prefix at all as are already in the context of Poa~ view model;

app/views/claimToAdjustPoa/EnterPoAAmountView.scala.html
```diff
@@ -81,7 +81,7 @@
 }
 
 @getRow1 = @{
-    if(viewModel.poaPreviouslyAdjusted) {
+    if(viewModel.poaPreviouslyAdjusted.contains(true)) {
```
avelx, line 84: if we can move this part of the logic on the viewModel level ?
`> viewModel.poaPreviouslyAdjusted.contains(true)`

app/views/claimToAdjustPoa/EnterPoAAmountView.scala.html
```diff
@@ -157,12 +157,12 @@
         firstCellIsHeader = true
     ))
 
-    @if(viewModel.partiallyOrFullyPaidPoaExists && !viewModel.poaPreviouslyAdjusted) {
+    @if(viewModel.partiallyOrFullyPaidPoaExists && !viewModel.poaPreviouslyAdjusted.contains(true)) {
```
avelx, line 160: lines 160 and 165 seems to be also something we can move on the viewModel level

avelx:
might be a good idea to create a separate folder for viewModel, 
ie => /app/models/claimToAdjustPoa/viewModels/ and move all viewModels specific to this sub-domain in there;

avelx (approved):

PR #2392, opened 2024-07-09
MISUV-7606: change return type of isTaxYearCrystallised to Future[Boolean]

ayubAz:


app/services/CalculationListService.scala
```diff
@@ -78,9 +78,9 @@ class CalculationListService @Inject()(calculationListConnector: CalculationList
     }
 
     isCrystallised.map {
-      case None => Some(false)//when the flag is not present then default to false i.e not-crystallised
-      case Some(b) => Some(b)
+      case None        => false //when the flag is not present then default to false i.e not-crystallised
```
avelx, line 81: consider simplify this pattern matching:
```
isCrystallised.map {
      case None | Some(false)      => false
      case Some(true)  => true
    }
```

avelx (approved):
approved with comments

avelx (approved):

PR #2394, opened 2024-07-10
Misuv 7945 remove use of mongo in 1 year opt out

ian-hmrc:
This PR removes the need, within the handling of the submit action from the Opt Out Confirmation Page, to persist the tax year to opt-out for in mongo, just to pull it back again in the next function.

This simplification should make it easier to persist the 3 ITSA statuses and the Crystallised status to mongo, giving a consistent context in which to generate all the pages in the journey and to address the recent issues we've seen in E2E testing.

Moreover, I think the original code could have had a race condition in it in that the write to the DB was implemented async, yet not waited for before reading it back. Thought I've not reproduce the race condition, I think we're safer without this possibility.

I would have liked to go further and transform the `makeOptOutUpdateRequest()` function not take a `Future`, but instead the `Option[Boolean]` and use it in a `flatMap()` at it's call site, but I've not been able to wrap my head around how to do this because of the complexity of the for comprehension inside that function (though I have simplified it a little). This may be possible with a future PR, but I didn't want to block this PR trying to resolve this as we have more pressing matters at hand.

### DIFF
--- a/app/services/optout/OptOutService.scala
+++ b/app/services/optout/OptOutService.scala
@@ -165,8 +165,8 @@ class OptOutService @Inject()(itsaStatusUpdateConnector: ITSAStatusUpdateConnect
 
     val result = for {
       intentTaxYear <- OptionT(intentFuture)
-      yearsToUpdate <- OptionT(Future.successful(Option(optOutProposition.optOutYearsToUpdate(intentTaxYear))))
-      responsesSeqOfFutures  <- OptionT(Future.successful(Option(makeUpdateCalls(yearsToUpdate))))
+      yearsToUpdate = optOutProposition.optOutYearsToUpdate(intentTaxYear)
+      responsesSeqOfFutures = makeUpdateCalls(yearsToUpdate)
       responsesSeq <- OptionT(Future.sequence(responsesSeqOfFutures).map(v => Option(v)))
     } yield findAnyFailOrFirstSuccess(responsesSeq)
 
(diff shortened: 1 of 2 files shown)


app/services/optout/OptOutService.scala
```diff
@@ -150,7 +149,7 @@ class OptOutService @Inject()(itsaStatusUpdateConnector: ITSAStatusUpdateConnect
     repository.set(data)
   }
 
-  def makeOptOutUpdateRequest(optOutProposition: OptOutProposition)
+  def makeOptOutUpdateRequest(optOutProposition: OptOutProposition, intentFuture: Future[Option[TaxYear]])
```
james-king-hmrc, line 152: I understand passing Future as method argument isn't within our coding standards, is that correct @avelx

james-king-hmrc: just thinking, how about passing 'intentCallback: () => Future[Option[TaxYear]]' instead?

ian-hmrc: I mentioned in the PR description that I'm not a big fan of this approach and would like to get to a cleaner implementation, but have been hampered by the complexity of the for comprehension. I'd rather do this by (flat)mapping the result of getting the customer intent into a function that would then make the ITSA status upsert requests in that context, but we're short on time to fix the E2E testing issue so I'm proposing this as a pragmatic solution for now.

james-king-hmrc (approved):

app/services/optout/OptOutService.scala
james-king-hmrc: approved

PR #2395, opened 2024-07-11
MISUV-7931: CamelCase output in Error Summary - Income Tax V&C - Income Source

Aashvin:


app/forms/incomeSources/cease/IncomeSourceEndDateForm.scala
```diff
@@ -84,21 +86,27 @@ class IncomeSourceEndDateForm @Inject()(val dateService: DateService)(implicit v
...
+            case (day, month, year) =>
+              LocalDate.of(year.toInt, month.toInt, day.toInt)
+          },
+          date => (date.getDayOfMonth.toString, date.getMonthValue.toString, date.getYear.toString)
+        ).verifying(firstError(dateConstraints: _*))
       )(DateFormElement.apply)(DateFormElement.unapply))
   }
 
-  private def checkRequiredFields: Constraint[(String, String, String)] = Constraint("constraints.requiredFields") {
+  private def dateMustBeCompleteKey(incomeSourceType: IncomeSourceType): String = incomeSourceType match {
```
james-king-hmrc, line 101: nice-to-have: i would move dateMustBeCompleteKey method to below checkRequiredFields method, since dateMustBeCompleteKey is called from checkRequiredFields

app/forms/incomeSources/cease/IncomeSourceEndDateForm.scala
```diff
@@ -30,9 +30,11 @@ import services.DateService
 import java.time.LocalDate
 import javax.inject.Inject
 
-class IncomeSourceEndDateForm @Inject()(val dateService: DateService)(implicit val appConfig: FrontendAppConfig) extends CustomConstraints with FeatureSwitching{
+class IncomeSourceEndDateForm @Inject()(val dateService: DateService)(implicit val appConfig: FrontendAppConfig) extends CustomConstraints with FeatureSwitching {
```
james-king-hmrc, line 33: I can see a lot of file format validation highlights, can these be addressed?

Screenshot 2024-07-11 at 11 15 38

app/forms/incomeSources/cease/IncomeSourceEndDateForm.scala
```diff
@@ -30,9 +30,11 @@ import services.DateService
...
 
-  val dateMustBeComplete = "dateForm.error.dayMonthAndYear.required"
+  val dateMustBeCompleteSE = "dateForm.error.dayMonthAndYear.required.se"
```
simondrugan16, line 35: This change is in the incomeSources (old journey) domain. So here the old journey is being updated with the new journey content. If we do this we might get more entangled with the old and new journeys. I think it would be better if dateForm.error.dayMonthAndYear.required was kept as is, but the deleted message in the messages/welsh messages file was re-added?

app/forms/incomeSources/cease/IncomeSourceEndDateForm.scala
james-king-hmrc, line 33: I don't see a IT test updates for this, please add related IT test

Aashvin: The code formatting aspects of this have been addressed, but there are still a couple more warnings - should these be addressed too?

james-king-hmrc: should we have IT tests for this change @ian-hmrc?

app/forms/incomeSources/cease/IncomeSourceEndDateForm.scala
simondrugan16: Yep, no specific text. It's just that the new journey was a replica of the old journey, so the new version is the updated one, and this file will be deleted soon, so while it's not wrong, there's not a massive benefit to implementing the new logic. I don't know if design would want parts of the new design in the old (currently in prod) version or not (for consistency purposes maybe)

app/forms/incomeSources/cease/IncomeSourceEndDateForm.scala
Aashvin: This function no longer exists from addressing Simon's comment

PR #2402, opened 2024-07-12
Misuv 7945 fix e2 e bug due to itsa status change

ian-hmrc:


app/models/optout/OptOutContextData.scala
```diff
@@ -0,0 +1,49 @@
...
+  implicit val format: OFormat[OptOutContextData] = Json.format[OptOutContextData]
+
+  def statusToString(status: ITSAStatus): String =
+    status match {
+      case ITSAStatus.NoStatus  => "U"
+      case ITSAStatus.Voluntary => "V"
+      case ITSAStatus.Annual    => "A"
+      case ITSAStatus.Mandated  => "M"
+      // This will be validated earlier on in a future ticket
+      case _ => throw new RuntimeException("Unexpected status")
```
james-king-hmrc, line 38: I feel throwing exceptions isn't within the spirit of the coding style of this project, using Either is more appropriate in my view

app/services/optout/OptOutService.scala
```diff
@@ -128,13 +156,26 @@ class OptOutService @Inject()(itsaStatusUpdateConnector: ITSAStatusUpdateConnect
...
 
-  def saveIntent(intent: TaxYear)(implicit hc: HeaderCarrier): Future[Boolean] = {
-    val data = UIJourneySessionData(
-      sessionId = hc.sessionId.get.value,
-      journeyType = OptOutJourney.Name,
-      optOutSessionData = Some(OptOutSessionData(selectedOptOutYear = Some(intent.toString)))
-    )
-    repository.set(data)
+  def saveIntent(intent: TaxYear)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Boolean] = {
+    OptionT(repository.get(hc.sessionId.get.value, OptOutJourney.Name)).
```
james-king-hmrc, line 160: let's try and simplify to imporve readability

app/models/optout/OptOutContextData.scala
ian-hmrc: Yes, this is a pragmatic solution to the problem here, and we have a ticket in the sprint to do this better.

I'm still unsure whether an exception or Either is better here. If we have an Either, then it may affect a lot of the code we have and we'll have to make sure this error case will weave its way through all the layers and show the error page. My thoughts are leaning towards using an exception as we've been promised that we should never get any of the other states from the API, and the Opt-Out design does not account for other states. To me it's pretty similar to not getting a response back from the API. Anyway, I'd rather defer this decision for now and pick up with the other ticket.

app/services/optout/OptOutService.scala
```diff
@@ -50,40 +51,66 @@ class OptOutService @Inject()(itsaStatusUpdateConnector: ITSAStatusUpdateConnect
...
+    val previousYear = currentYear.previousYear
+    val nextYear = currentYear.nextYear
+
+    OptionT(recallOptOutInitialState()).
+      map(initialState => {
+        createOptOutProposition(previousYear, currentYear, nextYear, initialState)
+      }).getOrElseF(Future.failed(new RuntimeException("Failed to recall Opt Out journey initial state")))
   }
 
+  private case class OptOutInitialState(finalisedStatus: Boolean,
```
james-king-hmrc, line 90: can be moved to OptOutService object

ian-hmrc: Done.

james-king-hmrc: may be like this
Screenshot 2024-07-12 at 15 00 15

app/services/optout/OptOutService.scala
```diff
@@ -50,40 +51,66 @@ class OptOutService @Inject()(itsaStatusUpdateConnector: ITSAStatusUpdateConnect
...
+                                ec: ExecutionContext): Future[OptOutProposition] = {
+
+    val currentYear = dateService.getCurrentTaxYear
+    val previousYear = currentYear.previousYear
+    val nextYear = currentYear.nextYear
+
+    OptionT(recallOptOutInitialState()).
+      map(initialState => {
+        createOptOutProposition(previousYear, currentYear, nextYear, initialState)
+      }).getOrElseF(Future.failed(new RuntimeException("Failed to recall Opt Out journey initial state")))
```
james-king-hmrc, line 87: also this reads better for me
Screenshot 2024-07-12 at 15 04 57

app/services/optout/OptOutService.scala
james-king-hmrc: may be like this
Screenshot 2024-07-12 at 15 00 15

joesheridan (approved):

PR #2410, opened 2024-07-18
MISUV-7891 | Refactor Credits and Refunds

owen-gilfellon:
Refactored existing Credits and Refund code to make minimal model
Implemented method in connector to call backend, using HTTPClient v2
Created abstract HttpParser for use with HttpClient v2, implemented subclass for new credits model
Created builders for testing Credits and Refunds models
Merged agent + individual integration tests
Removed .gets and exceptions in view
Fixed display of "credit-and-refund.no-credit" in view (balanceDetails is a required field in the API response, so would never display)


app/connectors/FinancialDetailsConnector.scala
```diff
@@ -90,6 +98,31 @@ class FinancialDetailsConnector @Inject()(val http: HttpClient,
 
   }
 
+  def getCreditsAndRefund(taxYear: Int, nino: String)
```
avelx, line 101: * we should use TaxYear type in all of the cases were possible;
* dateFrom and dateTo should be accessible as attributes from taxYear value

app/models/core/ResponseModel.scala
```diff
@@ -0,0 +1,88 @@
...
+
+import play.api.Logger
+import play.api.http.Status.{INTERNAL_SERVER_ERROR, NOT_FOUND, OK}
+import play.api.libs.json.{Format, JsSuccess, JsValue}
+import uk.gov.hmrc.http.{HttpReads, HttpResponse}
+
+import scala.util.Try
+
+
+object ResponseModel {
```
avelx, line 27: these models/types specifically related to CreditModel read functionality, if we can add relevant prefix to all of the name, as with the current naming I am under impression that these are generic types we can use everythere

owen-gilfellon: These are generic types that can be reused. It's subclassed in CreditsModel:

`implicit val reads: CreditsResponseReads = new CreditsResponseReads

  class CreditsResponseReads extends AResponseReads[CreditsModel] {
    implicit val format: Format[CreditsModel] = CreditsModel.format
  }`

It definitely works with HttpClientV2, haven't tried it with V1 - The Reads for ClaimToAdjust could be refactored to use it pretty simply as that's already on V2. It should hopefully come in useful when we start doing the work to move over to V2 throughout, and cut down on code duplication.

app/models/creditsandrefunds/CreditsModel.scala
```diff
@@ -0,0 +1,50 @@
...
+
+  implicit val reads: CreditsResponseReads = new CreditsResponseReads
+  class CreditsResponseReads extends AResponseReads[CreditsModel] {
+    implicit val format: Format[CreditsModel] = CreditsModel.format
+  }
+}
+
+case class Transaction(transactionType: CreditType,
+                       amount: BigDecimal,
+                       taxYear: Option[String],
```
avelx, line 42: we are consuming taxYear as Integer in most of the cases + we have control over this in our BE (not directly used from relevant REST API),  I think we can convert this field to Integer

test/testConstants/FinancialDetailsTestConstants.scala
```diff
@@ -1543,58 +1542,60 @@ object FinancialDetailsTestConstants {
...
 }
 
-case class ANewCreditAndRefundViewModel(model: CreditAndRefundViewModel = CreditAndRefundViewModel(List(), None)) {
+case class ANewCreditAndRefundModel(model: CreditsModel = CreditsModel(0.0, 0.0, Nil)) {
 
-  def withAvailableCredit(availableCredit: BigDecimal): ANewCreditAndRefundViewModel = {
-    val balanceDetails = model.balanceDetails.getOrElse(BalanceDetails(0.0, 0.0, 0.0, None, None, None, None, None))
-      .copy(availableCredit = Some(availableCredit))
-    ANewCreditAndRefundViewModel(model.copy(model.creditCharges, Some(balanceDetails)))
+  private def getTaxYear(localDate: LocalDate): Int = {
```
avelx, line 1547: yet another generic conversion we can create for TaxYear type: LocalDate => TaxYear

avelx (approved):
approved with comments

PR #2419, opened 2024-07-23
MISUV-7891 | Tidying based on PR comments

owen-gilfellon:
- Use TaxYear type instead of Int / String
- Parse TaxYear directly from JSON, with validation

### DIFF
--- a/app/models/incomeSourceDetails/TaxYear.scala
+++ b/app/models/incomeSourceDetails/TaxYear.scala
@@ -16,6 +16,7 @@
 
 package models.incomeSourceDetails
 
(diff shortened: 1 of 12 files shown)


app/models/incomeSourceDetails/TaxYear.scala
```diff
@@ -88,6 +98,14 @@ object TaxYear {
     }
   }
 
+  def getTaxYear(localDate: LocalDate): TaxYear = {
```
avelx, line 101: ideally as implicit, so we can convert LocalDate to TaxYear, but this approach is fine as well;

avelx (approved):
approved with comments

PR #2427, opened 2024-07-26
MISUV-7930  Update - Success message page after POAs are updated ('Overdue charges' if user increases POAs after making payment)

danielmaltz1:


app/controllers/claimToAdjustPoa/PoaAdjustedController.scala
```diff
@@ -66,12 +69,17 @@ class PaymentsOnAccountAdjustedController @Inject()(val authorisedFunctions: Aut
     }
   }
 
-  private def setJourneyCompletedFlag(isAgent: Boolean, poa: PaymentOnAccountViewModel)(implicit user: MtdItUser[_]): Future[Result] = {
+  private def setJourneyCompletedFlag(isAgent: Boolean, poa: PaymentOnAccountViewModel, session: PoAAmendmentData)(implicit user: MtdItUser[_]): Future[Result] = {
```
simondrugan16, line 72: Not part of this ticket but I feel like this method shouldn't be delivering the action here

simondrugan16:
Are there test scenarios for both scenarios (overdue charges and no overdue charges) in UTs/ITs?

test/views/claimToAdjustPoa/PoaAdjustedViewSpec.scala
```diff
@@ -102,12 +108,18 @@ class PaymentsOnAccountAdjustedViewSpec extends TestSupport{
...
             msgs("claimToAdjustPoa.success.afterLinkText")
-        document.getElementById("p3").getElementById("link").attr("href") shouldBe taxYearSummaryUrl(true)
+        document.getElementById("p3").getElementById("TYSLink").attr("href") shouldBe taxYearSummaryUrl(true)
       }
-      "render the final paragraph" in new Setup(isAgent = true) {
+      "render the final paragraph when no overdue charges" in new Setup(isAgent = true) {
         document.getElementById("p4").text shouldBe
           msgs("claimToAdjustPoa.success.check") + " " + msgs("claimToAdjustPoa.success.whatYouOwe") + " " + msgs("claimToAdjustPoa.success.forUpcomingCharges")
-        document.getElementById("p4").getElementById("link").attr("href") shouldBe whatYouOweUrl(true)
+        document.getElementById("p4").getElementById("WYOLink").attr("href") shouldBe whatYouOweUrl(true)
```
simondrugan16, line 116: Could we have a view spec test to specifically test that the section is missing when the condition is false

app/controllers/claimToAdjustPoa/PoaAdjustedController.scala
danielmaltz1: Because of the EitherT stuff there isn't a neat way to move the view call back into show without ruining the point of putting this logic in another method. Have renamed the method

simondrugan16 (approved):

PR #2428, opened 2024-07-26
MISUV-7810-ctaService-getCH-fetch-poa-charge-not-head

simondrugan16:
amended the getChargeHistory method in ClaimToAdjustHelper to fetch only POA charge histories, rather than just grabbing the item at the top of the list

### DIFF
--- a/it/test/testConstants/IncomeSourceIntegrationTestConstants.scala
+++ b/it/test/testConstants/IncomeSourceIntegrationTestConstants.scala
@@ -1138,7 +1138,7 @@ object IncomeSourceIntegrationTestConstants {
         "taxYear" -> "2019",
         "documentId" -> documentId,
         "documentDate" -> "2018-02-14",
-        "documentDescription" -> "ITSA - POA 1",
+        "documentDescription" -> "ITSA- POA 1",
         "totalAmount" -> amount,
         "reversalDate" -> "2019-02-14",
         "reversalReason" -> "Customer Request",
--- a/it/test/testConstants/IncomeSourceIntegrationTestConstants.scala
+++ b/it/test/testConstants/IncomeSourceIntegrationTestConstants.scala
@@ -315,7 +315,7 @@ object IncomeSourceIntegrationTestConstants {
       Json.obj(
         "taxYear" -> taxYear,
         "transactionId" -> "1040000123",
-        "documentDescription" -> "ITSA- POA 1",
+        "documentDescription" -> poa1Description,
         "outstandingAmount" -> outstandingAmount,
         "originalAmount" -> originalAmount,
         "documentDate" -> "2018-03-29"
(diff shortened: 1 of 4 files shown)


test/services/claimToAdjust/ClaimToAdjustServiceSpec.scala
```diff
@@ -440,5 +440,64 @@ class ClaimToAdjustServiceSpec extends TestSupport with MockFinancialDetailsConn
...
+        chargeHistoryDetails = Some(chargeHistoriesNoPoaAtHead)
+      ))
+
+      setupMockGetFinancialDetails(taxYear, testNino)(financialDetailsWithUnpaidPoAs(taxYear))
+
+      val f = fixture(LocalDate.of(taxYear, 8, 27))
+
+      val result = f.testClaimToAdjustService.getPoaViewModelWithAdjustmentReason(testUserNino)
+
+      whenReady(result) {
```
avelx, line 496: we are using option 2 across project as its more concise:
https://www.scalatest.org/scaladoc/3.2.7/org/scalatest/concurrent/Futures.html
Please avoid using whenReady pattern and use futureValue instead

it/test/testConstants/IncomeSourceIntegrationTestConstants.scala
```diff
@@ -1138,7 +1138,7 @@ object IncomeSourceIntegrationTestConstants {
         "taxYear" -> "2019",
         "documentId" -> documentId,
         "documentDate" -> "2018-02-14",
-        "documentDescription" -> "ITSA - POA 1",
+        "documentDescription" -> "ITSA- POA 1",
```
avelx, line 1141: Please define/use next constants in the test scope:

test/services/claimToAdjust/ClaimToAdjustServiceSpec.scala
```diff
@@ -340,22 +329,21 @@ class ClaimToAdjustServiceSpec extends TestSupport with MockFinancialDetailsConn
       ))
 
       val f = fixture(LocalDate.of(taxYear, 8, 27))
-      val result = f.testClaimToAdjustService.getAmendablePoaViewModel(testUserNino)
+      val futureResult = f.testClaimToAdjustService.getAmendablePoaViewModel(testUserNino)
+
+      val result = futureResult.futureValue
+
+      result.isLeft shouldBe true
+      result.left.exists(_.getMessage == "Failed to retrieve non-crystallised financial details") shouldBe true
```
avelx, line 337: if this can be done by use of next scala test trait: https://www.scalatest.org/user_guide/using_EitherValues
So we can compare to exact value in the Exception like: 

`result.left.value.getMessage shouldBe "Failed to retrieve non-crystallised financial details"`

avelx (approved):
approved with comment

PR #2432, opened 2024-07-30
MISUV-8047 | Investigate using global error handler

owen-gilfellon:
- Agent recover block in `BaseFrontendController` threw exceptions, Individual recover block in `AuthenticationPredicate` consumed and displayed error page. Both now throw exceptions, to be caught by global error handler.
- Agent and individuals showed different messages on error due to above. Changed global error messaging in `messages.*` to display "Sorry, there is a problem with the service - GOV.UK”.
- In order to differentiate between Individuals and Agents on the error page (where no auth is performed, so cannot retrieve Affinity Group), could wrap exceptions caught by recover blocks above.
-  Recover block in individuals is currently after SessionTimeout predicate - should be moved earlier to cover all functionality?
- Without consuming exceptions in `AuthenticationPredicate`, it revealed some incorrectly configured tests - mocks were updated in `TaxSummaryControllerSpec`, `CreditSummaryControllerSpec`, `ManageIncomeSourceDetailsControllerSpec`, `TaxControllerSpec`.
- Some test file would pass when ran with the rest of the suite, fail when ran individually. Added missing disableAllSwitches before failing tests (`NavBarPredicate` was getting called and failing due to missing "Origin").
- Added some specific, rather than generic, exceptions, to aid with testing
- Added it tests for `CreditsAndRefundsController` checking for redirect to error page
- Removed explicit calls to error handler from `CreditsAndRefundsController`
- Test in `CreditSummaryControllerSpec` is failing now mocked correctly - not sure what it is meant to be testing?

### DIFF
--- a/app/config/ItvcErrorHandler.scala
+++ b/app/config/ItvcErrorHandler.scala
@@ -16,22 +16,31 @@
 
 package config
 
+import play.api.Logging
 import play.api.i18n.{I18nSupport, MessagesApi}
 import play.api.mvc.Results.InternalServerError
-import play.api.mvc.{Request, Result}
+import play.api.mvc.{Request, RequestHeader, Result}
 import play.twirl.api.Html
 import uk.gov.hmrc.play.bootstrap.frontend.http.FrontendErrorHandler
 import views.html.errorPages.templates.ErrorTemplate
 
 import javax.inject.Inject
+import scala.concurrent.Future
 
 trait ShowInternalServerError {
   def showInternalServerError()(implicit request: Request[_]): Result
 }
 
 class ItvcErrorHandler @Inject()(val errorTemplate: ErrorTemplate,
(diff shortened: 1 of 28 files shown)


app/config/ItvcErrorHandler.scala
```diff
@@ -16,21 +16,51 @@
...
+import scala.concurrent.Future
+import scala.language.implicitConversions
 
 trait ShowInternalServerError {
   def showInternalServerError()(implicit request: Request[_]): Result
 }
 
 class ItvcErrorHandler @Inject()(val errorTemplate: ErrorTemplate,
-                                 val messagesApi: MessagesApi) extends FrontendErrorHandler with I18nSupport with ShowInternalServerError {
+                                 val messagesApi: MessagesApi) extends FrontendErrorHandler with Logging with I18nSupport with ShowInternalServerError {
```
avelx, line 37: I would assume its worth to mark global error handler as @Singleton ?

PR #2435, opened 2024-07-30
MISUV-7973 Payment, credit and refund history tax year font size should match V&C service pages

tioadesola1:
Added a short line in the view page to change the font size to 16.5 for the tax year credits

### DIFF
--- a/app/views/PaymentHistory.scala.html
+++ b/app/views/PaymentHistory.scala.html
@@ -88,7 +88,7 @@ btaNavPartial: Option[Html] = None, origin: Option[String] = None)(implicit requ
                     </a>
                     @if(payment.isCredit) {
                     <br>
-                    <a class="govuk-text" style="font-size: 16.5px;">
+                    <a class="govuk-body-s">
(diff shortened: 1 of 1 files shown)


app/views/PaymentHistory.scala.html
```diff
@@ -88,7 +88,7 @@
                     </a>
                     @if(payment.isCredit) {
                     <br>
-                    <a class="govuk-text">
+                    <a class="govuk-text" style="font-size: 16.5px;">
```
danielmaltz1, line 91: Instead of setting the font size specifically, we usually use gov classes. This text in the prototype uses "govuk-body-s", so I think it's better if you just replace the class with this, and leave out the style

danielmaltz1: This is just so that, if the gov standard design updates, our pages will also automatically update

tioadesola1: Made those changes now.

danielmaltz1 (approved):

PR #2438, opened 2024-08-01
Misuv 8031 fix opt out flakey tests

ian-hmrc:


it/test/repositories/OptOutSessionDataRepositoryISpec.scala
```diff
@@ -51,7 +51,7 @@ class OptOutSessionDataRepositoryISpec extends ComponentSpecBase with ScalaFutur
           journeyType = OptOutJourney.Name,
           optOutSessionData = Some(expectedOptOutSessionData))
 
-        repository.set(expectedSessionData)
+        repository.set(expectedSessionData).futureValue
```
avelx, line 54: I think we are using await for this kind of situation if we need blocking

ian-hmrc: Done

james-king-hmrc (approved):

PR #2444, opened 2024-08-02
Misuv 7993 log error as json

james-king-hmrc:
log error as json

### DIFF
--- a/app/connectors/optout/ITSAStatusUpdateConnector.scala
+++ b/app/connectors/optout/ITSAStatusUpdateConnector.scala
@@ -54,9 +54,7 @@ class ITSAStatusUpdateConnector @Inject()(val http: HttpClient, val appConfig: F
     ).map { response =>
       response.status match {
         case Status.NO_CONTENT => OptOutUpdateResponseSuccess()
-        case Status.NOT_FOUND =>
-          log.error(s"response status: ${response.status}")
-          OptOutUpdateResponseFailure.notFoundFailure(buildRequestUrlWith(taxableEntityId))
+
         case _ =>
           response.json.validate[OptOutUpdateResponseFailure].fold(
             invalid => {
@@ -65,6 +63,9 @@ class ITSAStatusUpdateConnector @Inject()(val http: HttpClient, val appConfig: F
             },
             valid => {
               log.error(s"response status: ${response.status}")
+              if(response.status == Status.NOT_FOUND) {
+                log.error(s"Not found error from itsa-status update response, error $valid")
+              }
               valid
             }
           )
--- a/app/connectors/optout/ITSAStatusUpdateConnector.scala
+++ b/app/connectors/optout/ITSAStatusUpdateConnector.scala
@@ -59,13 +59,15 @@ class ITSAStatusUpdateConnector @Inject()(val http: HttpClient, val appConfig: F
           response.json.validate[OptOutUpdateResponseFailure].fold(
             invalid => {
               log.error(s"Json validation error parsing itsa-status update response, error $invalid")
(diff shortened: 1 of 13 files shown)


ian-hmrc:
Could we write some Json tests for OptOutUpdateResponseFailure as we've just done for the success responses?

app/connectors/optout/OptOutUpdateRequestModel.scala
```diff
@@ -17,30 +17,27 @@
...
+    def defaultFailure(): OptOutUpdateResponseFailure =
+      OptOutUpdateResponseFailure(
         List(ErrorItem("INTERNAL_SERVER_ERROR", "Request failed due to unknown error"))
       )
 
-    def notFoundFailure(correlationId: String = "unknown", url: String): OptOutUpdateResponseFailure =
-      OptOutUpdateResponseFailure(correlationId,
-        INTERNAL_SERVER_ERROR,
+    def notFoundFailure(url: String): OptOutUpdateResponseFailure =
+      OptOutUpdateResponseFailure(
```
ian-hmrc, line 40: Would it be possible to parse the actual errors from the backend, rather than assuming we know the response? The different error responses all have different reason strings, which we could just pick out.

app/connectors/optout/ITSAStatusUpdateConnector.scala
```diff
@@ -57,16 +58,17 @@ class ITSAStatusUpdateConnector @Inject()(val http: HttpClient, val appConfig: F
...
         case Status.NOT_FOUND =>
           log.error(s"response status: ${response.status}")
-          OptOutUpdateResponseFailure.notFoundFailure(correlationId, buildRequestUrlWith(taxableEntityId))
+          OptOutUpdateResponseFailure.notFoundFailure(buildRequestUrlWith(taxableEntityId))
         case _ =>
           response.json.validate[OptOutUpdateResponseFailure].fold(
             invalid => {
-              log.error(s"Json validation error parsing itsa-status update response, error $invalid")
-              OptOutUpdateResponseFailure.defaultFailure(correlationId)
+              val invalidJsonString = Json.prettyPrint(JsError.toJson(invalid))
```
ian-hmrc, line 65: Hmm... if the json fails to validate, could it also fail to parse as Json and so not be possible to pretty-print it? Would it be safer here to just log the whole response body?

app/connectors/optout/OptOutUpdateRequestModel.scala
james-king-hmrc: actually this was a tactical solution before the BE end-point implementation and it remained after that, also 'notFoundFailure' meaning the service wasn't reachable so there would be no message to parse or pick

james-king-hmrc: done

app/connectors/optout/ITSAStatusUpdateConnector.scala
james-king-hmrc: done

ian-hmrc: Cool. :)

app/connectors/optout/OptOutUpdateRequestModel.scala
ian-hmrc: I think we've got there with this one now, right?

app/connectors/optout/ITSAStatusUpdateConnector.scala
ian-hmrc: Looks like this is done. :)

ian-hmrc (approved):
LGTM

PR #2447, opened 2024-08-05
MISUV-8052-investigate-httpClientV2

simondrugan16:
investigation for implementing http client v2 + separate HTTP reads


test/connectors/ChargeHistoryConnectorSpec.scala
```diff
@@ -18,29 +18,31 @@ package connectors
...
+import testConstants.ChargeHistoryTestConstants.{testChargeHistoryErrorModelParsing, testValidChargeHistoryModel}
 import testUtils.TestSupport
-import uk.gov.hmrc.http.HttpResponse
 import uk.gov.hmrc.play.bootstrap.config.ServicesConfig
 
+import java.time.LocalDate
 import scala.concurrent.Future
 
-class ChargeHistoryConnectorSpec extends TestSupport with MockHttp with MockAuditingService {
+class ChargeHistoryConnectorSpec extends TestSupport with MockHttp with MockHttpV2 with MockAuditingService {
```
avelx, line 34: I think we don't need MockHttp to be exposed in the scope of this spec? as we can use only one or anther ?

owen-gilfellon:
I was hoping `AResponseReads` might be useful in cutting down on boiler plate when implementing the HttpReads, as per CreditsModel 🙂

`object CreditsModel {

  implicit val format: OFormat[CreditsModel] = Json.format[CreditsModel ]

  implicit val reads: CreditsResponseReads = new CreditsResponseReads
  class CreditsResponseReads extends AResponseReads[CreditsModel] {
    implicit val format: Format[CreditsModel] = CreditsModel.format
  }
}`

simondrugan16:
> I was hoping `AResponseReads` might be useful in cutting down on boiler plate when implementing the HttpReads, as per CreditsModel 🙂
> 
> ```
> object CreditsModel {
> 
>   implicit val format: OFormat[CreditsModel] = Json.format[CreditsModel ]
> 
>   implicit val reads: CreditsResponseReads = new CreditsResponseReads
> 
>   class CreditsResponseReads extends AResponseReads[CreditsModel] {
>     implicit val format: Format[CreditsModel] = CreditsModel.format
>   }
> }
> ```

I like the idea of using AResponseReads but I didn't really want to mess with the logic since the current read logic is a bit different to AResponseReads (e.g: current logic treats NOT_FOUND and FORBIDDEN in the same way 😵‍💫 ). I think it would be a good idea to revisit and simplify these connectors in the future, but probably outside the scope of this ticket? What do you think

owen-gilfellon:
> I like the idea of using AResponseReads but I didn't really want to mess with the logic since the current read logic is a bit different to AResponseReads (e.g: current logic treats NOT_FOUND and FORBIDDEN in the same way 😵‍💫 ). I think it would be a good idea to revisit and simplify these connectors in the future, but probably outside the scope of this ticket? What do you think

Yeah, I was wondering about that - it returns an empty model if FORBIDDEN, seemed strange! Fair enough, if it's going to cause issues

avelx (approved):

PR #2448, opened 2024-08-05
MISUV-6918 Replace nested set mongo key calls with set mongo data in IncomeSourceEndDateController

tioadesola1:
Created several methods to handle multiple inputs to the mongo database and created unit tests to test the functionality of the newly created methods.


app/controllers/manageBusinesses/cease/IncomeSourceEndDateController.scala
```diff
@@ -136,28 +136,53 @@ class IncomeSourceEndDateController @Inject()(val authorisedFunctions: FrontendA
       )
   }
 
-  private def handleValidatedInput(validatedInput: DateFormElement,
-                                   incomeSourceType: IncomeSourceType,
-                                   incomeSourceIdMaybe: Option[IncomeSourceId],
-                                   redirectAction: Call)(
-                                    implicit headerCarrier: HeaderCarrier) = {
+//  private def handleValidatedInput(validatedInput: DateFormElement,
```
simondrugan16, line 139: please remove commented code

app/controllers/manageBusinesses/cease/IncomeSourceEndDateController.scala
```diff
@@ -136,28 +136,53 @@ class IncomeSourceEndDateController @Inject()(val authorisedFunctions: FrontendA
...
+        val dataToSet = Map(
+          CeaseIncomeSourceData.dateCeasedField -> validatedInput.date.toString,
+          CeaseIncomeSourceData.incomeSourceIdField -> incomeSourceId.value
+        )
 
+        sessionService.setMultipleMongoData(dataToSet, JourneyType(Cease, incomeSourceType)).flatMap {
+          case Right(_) => Future.successful(result)
           case Left(_) => Future.failed(new Error(
-            s"Failed to set end date value in session storage. incomeSourceType: $incomeSourceType, incomeSourceType: $incomeSourceType"))
+            s"Failed to set data in session storage. incomeSourceType: $incomeSourceType."))
```
simondrugan16, line 185: It'd be good to have more more detail about what has gone wrong in this case, like adding an error log which would include the date ceased

simondrugan16: Income source Id would be good too but i'm not sure if this is personal identifying information or not. Better to be safe than sorry and leave it out :D

app/repositories/UIJourneySessionDataRepository.scala
```diff
@@ -116,6 +116,16 @@ class UIJourneySessionDataRepository @Inject()(
       update = Document("$set" -> Document(key -> value))
     ).toFuture()
   }
+//  def updateMultipleData(data: UIJourneySessionData, fieldsToUpdate: Map[String, String])
```
simondrugan16, line 119: Remove commented code

app/services/SessionService.scala
```diff
@@ -100,6 +100,31 @@ class SessionService @Inject()(
       uiJourneySessionDataRepository.set(uiJourneySessionData)
   }
 
+  def setMultipleMongoData(kv: Map[String, String], journeyType: JourneyType)
+                 (implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Either[Throwable, Boolean]] = {
+    val uiJourneySessionData = UIJourneySessionData(hc.sessionId.get.value, journeyType.toString)
+    kv.map { kv_ =>
+        journeyType.operation match {
+          case Add => (kv_._2, AddIncomeSourceData.getJSONKeyPath(kv_._1))
```
simondrugan16, line 108: try swapping these round, like

`          case Add => (AddIncomeSourceData.getJSONKeyPath(kv_._1), kv_._2)
`
I think it's a bit confusing to call it kv (key value) but then change the order to value, key

app/services/SessionService.scala
```diff
@@ -100,6 +100,31 @@ class SessionService @Inject()(
       uiJourneySessionDataRepository.set(uiJourneySessionData)
   }
 
+  def setMultipleMongoData(kv: Map[String, String], journeyType: JourneyType)
+                 (implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Either[Throwable, Boolean]] = {
+    val uiJourneySessionData = UIJourneySessionData(hc.sessionId.get.value, journeyType.toString)
+    kv.map { kv_ =>
```
simondrugan16, line 106: is kv_ specific naming convention for tuples? i think there might be a more descriptive name for this

app/services/SessionService.scala
```diff
@@ -100,6 +100,31 @@ class SessionService @Inject()(
       uiJourneySessionDataRepository.set(uiJourneySessionData)
   }
 
+  def setMultipleMongoData(kv: Map[String, String], journeyType: JourneyType)
```
simondrugan16, line 103: i think this method would return Future(Right(false)) if kv is empty. Would this be ok?

test/controllers/manageBusinesses/cease/IncomeSourceEndDateControllerSpec.scala
```diff
@@ -556,22 +556,32 @@ class IncomeSourceEndDateControllerSpec extends TestSupport with MockAuthenticat
...
           "called .submit" in {
             setupMockCreateSession(true)
-            setupMockSetSessionKeyMongo(CeaseIncomeSourceData.dateCeasedField)(Right(true))
-            setupMockSetSessionKeyMongo(CeaseIncomeSourceData.incomeSourceIdField)(Left(new Exception()))
+            val testMap: Map[String, String] = Map(
+              CeaseIncomeSourceData.dateCeasedField -> "2022-08-27",
+              CeaseIncomeSourceData.incomeSourceIdField -> testSelfEmploymentId
+            )
+            setupMockSetMultipleMongoData(testMap)(Right(true))
+            setupMockSetMultipleMongoData(testMap)(Left(new Exception()))
```
simondrugan16, line 574: this second mock will overwrite the first mock. in the code I think an exception would be returned if setMultipleMongoData can't set either field, so you can just remove

`            setupMockSetMultipleMongoData(testMap)(Right(true))
`

test/services/SessionServiceSpec.scala
```diff
@@ -83,6 +83,23 @@ class SessionServiceSpec extends TestSupport with MockUIJourneySessionDataReposi
...
+        }
+        "return a future left exception" in {
+          mockRepositoryUpdateDataFailure()
+          val testMap: Map[String, String] = Map("key1" -> "value1", "key2" -> "value2")
+          val result: Either[Throwable, Boolean] = TestSessionService.setMultipleMongoData(testMap,
+            JourneyType(Add, SelfEmployment))(headerCarrier, ec).futureValue
+          result.toString shouldBe Left(new Exception("Mongo Save data operation was not acknowledged")).toString
+        }
+      }
+
```
simondrugan16, line 102: I thought of some additional tests you can do :)

mockRepositoryUpdateDataFailure will return isAcknowledged = false no matter what, but we want to test that the whole operation fails if the first updateData passes, and only the second one fails. We can make new mocks in this test file to do this, by specifying the values in the mocks (eg: ArgumentMatchers.eq("key1")

app/services/SessionService.scala
```diff
@@ -100,6 +100,31 @@ class SessionService @Inject()(
...
+                case false => Left(new Exception("Mongo Save data operation was not acknowledged"))
+              }
+            )
+          case Left(ex) =>
+            Future.successful(Left(ex))
+        }
+      }
+  }
+
   def setMongoKey(key: String, value: String, journeyType: JourneyType)
```
avelx, line 128: this method will become a special case of setMultipleMongoData, as you can wire it to use setMultipleMongoData;

avelx:
We would need to try next approach in the next iteration:

add next method to UIJourneySessionDataRepository => 
and use it instead setMultipleMongoData:
`  

def updateMultipleData(data: UIJourneySessionData, kv: Map[String, String]): Future[UpdateResult] = {
    collection.updateMany(
      filter = dataFilter(data),
      update = { kv.map(e =>  Document("$set" -> Document(e._1 -> e._2)) ) }.toList
    ).toFuture()
  }`

test/services/SessionServiceSpec.scala
```diff
@@ -74,6 +75,16 @@ class SessionServiceSpec extends TestSupport with MockUIJourneySessionDataReposi
           result shouldBe true
         }
       }
+
+      "setMultipleMongoData method" should {
+        "return a future boolean value" in {
+          updateMultipleData()
+          val result: Either[Throwable, Boolean] = TestSessionService.setMultipleMongoData(Map("key" -> "value"),
```
simondrugan16, line 82: worth testing an error case + setMultipleMongoData with multiple values in it?

test/controllers/incomeSources/cease/IncomeSourceEndDateControllerSpec.scala
```diff
@@ -412,7 +412,9 @@ class IncomeSourceEndDateControllerSpec extends TestSupport with MockAuthenticat
       }
     }
   }
+
   "IncomeSourceEndDateController submit/submitChange/submitAgent/submitChangeAgent" should {
```
simondrugan16, line 416: why are there no changes to manage businesses tests in this PR? i thought they would have broken

app/repositories/UIJourneySessionDataRepository.scala
```diff
@@ -116,7 +117,12 @@ class UIJourneySessionDataRepository @Inject()(
       update = Document("$set" -> Document(key -> value))
     ).toFuture()
   }
-
+  def updateMultipleData(data: UIJourneySessionData, keyAndValue: Map[String, String]): Future[UpdateResult] = {
```
simondrugan16, line 120: could do with an autoformatting

app/controllers/incomeSources/cease/IncomeSourceEndDateController.scala
```diff
@@ -217,22 +217,15 @@ class IncomeSourceEndDateController @Inject()(val authorisedFunctions: FrontendA
...
-
+        val mongoSetValues = Map(
+          CeaseIncomeSourceData.dateCeasedField -> validatedInput.date.toString,
+          CeaseIncomeSourceData.incomeSourceIdField -> incomeSourceId.value
+        )
+        sessionService.setMultipleMongoData(mongoSetValues, JourneyType(Cease, incomeSourceType)).flatMap {
+          case Right(_) => Future.successful(result)
           case Left(_) => Future.failed(new Error(
-            s"Failed to set end date value in session storage. incomeSourceType: $incomeSourceType, incomeSourceType: $incomeSourceType"))
+            s"Failed to set data in session storage. incomeSourceType: $incomeSourceType."))
```
simondrugan16, line 227: Should we log incomeSourceId or is it PII?

test/controllers/incomeSources/cease/IncomeSourceEndDateControllerSpec.scala
avelx: surprisingly there are no broken tests in manageBusiness, might be missing UTs??

app/controllers/incomeSources/cease/IncomeSourceEndDateController.scala
avelx: I don't think we need it, there will be incomeSourceHashId present in the Kibana log

test/services/SessionServiceSpec.scala
avelx: done

simondrugan16 (approved):

PR #2454, opened 2024-08-07
Misuv 8129 tidy up opt out test

ian-hmrc:


test/services/optout/OptOutServiceSpec.scala
```diff
@@ -109,45 +128,68 @@ class OptOutServiceSpec extends UnitSpec
...
-        isReset shouldBe true
-        noneIntent.isDefined shouldBe false
       }
 
       f.futureValue shouldBe Succeeded
     }
   }
 
+  "OptOutService.recallOptOutInitialState" should {
+    "retrieve the intent year only" in {
```
james-king-hmrc, line 147: not clear what the intent of the test is and
can't see how the asserts in the test match the test name

james-king-hmrc, line 147: should be renamed something like "retrieve saved itsa-status"

PR #2455, opened 2024-08-07
MISUV-6919: Investigation: Fix error message highlighting

ayubAz:


app/viewmodels/govuk/ErrorSummaryFluency.scala
```diff
@@ -0,0 +1,62 @@
...
+import play.api.data.Form
+import play.api.i18n.Messages
+import uk.gov.hmrc.govukfrontend.views.viewmodels.content.{Content, Text}
+import uk.gov.hmrc.govukfrontend.views.viewmodels.errorsummary.{ErrorLink, ErrorSummary}
+
+object errorsummary extends ErrorSummaryFluency
+
+trait ErrorSummaryFluency {
+
+  object ErrorSummaryViewModel {
```
avelx, line 28: this is not best practice to hide singleton within trait, I would think singleton should be outside of the trait??

avelx: I think if you would analyse this file, its seems like custom package object implementation, there is no need for these complexities , at least in our code base; in case you keen to use package object that fine

PR #2456, opened 2024-08-07
MISUV-8136-choose-year-ui

james-king-hmrc:
opt-in multi-year-choice

### DIFF
--- a/app/models/optin/ChooseTaxYearViewModel.scala
+++ b/app/models/optin/ChooseTaxYearViewModel.scala
@@ -1,0 +1,24 @@
+/*
+ * Copyright 2024 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.optin
+
+import models.incomeSourceDetails.TaxYear
+
+case class ChooseTaxYearViewModel(
+                                   availableOptInTaxYear: List[TaxYear] =
+                                   List(TaxYear.forYearEnd(2023), TaxYear.forYearEnd(2024)),
+                                   cancelURL: String, isAgent: Boolean)
--- a/app/controllers/optin/OptInChooseYearController.scala
+++ b/app/controllers/optin/OptInChooseYearController.scala
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2024 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package controllers.optin
+
+import auth.{FrontendAuthorisedFunctions, MtdItUser}
+import config.featureswitch.FeatureSwitching
+import config.{AgentItvcErrorHandler, FrontendAppConfig, ItvcErrorHandler}
+import controllers.agent.predicates.ClientConfirmedController
+import forms.optin.ChooseTaxYearForm
+import models.optin.ChooseTaxYearViewModel
+import play.api.Logger
+import play.api.i18n.I18nSupport
+import play.api.mvc.{Action, AnyContent, MessagesControllerComponents, Result}
+import utils.AuthenticatorPredicate
+import views.html.optin.ChooseTaxYearView
+
+import javax.inject.Inject
+import scala.concurrent.{ExecutionContext, Future}
+
+class OptInChooseYearController @Inject()(val view: ChooseTaxYearView,
+                                          val authorisedFunctions: FrontendAuthorisedFunctions,
+                                          val auth: AuthenticatorPredicate)
+                                         (implicit val appConfig: FrontendAppConfig,
+                                          mcc: MessagesControllerComponents,
+                                          val ec: ExecutionContext,
+                                          val itvcErrorHandler: ItvcErrorHandler,
+                                          val itvcErrorHandlerAgent: AgentItvcErrorHandler)
+  extends ClientConfirmedController with FeatureSwitching with I18nSupport {
+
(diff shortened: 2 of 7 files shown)


app/controllers/optIn/ChooseYearController.scala
```diff
@@ -0,0 +1,95 @@
...
+        Logger("application").error(s"request failed :: $ex")
+        errorHandler(isAgent).showInternalServerError()
+    }
+  }
+
+  def show(isAgent: Boolean = false): Action[AnyContent] = auth.authenticatedAction(isAgent) {
+    implicit user =>
+      withRecover(isAgent) {
+        Future.successful(Ok(view(ChooseTaxYearForm(optInService.availableOptInTaxYear().map(_.toString)), viewModel(isAgent))))
+      }
```
ian-hmrc, line 61: I can see several calls to `optInService.availableOptInTaxYear()` which will probably necessitate several API calls. I presume this is addressed with the next PR.

app/controllers/optIn/ChooseYearController.scala
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  private def redirectToCheckpointPage(isAgent: Boolean): Result = {
+    val nextPage = controllers.optIn.routes.CheckYourAnswersController.show(isAgent)
+    Logger("application").info(s"redirecting to : $nextPage")
+    Redirect(nextPage)
+  }
+
+  private def cancelUrl(isAgent: Boolean): String = {
+    routes.ChooseYearController.show(isAgent).url //todo change this to the correct url
```
ian-hmrc, line 88: Is it possible to put a placeholder in for now?

app/models/optin/ChooseTaxYearViewModel.scala
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package models.optin
+
+import models.incomeSourceDetails.TaxYear
+
+case class ChooseTaxYearViewModel(
+                                   availableOptInTaxYear: List[TaxYear],
+                                   cancelURL: String, isAgent: Boolean
```
ian-hmrc, line 23: Might be clearer with each argument on its own line?

app/models/optin/OptInContextData.scala
```diff
@@ -0,0 +1,49 @@
...
+
+case class OptInContextData(crystallisationStatus: Boolean,
+                             previousYearITSAStatus: String,
+                             currentYearITSAStatus: String,
+                             nextYearITSAStatus: String)
+
+object OptInContextData {
+  implicit val format: OFormat[OptInContextData] = Json.format[OptInContextData]
+
+  def statusToString(status: ITSAStatus): String =
```
ian-hmrc, line 31: This is all duplicated from Opt Out. We should look to unify this.

test/forms/optIn/ChooseTaxYearFormSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+  val invalidFormData: Map[String, String] = Map(ChooseTaxYearForm.choiceField -> "")
+
+  when(mockMessages.apply(ChooseTaxYearForm.noResponseErrorMessageKey)
+  ).thenReturn(errorMessage)
+
+  "ChooseTaxYearForm" when {
+
+    "bind with a valid response" should {
+      "contain the response - true" in {
+        val completedForm: Form[ChooseTaxYearForm] = ChooseTaxYearForm(choices).bind(validFormData)
```
ian-hmrc, line 44: I have a slight preference for making `validFormData` a function here, so it can be clearly shown that it is being passed `taxYear`. It's then easy to see when reading the test, that the input is verified in the assertion after, and you don't need to look elsewhere to know that the assertion matches.

app/controllers/optIn/ChooseYearController.scala
james-king-hmrc: yes we fetch saved from repository if exists or make api call

app/controllers/optIn/ChooseYearController.scala
james-king-hmrc: this is the placeholder, but what do you suggest?

app/models/optin/OptInContextData.scala
james-king-hmrc: I'll do this in the service pr

app/controllers/optIn/ChooseYearController.scala
ian-hmrc: OK, but if we put repo calls where we currently have API calls, we'll be making several repo calls, so effectively it will have the same issue.

I probably should have said that I meant that the APIs should be refactored so that the data can be shared to where it is required without having to make repeated calls for one page view. I can see that that's been done in the Service PR, so I'll mark this as done.

app/controllers/optIn/ChooseYearController.scala
ian-hmrc: I meant a placeholder for the page it should go to, a bit like you've done with the Check Your Answers page. That way it won't get accidentally left behind when that page become available.

app/models/optin/OptInContextData.scala
ian-hmrc: Cool

test/forms/optIn/ChooseTaxYearFormSpec.scala
ian-hmrc: Cool

app/controllers/optIn/ChooseYearController.scala
james-king-hmrc: Okay, so what is the page it should go to?

ian-hmrc: Ah... good question. I thought the ticket or prototype would say. I suspect it's the new Your Reporting Frequency page, but we can check at stand-up.

james-king-hmrc: done

app/services/optIn/OptInService.scala
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  private def getITSAStatusesFrom(currentYear: TaxYear)(implicit user: MtdItUser[_],
+                                                         hc: HeaderCarrier,
+                                                         ec: ExecutionContext): Future[Map[TaxYear, ITSAStatus]] = {
+    itsaStatusService.getStatusTillAvailableFutureYears(currentYear.previousYear).map(_.view.mapValues(_.status).toMap.withDefaultValue(ITSAStatus.NoStatus))
+    //todo is passing currentYear.previousYear correct here?
+  }
+
+  private def createOptInProposition( currentYear: TaxYear,
```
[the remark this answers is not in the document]
james-king-hmrc: not clear how 'createOptInProposition' can be moved to 'OptOutProposition' one says OptIn the other says OptOut? are you thinking of merging the two concepts? in that case OptOutProposition should be renamed to a more general name, right?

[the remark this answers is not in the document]
ian-hmrc: Sorry. That was my mistake. I meant `OptInProposition`. We should keep the 2 concepts separate.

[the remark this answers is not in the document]
james-king-hmrc: done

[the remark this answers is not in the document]
ian-hmrc: Cool

ian-hmrc (approved):
LGTM

PR #2464, opened 2024-08-16
MISUV-8145 | Merge Agent and Individual Auth

owen-gilfellon:
Simplified the logic for Auth (e.g. authorising on` HMRC-AS-AGENT` enrolment to ensure ARN, rather than empty enrolment and manually checking).

Used composite predicates (`uk.gov.hmrc.auth.core.authorise.Predicate`) to check for Agent / Individual on same auth call.

Split up some of the logic into different actions to add comprehensibility.

More abstract "user" type introduced - EnroledUser. Would aid in supporting agents without confirmed clients, but needs more work in that direction.

IV-uplift URL in production is incorrect - it looks like the service base url appears twice, and the URLs passed as parameters need to be URL encoded (corrected here).

Identifier key for `testConstants.BaseTestConstants#arnEnrolment` ("ARN") looks to be incorrect. ARN is parsed from enrolments here: `controllers.predicates.IncomeTaxAgentUser#agentReferenceNumber`, where "AgentReferenceNumber" is used. Updated here, but no impact in the tests...

### DIFF
--- a/app/auth/authV2/actions/SessionTimeoutPredicateV2.scala
+++ b/app/auth/authV2/actions/SessionTimeoutPredicateV2.scala
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2023 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package auth.authV2.actions
+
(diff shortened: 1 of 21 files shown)


app/auth/authV2/actions/SessionTimeoutPredicateV2.scala
```diff
@@ -0,0 +1,49 @@
...
+package auth.authV2.actions
+
+import play.api.Logger
+import play.api.mvc.Results.Redirect
+import play.api.mvc._
+import uk.gov.hmrc.http.SessionKeys
+
+import javax.inject.{Inject, Singleton}
+import scala.concurrent.{ExecutionContext, Future}
+
```
avelx, line 26: I would assume we need a new version for SessionTimeoutPredicate (V2?) in order to break dependency from BaseController and make it properly composable with the rest of the actions we have?

avelx:
@owen-gilfellon  - did you have a chance to run ATs on this PR? as I am getting few failures

app/auth/authV2/actions/SessionTimeoutPredicateV2.scala
owen-gilfellon: Yes, where-ever I needed to modify and existing class, I made a duplicate of it.

PR #2471, opened 2024-08-19
MISUV-2467: PT Jenkins job: Automate switching on all features in the frontend before running PTs

ayubAz:


app/testOnly/controllers/FeatureSwitchController.scala
```diff
@@ -106,4 +106,14 @@ class FeatureSwitchController @Inject()(featureSwitchView: FeatureSwitchView,
...
+  def enableAll(): Action[AnyContent] = Action.async { implicit user =>
+    for {
+      featureSwitches <- featureSwitchService.getAll
+      _ <- Future.sequence(
+        featureSwitches.map(featureSwitch =>
+          featureSwitchService.set(featureSwitch.name, enabled = true)
+        )
+      )
+    } yield Redirect(testOnly.controllers.routes.FeatureSwitchController.show)
+  }
```
avelx, line 118: @ayubAz  - could you pls add basic tests for this change?

app/testOnly/controllers/FeatureSwitchController.scala
```diff
@@ -106,4 +106,14 @@ class FeatureSwitchController @Inject()(featureSwitchView: FeatureSwitchView,
 
   }
 
+  def enableAll(): Action[AnyContent] = Action.async { implicit user =>
```
simondrugan16, line 109: What environment do these work on? It would be good to preserve the original state of the FSs if in Staging/QA so that they don't just constantly get turned on?

avelx: testOnly endpoint available only in staging (and QA?) as far as I know

avelx (approved):
approved with comments

app/testOnly/controllers/FeatureSwitchController.scala
simondrugan16: Why not local? What environment do PTs run on? This would be most important I think

avelx: In local we have "FS read from mongo" feature disabled and this for good; as I understand scope of this story its just to apply this where we have "FS read from mongo" set to on

PR #2472, opened 2024-08-19
MISUV-8107_Charge_definition_Add_Late_payment_interest_on_charge_name_definition_on_relevant_charge_pages

tioadesola1:
Charge definition add late payment interest on charge name definition on relevant charge pages


app/views/ChargeSummary.scala.html
```diff
@@ -160,6 +185,7 @@ <h1 class="govuk-heading-xl">
     @{
         viewModel.documentType match {
             case Poa1Charge | Poa2Charge => getPoaExplanationContent
+            case TRMNewCharge if viewModel.latePaymentInterestCharge => getBalancingPaymentsExplanationContent
```
simondrugan16, line 188: Where did you find that TRMNewCharge relates to lpi charge?

tioadesola1: I printed the document type and it stated it was a TRMNewCharge

simondrugan16: So you printed off a document type? Do you know that this relates to all lpi charges? We had a meeting earlier which showed that a lot of the information is not well known, and lots of the test data is wrong. I don't think that TRMNewCharge directly relates to lpi charge. Could you post in the team-lunar slack channel to check on this to ensure the right thing is done?

tioadesola1: Yea, so I spoke with Pavel and Amna on this matter at the time and they were made aware of this. But they may have just forgotten about the issue so I'll send a message.

avelx: @simondrugan16 , as discussed during the meeting earlier, **_there number of way we can get specific charge type_**,
above combination seem to work well for "Late payment interest on balancing payment" based on the data we have, but this might not a preferred way / approach to do this;

tioadesola1: Ideally it should be the specific charge type based on the page. As that addresses all charge types, is it really okay to use that instead or should it be refactored?

tioadesola1: With the first case, it's very clear what page would be accessed. I guess also for readability purposes would it not be better to use the specific charge type?

simondrugan16: Owen has used the document type BalancingCharge [here](https://github.com/hmrc/income-tax-view-change-frontend/blob/c5da1e75d74b96d777f380c0c57fa1d654938ccc/app/views/ChargeSummary.scala.html#L175) to refer to the balancing charge. Then this ticket could add an extra check for if it is an LPI charge. I think this would fit in better with the code that Owen merged yesterday? (Then follow the same pattern for POAs?)

avelx: @tioadesola1  - I would assume these comments addressed ?

tioadesola1: I wrote a couple comments on this issue. It seems the charge types are still different for different pages. As you said before the way we are getting the charge type doesn't seem to be a problem as you stated multiple ways of getting it. In the code base however.

Balancing Payments page = TRMNewCharge
1st Late Balancing Payments page = BalancingCharge
2nd Late Balancing Payments page = TRMNewCharge

So on the tax summary year page there are two links for late balancing payments. One uses BalancingCharge and the other uses TRMNewCharge. Do we still want to keep it like this or should we discuss making the charge types more consistent?

avelx: I would assume we can use approach you decided to go with, reset of the changes can be covered by https://jira.tools.tax.service.gov.uk/browse/MISUV-8215, I would think that story is still applicable

app/views/ChargeSummary.scala.html
```diff
@@ -80,14 +108,33 @@
...
+        }
+    </div>
+}
+
 @getExplanationContent = @{viewModel.documentType match {
-        case Poa1Charge | Poa2Charge    => paymentOnAccountExplanation
-        case BalancingCharge            => balancingChargeExplanation
+        case Poa1Charge | Poa2Charge if !viewModel.latePaymentInterestCharge => paymentOnAccountExplanation
+        case Poa1Charge | Poa2Charge if viewModel.latePaymentInterestCharge => getLpiPoaExplanationContent
+        case TRMNewCharge if !viewModel.latePaymentInterestCharge => balancingChargeExplanation //TRMNewCharge used instead of Balancing charge
```
avelx, line 129: I would assume these comments is something we need to discuss / address?

avelx (dismissed):

app/views/ChargeSummary.scala.html
tioadesola1: Yes, we should raise another ticket to possibly tidy up the charge types to be consistent over each type. But it doesn't pose a problem at the time being

avelx (approved):

PR #2475, opened 2024-08-21
Misuv 8167 improve opt out journey repository

ian-hmrc:
Introduce OptOutSessionDataRepository to encapsulate saving and restoring of Opt Out journey artefacts. This moves responsibilities out of OptOutService and makes a cleaner separation of the responsibilities of the service and storing/retrieving data.


app/repositories/OptOutSessionDataRepository.scala
```diff
@@ -0,0 +1,118 @@
...
+import repositories.OptOutContextData.{statusToString, stringToStatus}
+import services.optout.OptOutProposition
+import services.optout.OptOutProposition.createOptOutProposition
+import uk.gov.hmrc.http.HeaderCarrier
+import utils.OptOutJourney
+
+import javax.inject.Inject
+import scala.concurrent.{ExecutionContext, Future}
+
+class OptOutSessionDataRepository @Inject()(val repository: UIJourneySessionDataRepository) {
```
avelx, line 34: * this seems to me a service rather than repository;
* we are moving away from UIJourneySessionDataRepository in favour of SensitiveUIJourneySessionDataRepository;

ian-hmrc: It feels more like a repository to me as its role is to store and retrieve journey data. It's a wrapping layer around the generic journey repository that provides Opt Out specific functionality, and reduces the size of the large `OptOutService` class.

We have no sensitive data in this journey, so it seems unnecessary to migrate to storing this as sensitive data.

avelx: @ian-hmrc - as I understand UIJourneySessionDataRepository will be decommission, you really need to have your own repository in this case

ian-hmrc: OK. Thanks @avelx. I'll raise a tech-debt ticket to migrate to the Sensitive repo.

ian-hmrc: I've created [MISUV-8248](https://jira.tools.tax.service.gov.uk/browse/MISUV-8248) for this.

james-king-hmrc, line 34: this is all about OptOut specific operations over the repository so I think 
- it should be renamed to OptOutRepositoryService
- and moved to the services.optout package

app/repositories/OptOutSessionDataRepository.scala
```diff
@@ -0,0 +1,118 @@
...
+case class OptOutContextData(currentYear: String,
+                             crystallisationStatus: Boolean,
+                             previousYearITSAStatus: String,
+                             currentYearITSAStatus: String,
+                             nextYearITSAStatus: String)
+
+object OptOutContextData {
+  implicit val format: OFormat[OptOutContextData] = Json.format[OptOutContextData]
+
+  def statusToString(status: ITSAStatus): String =
```
james-king-hmrc, line 100: I think 
- should pull changes from main 
- and move 'statusToString', 'stringToStatus' to the new OptOutRepositoryService object 
- and reuse these for both OptOut and OptIn
- and delete repositories.ITSAStatusRepositorySupport.scala from main (which I don't like being there)

app/repositories/OptOutSessionDataRepository.scala
```diff
@@ -0,0 +1,118 @@
...
+import uk.gov.hmrc.http.HeaderCarrier
+import utils.OptOutJourney
+
+import javax.inject.Inject
+import scala.concurrent.{ExecutionContext, Future}
+
+class OptOutSessionDataRepository @Inject()(val repository: UIJourneySessionDataRepository) {
+
+  def saveIntent(intent: TaxYear)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Boolean] = {
+    OptionT(repository.get(hc.sessionId.get.value, OptOutJourney.Name)).
```
james-king-hmrc, line 37: should it not add optOutSessionData if missing?

james-king-hmrc (dismissed):
approved with 3 comments

app/repositories/OptOutSessionDataRepository.scala
ian-hmrc: Good spot! I missed that one when merging main. I've pushed up a change that uses the centralised functions in ITSAStatusRepositorySupport. I think this is the right thing to do as ITSAStatusRepositorySupport is then shared between both Opt Out and Opt In.

app/repositories/OptOutSessionDataRepository.scala
ian-hmrc: I see it differently. I think modules like this class should be named after the role the play in supporting the modules that use them; in this case the `OptOutService`. `OptOutService` requires journey information to be stored and retrieved, and that's what a `Repository` does.

I could be persuaded to contract the name to `OptOutRepository` or maybe `OptOutSessionRepository`, but I don't think adding `Service` helps much. I find that `Service` is over-used and is anodyne rather than  domain centric, and because of this lack of specificity,  can end up with the file being a bit of a junk-draw for random bits of code.

app/repositories/OptOutSessionDataRepository.scala
ian-hmrc: That's a good question. The only way that the user should get to this point is by going through the journey, and by doing that this will be setup at the start, so, under normal circumstances, it should never be missing.

I suppose some user could jump directly to this page, but that would show the error page (correctly, I think) as it would not be able to pull back the `OptOutProposition` if the session data were missing.

The only other scenario I can think of is if the user pauses between the setup at the start, and coming to this page. The service wipes old sessions from mongo, but the user would also be logged-out and having to sign-in again, so they would have to re-start the journey.

Are there other scenarios you can think of that might be problematic?

I think that it's simple in it's current form and covers all of the scenarios I can think of. If we don't need to do this for a particular reason, then I'd rather keep this as it is.

james-king-hmrc (approved):

PR #2476, opened 2024-08-21
MISUV-8223-optin-check-your-answer

james-king-hmrc:
optin-check-your-answer


app/connectors/optout/ITSAStatusUpdateConnector.scala
```diff
@@ -74,4 +75,36 @@ class ITSAStatusUpdateConnector @Inject()(val http: HttpClient, val appConfig: F
       }
     }
   }
+
+  /* todo this call to replace requestOptOutForTaxYear */
+  def makeITSAStatusUpdate(taxYear: TaxYear, taxableEntityId: String, updateReason: String)
```
avelx, line 80: We are in process of migrating to HttpClientV2, please use new version of http client as this one is deprecated already

app/models/incomeSourceDetails/TaxYear.scala
```diff
@@ -48,6 +48,11 @@ case class TaxYear(startYear: Int, endYear: Int) {
 
   def isBefore(taxYear: TaxYear): Boolean = this.startYear < taxYear.startYear
 
+  def isNextTaxYear(implicit dateService: DateServiceInterface): Boolean = {
+    val currentTaxYearEnd = dateService.getCurrentTaxYearEnd
+    endYear > currentTaxYearEnd && (endYear - currentTaxYearEnd == 1)
+  }
```
ian-hmrc, line 54: This method has the Primitive Obsession code smell, and could be implemented more simply using domain concepts with the code `dateService.getCurrentTaxYear.nextYear == this`

Also, passing in Services like `DateService` into value types couples them to those services and reduces the context independence of those value types. So I think this code should be moved off `TaxYear`. Using the above simplification, we could either move the whole expression to the call site, or another alternative might be to split it and add a `nextTaxYear()` method to `DateService` and use something like `dateService.nextTaxYear == xxx` at the call site?

app/controllers/optIn/CheckYourAnswersController.scala
```diff
@@ -55,9 +60,28 @@ class CheckYourAnswersController @Inject()(val optInService: OptInService,
...
-          Ok(s"Check your answers (${taxYear.get}) page! in MISUV-8006")
-        }
+
+        val result = for {
+          taxYear <- OptionT(optInService.fetchSavedChosenTaxYear())
+          cancelURL = ReportingFrequencyPageController.show(isAgent).url
+          intentIsNextYear = taxYear.isNextTaxYear(dateService)
+        } yield Ok(view(MultiYearCheckYourAnswersViewModel(taxYear, isAgent, cancelURL, intentIsNextYear)))
+
+        result.getOrElse(errorHandler(isAgent).showInternalServerError())
```
ian-hmrc, line 70: Could we move this logic out of the controller and instead build a view model in the Service? This will make testing this logic much simpler.

app/services/optIn/OptInService.scala
```diff
@@ -75,6 +76,17 @@ class OptInService @Inject()(itsaStatusUpdateConnector: ITSAStatusUpdateConnecto
     }
   }
 
+  def makeOptInCall()(implicit user: MtdItUser[_],
+                      hc: HeaderCarrier,
+                      ec: ExecutionContext): Future[ITSAStatusUpdateResponse] = {
+
+    fetchSavedChosenTaxYear() flatMap {
+      case Some(intentTaxYear) => itsaStatusUpdateConnector.makeITSAStatusUpdate(taxYear = intentTaxYear, user.nino, optInUpdateReason)
```
ian-hmrc, line 84: When we move Opt Out over to use the same code on the connector, could we make 2 wrapper methods on the connector, maybe `optOut()` and `optIn()` to keep the update reasons closer to the connector and stop them spilling out into the service?

app/models/incomeSourceDetails/TaxYear.scala
james-king-hmrc: done

app/controllers/optIn/CheckYourAnswersController.scala
james-king-hmrc: will implement this comment in this PR https://github.com/hmrc/income-tax-view-change-frontend/pull/2479/files as the same code is touched, added your comment there

app/services/optIn/OptInService.scala
james-king-hmrc: created sub-task for this: https://jira.tools.tax.service.gov.uk/browse/MISUV-8280

ian-hmrc (approved):
LGTM

PR #2479, opened 2024-08-27
MISUV-8224-optin-with-previously-submitted-updates

james-king-hmrc:
optin-with-previously-submitted-updates


app/controllers/optIn/CheckYourAnswersController.scala
```diff
@@ -62,10 +63,15 @@ class CheckYourAnswersController @Inject()(val view: CheckYourAnswersView,
       withRecover(isAgent) {
 
         val result = for {
-          taxYear <- OptionT(optInService.fetchSavedChosenTaxYear())
+          intentTaxYear <- OptionT(optInService.fetchSavedChosenTaxYear())
           cancelURL = ReportingFrequencyPageController.show(isAgent).url
-          intentIsNextYear = taxYear.isNextTaxYear(dateService)
-        } yield Ok(view(MultiYearCheckYourAnswersViewModel(taxYear, isAgent, cancelURL, intentIsNextYear)))
+          intentIsNextYear = intentTaxYear.isNextTaxYear(dateService)
```
james-king-hmrc, line 68: Ian: "Could we move this logic out of the controller and instead build a view model in the Service? This will make testing this logic much simpler."

james-king-hmrc: @ian-hmrc Done

ian-hmrc: Cool

ian-hmrc:
> Sure @ian-hmrc , but there are no stacked tickets etc at the moment, let's get this PR reviewed and we can look at other tickets next
OK, I can see you've merged main, and that's taken the number of files to review from 33 to 11. I'll have a look at that.

app/services/optIn/OptInService.scala
```diff
@@ -165,4 +169,54 @@ class OptInService @Inject()(itsaStatusUpdateConnector: ITSAStatusUpdateConnecto
...
+                                                              hc: HeaderCarrier,
+                                                              ec: ExecutionContext): Future[Option[MultiYearCheckYourAnswersViewModel]] = {
+    val result = for {
+      intentTaxYear <- OptionT(fetchSavedChosenTaxYear())
+      cancelURL = ReportingFrequencyPageController.show(isAgent).url
+      intentIsNextYear = isNextTaxYear(dateService.getCurrentTaxYear, intentTaxYear)
+      proposition <- OptionT.liftF(fetchOptInProposition())
+      quarterlyUpdatesCountForOfferedYears <- OptionT.liftF(getQuarterlyUpdatesCountForOfferedYears(proposition))
+      showPreviouslySubmittedUpdatesWarning = quarterlyUpdatesCountForOfferedYears.counts
+        .filter(v => v.taxYear == intentTaxYear).map(_.count).headOption.getOrElse(ZeroCount) > ZeroCount
```
ian-hmrc, line 208: I'm questioning if this is the right way to calculate this for Opt In? If the user is Opting In, then they should currently be reporting Annually, so shouldn't have any quarterly updates, so won't the count of quarterly updates always be zero? This might be a more subtle difference between Obligations (Updates) and Submissions. I've asked Design, but let's pick up tomorrow and see if we can make sure we've understood this correctly.

PR #2483, opened 2024-08-29
MISUV-8152 | Added content for HMRC adjustment on charge summary page

owen-gilfellon:
- Added HMRC Adjustment explantation on Charge Summary page
- Removed incorrect codes for MFA Debits in MFADebitUtils (which were unused, but confusing)
- Changed explanation content HTML to remove ids on paragraphs and bullet points, in favour of a wrapping div with id (makes tests more readable, avoids inadvertent duplicate ids in future)


avelx (approved):

app/models/financialDetails/MfaDebitUtils.scala
```diff
@@ -17,14 +17,14 @@
 package models.financialDetails
 
 object MfaDebitUtils {
-  private val MFADebitType: Map[Int, String] = Map(
-    4004 -> "ITSA PAYE Charge",
-    4005 -> "ITSA Calc Error Correction",
-    4006 -> "ITSA Manual Penalty Pre CY-4",
-    4007 -> "ITSA Misc Charge")
+  private val MFADebitType: List[String] = List(
```
avelx, line 20: If we can use "Main transaction code" instead to detect MFA charges:
> codes between 4000-4003 see below

avelx (approved):

PR #2488, opened 2024-08-30
MISUV-8221-optin-excel-sheet-test

james-king-hmrc:
optin-excel-sheet-test

### DIFF
--- a/test/services/optIn/OptInPropositionExcelSpec.scala
+++ b/test/services/optIn/OptInPropositionExcelSpec.scala
@@ -142,8 +142,7 @@ class OptInPropositionExcelSpec extends UnitSpec {
     ("A", " ", "CY+1", true, List("CY", "CY+1"), "CY+1", "A", "V"),
     ("A", "V", "CY", true, List("CY"), "CY", "V", "V"),
     ("A", "V", "CY+1", false, List("CY"), " ", "A", "V"),
-    /* todo: check this line */
-    //("A", "A", "CY", true, List("CY", "CY+1"), "CY", "V", "V"),
+    ("A", "A", "CY", true, List("CY", "CY+1"), "CY", "V", "A"),
     ("A", "A", "CY+1", true, List("CY", "CY+1"), "CY+1", "A", "V"),
     (" ", "A", "CY+1", false, List("CY+1"), "CY+1", " ", "V"),
     (" ", "M", "CY+1", false, List(), " ", " ", "M"),
--- a/test/services/optIn/OptInPropositionExcelSpec.scala
+++ b/test/services/optIn/OptInPropositionExcelSpec.scala
@@ -93,7 +93,7 @@ class OptInPropositionExcelSpec extends UnitSpec {
     val tab = "  "
 
     rowsWithoutHeaders.foreach(line => {
-      val cells = line.split(tab).take(9)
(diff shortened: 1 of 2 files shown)


test/services/optIn/OptInPropositionExcelSpec.scala
```diff
@@ -0,0 +1,222 @@
...
+
+class OptInPropositionExcelSpec extends UnitSpec {
+
+  "Parse opt out scenarios from tsv file" ignore {
+    /*  The programme has specified all required Opt Out scenarios in a large spreadsheet.
+        This code can ingest the data and convert to parameters for table based test found below.
+        This generator is currently ignored but can be reactivated to regenerate the test data if needed.
+          - To reactivate generator, replace "ignore" with "in"
+
+      Select first 9 columns of Opt Out Scenarios Spreadsheet (currently v15).
```
JordHemsworth, line 35: Opt In sheet is v5 rather than 15

james-oncloud: done

test/services/optIn/OptInPropositionExcelSpec.scala
```diff
@@ -0,0 +1,223 @@
...
+    ("V", "A", "CY", false, List("CY+1"), " ", "V", "A"),
+    ("V", "A", "CY+1", true, List("CY+1"), "CY+1", "V", "V"),
+    ("A", "M", "CY", true, List("CY"), "CY", "V", "M"),
+    ("A", "M", "CY+1", false, List("CY"), " ", "A", "M"),
+    ("A", " ", "CY", true, List("CY", "CY+1"), "CY", "V", " "),
+    ("A", " ", "CY+1", true, List("CY", "CY+1"), "CY+1", "A", "V"),
+    ("A", "V", "CY", true, List("CY"), "CY", "V", "V"),
+    ("A", "V", "CY+1", false, List("CY"), " ", "A", "V"),
+    /* todo: check this line */
+    //("A", "A", "CY", true, List("CY", "CY+1"), "CY", "V", "V"),
```
JordHemsworth, line 146: With this test, the implementation more is perhaps more geared to opt out where we send multi year to the backend whereas with Opt In we only send the one

james-king-hmrc: What's the change required here please?

james-king-hmrc: @ian-hmrc please clarify this comment, not clear on what's the required change here?
Screenshot 2024-09-03 at 10 10 54

ian-hmrc: With Opt-In, we initially had some ambiguity as to whether the front-end or the back-end would own the 'and onwards' functionality, i.e. whether we would just send the user's chosen tax year, or all the tax years from their choice and onwards.

This information is the best we have from the business on this decision, so, according to the sheet, it falls to the back-end to implement this. Let's align to this and make that row of the test pass by just sending the chosen year for now. Should we get a change of specification, then we can revisit this.

james-king-hmrc: done

JordHemsworth (approved):

PR #2493, opened 2024-09-02
Address PR bot issues

Aashvin:


conf/testOnlyDoNotUseInAppConf.routes
```diff
@@ -30,7 +30,7 @@ DELETE      /report-quarterly/income-and-expenses/view/test-only/all-schemas/pro
 
 + nocsrf
 GET         /report-quarterly/income-and-expenses/view/test-only/custom-login                         testOnly.controllers.CustomLoginController.showLogin
-GET         /hmrc-frontend/assets/hmrc-frontend-5.19.0.min.css                                        testOnly.controllers.CustomLoginController.showCss
+GET         /test-only/hmrc-frontend/assets/hmrc-frontend-5.19.0.min.css                              testOnly.controllers.CustomLoginController.showCss
```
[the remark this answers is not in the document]
avelx: testOnly route will be removed from PROD (and I think from QA), so these CSS resources will be not available; might worth to revert this change

conf/logback-test.xml
```diff
@@ -14,34 +14,13 @@
         </encoder>
     </appender>
 
-    <appender name="STDOUT_IGNORE_NETTY" class="ch.qos.logback.core.ConsoleAppender">
```
avelx, line 17: we had a story in the [past](https://jira.tools.tax.service.gov.uk/browse/MISUV-7355) to add source of the error to the logs, this might be specific to our project, but removing these (even to satisfy PR Bot warning) would cause problem with investigation

conf/testOnlyDoNotUseInAppConf.routes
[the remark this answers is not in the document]
Aashvin: If it might be needed in prod, is it worth moving this route out of the testOnly file?

conf/logback-test.xml
Aashvin: We've still got the one above named STDOUT which includes this formatting - is it necessary to include STDOUT_IGNORE_NETTY too?

conf/testOnlyDoNotUseInAppConf.routes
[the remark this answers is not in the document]
avelx: not sure what you mean, change above is to move out of prod route into testOnly, this would cause issues

conf/logback-test.xml
avelx: Please check the story above and let me know if you have question regarding implementation,
changes you proposing would cause significant issue when we would have to do investigation; even not possible to do investigation in some cases

Aashvin: Based on the story, it seems like the changes that were made in [this file of this PR](https://github.com/hmrc/income-tax-view-change-frontend/pull/2236/files#diff-e357afe958839e3c860b492456bbb3c4fe973c35d3d4d06f5e2c280f6558ff62) are to the appender defined as STDOUT on line 11 of this file. I've checked locally and the messages still include the class and line number that the error is referencing

ian-hmrc: @avelx, I can't see anything being logged via that appender. To do that it would have to be logged to `"com.ning.http.client.providers.netty"` which would look like `Logger("com.ning.http.client.providers.netty").warn("...")` in the code. All logging currently is done to `Logger("application")` which is logged via the `root` level appender to both the `FILE` and `STDOUT` appenders.

The PR that shows how we can change to log class and method names automatically would also be logged via the `root` level, and equally go to the `FILE` and `STDOUT` appenders.

If you can find a counter-example where we will loose logging, then I'd be keen to hear about it.

conf/testOnlyDoNotUseInAppConf.routes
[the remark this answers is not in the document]
Aashvin: It's currently in the testOnly routes file. If it might be used in production, I think it might be more appropriate to move it into one of the other routes files
But taking a look at the method that the route calls, it calls the dynamic stub connector, so I'm not sure that it would be used in prod

avelx (approved):
approved with comments, 
we assume that there is no Impact
(to be cross check in staging after deployment)

avelx (approved):

PR #2496, opened 2024-09-03
Title of Balancing payment changed to lower case

tioadesola1:
Changed to lowercase b instead of B


tioadesola1:
DO NOT MERGE - Waiting on confirmation for the Welsh translation

tioadesola1:
Spoke with Amna, she mentioned everything should be correct as it was copied straight from the ticket so the welsh should be fine. It's okay to merge

avelx (approved):

PR #2506, opened 2024-09-09
WIP - implement case trait for Reason

Antony100:


app/models/chargeHistory/ChargeHistoryModel.scala
```diff
@@ -51,3 +48,16 @@ case class ChargeHistoryModel(taxYear: String,
 object ChargeHistoryModel {
   implicit val format: Format[ChargeHistoryModel] = Json.format[ChargeHistoryModel]
 }
+
```
simondrugan16, line 51: please refactor to a separate file

app/models/chargeHistory/ChargeHistoryModel.scala
```diff
@@ -51,3 +48,16 @@ case class ChargeHistoryModel(taxYear: String,
...
+sealed trait Reason {
+  val asString: String
+}
+case object Amend extends Reason {
+  override val asString: String = "amend"
```
simondrugan16, line 56: perhaps rename asString to partialKey, since this is how it's being used?

app/models/chargeHistory/ChargeHistoryModel.scala
```diff
@@ -31,14 +30,12 @@ case class ChargeHistoryModel(taxYear: String,
...
     else
       reversalReason match {
-        case "amended return" => "amend"
-        case "Customer Request" => "request"
-        case error =>
-          Logger("application").error(s"Missing or non-matching history reason: $error found")
-          "unrecognisedReason"
+        case "amended return" => Amend.asString
+        case "Customer Request" => Request.asString
+        case _ => "unrecognisedReason"
```
simondrugan16, line 38: This doesn't translate to a message key... The type signature for this method should be Option[String] where the case _ scenario would be None. Then wherever this is used we should do some validation, and if the value of this val is None then we should do the logging and maybe return an exception somewhere since this should not happen.

app/models/chargeHistory/ChargeHistoryModel.scala
simondrugan16: (to be more descriptive)

PR #2507, opened 2024-09-09
MISUV-8323-opt-in-completed

james-king-hmrc:
opt-in-completed


app/controllers/optIn/OptInCompletedController.scala
```diff
@@ -17,20 +17,23 @@
...
 import utils.AuthenticatorPredicate
+import views.html.optIn.{OptInCompletedCurrentYearView, OptInCompletedNextYearView}
 
-import java.time.LocalDateTime
 import javax.inject.Inject
 import scala.concurrent.{ExecutionContext, Future}
 
-class OptInCompletedController @Inject()(val optInService: OptInService,
+class OptInCompletedController @Inject()(val currentYearView: OptInCompletedCurrentYearView,
+                                         val nextYearView: OptInCompletedNextYearView,
```
ian-hmrc, line 35: I'm unsure if having separate views for the current and next year will be the best way forward here. There's a number of further scenarios for this page where variations would cut across both of these. We're still working out what needs to be done and making the tickets. Let me catch up with the BA and try to get a clear idea of what needs to be done, and then let's talk this through and find the best way to deliver the design.

ian-hmrc (approved):
LGTM

PR #2508, opened 2024-09-10
MISUV-8336-add-user-session-api-fs

simondrugan16:
added FS functionality


conf/application.conf
```diff
@@ -179,6 +179,7 @@ feature-switch {
   enable-opt-out = false
   enable-income-sources-new-journey = false
   enable-time-machine = false
+  enable-user-session-API = true
```
avelx, line 182: can we use different name for this FS? something more expressive like, enable-session-data-storage or similar

app/config/featureswitch/UserSessionApiFS.scala
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package config.featureswitch
+
+import config.FrontendAppConfig
+
+trait UserSessionApiFS {
```
avelx, line 21: no need to create trait here, why we need this trait btw?

simondrugan16: just how we did it for the other feature switches, i can remove it

conf/application.conf
simondrugan16: ya sure, this was just the one originally on the ticket

it/test/helpers/agent/ComponentSpecBase.scala
```diff
@@ -93,6 +93,7 @@ trait ComponentSpecBase extends TestSuite with CustomMatchers
     "microservice.services.contact-frontend.port" -> mockPort,
     "feature-switches.read-from-mongo" -> "false",
     "feature-switch.enable-time-machine" -> "false",
+    "feature-switch.enable-user-session-API" -> "true",
```
avelx, line 96: if you can fix naming here as well: 
_feature-switch.enable-session-data-storage_

simondrugan16: yep thanks

avelx (approved):
approved

PR #2511, opened 2024-09-12
MISUV-8314: Display new reconciliation debit charges on 'Tax year summary' page along with red banner for accruing interest

ayubAz:


app/models/financialDetails/ReviewAndReconcileDebitUtils.scala
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.financialDetails
+
+object ReviewAndReconcileDebitUtils {
+
+  lazy val poaOneMainTransaction = "4911"
```
avelx, line 21: I was under impression that we always going to have both codes in the 1553 response,
so if one present, second must be there as well, might be something we need to confirm with Amna K;

app/models/financialDetails/DocumentDetail.scala
```diff
@@ -221,9 +221,14 @@ case class DocumentDetail(taxYear: Int,
 
 case class DocumentDetailWithDueDate(documentDetail: DocumentDetail, dueDate: Option[LocalDate],
                                      isLatePaymentInterest: Boolean = false, dunningLock: Boolean = false,
-                                     codingOutEnabled: Boolean = false, isMFADebit: Boolean = false)(implicit val dateService: DateServiceInterface) {
+                                     codingOutEnabled: Boolean = false, isMFADebit: Boolean = false,
```
avelx, line 224: we should really avoid use of boolean flags, as I understand only one of these will be "true", so these need to be modelled in the different way

conf/application.conf
```diff
@@ -180,6 +180,7 @@ feature-switch {
   enable-income-sources-new-journey = false
   enable-time-machine = false
   enable-session-data-storage = true
+  enable-review-and-reconcile-debits = true
```
avelx, line 183: in confluence document, you set this to false for some reason;
if we can synch these pls => https://confluence.tools.tax.service.gov.uk/pages/viewpage.action?spaceKey=MISUV&title=Feature+Switches+State

app/views/TaxYearSummary.scala.html
```diff
@@ -231,6 +246,9 @@ <h2 class="govuk-heading-m">@messages(header)</h2>
                             @documentDetailWithDueDate.documentDetail.getBalancingChargeDueDate(codingOutEnabled).map(_.toLongDateShort).getOrElse(messages("tax-year-summary.na"))
                          </td>
                          <td class="govuk-table__cell govuk-table__cell--numeric">@documentDetailWithDueDate.documentDetail.originalAmount.toCurrencyString</td>
+                    } else if(documentDetailWithDueDate.isReviewAndReconcileDebit) {
+                         <td class="govuk-table__cell">@documentDetailWithDueDate.dueDate.map(_.toLongDateShort).getOrElse(messages("tax-year-summary.na"))</td>
```
avelx, line 250: we need to double check this with BA, as setting default value is not part of the requirements

avelx (approved):
approved with comments: tech debt story created: https://jira.tools.tax.service.gov.uk/browse/MISUV-8394

PR #2513, opened 2024-09-12
Misuv 7902 ac3: WIP =>

avelx:


app/enums/ReversalReason.scala
```diff
@@ -0,0 +1,43 @@
...
+case object  AdjustmentReversalReason extends ReversalReason {
+  override val value: String = "adjustment"
+}
+
+case object  CustomerRequestReason extends ReversalReason {
+  override val value: String = "request"
+}
+
+case object  UnknownReversalReason extends ReversalReason {
+  override val value: String = "unrecognisedReason"
```
simondrugan16, line 42: unrecognisedReason is used in the context of calling a key in the messages file. I can't see unrecognisedReason as a key in any messages file anywhere?

avelx: this PR is refactoring, and not related to what is message file

PR #2527, opened 2024-09-23
Misuv 8373 single year confirmation page

tharunprasath:


app/controllers/optIn/ConfirmTaxYearController.scala
```diff
@@ -0,0 +1,89 @@
...
+import play.api.mvc._
+import services.DateService
+import services.optIn.OptInService
+import utils.AuthenticatorPredicate
+import views.html.optIn.ConfirmTaxYear
+
+import javax.inject.Inject
+import scala.concurrent.{ExecutionContext, Future}
+
+class ConfirmTaxYearController @Inject()(val view: ConfirmTaxYear,
```
james-king-hmrc, line 38: there is already OptInCompletedController why not update that class with single-year work?

app/views/optIn/ConfirmTaxYear.scala.html
```diff
@@ -0,0 +1,70 @@
+@*
+ * Copyright 2024 HM Revenue & Customs
```
james-king-hmrc, line 2: this file could be named 'OptInCompletedSingleYearView' inline with existing file name

app/controllers/optIn/ConfirmTaxYearController.scala
james-king-hmrc: or may be renamed it 'OptInCompletedSingleYearController'

james-king-hmrc: Okay, sounds like I misunderstood

JordHemsworth (approved):
looks good!

PR #2531, opened 2024-09-24
MISUV-8315 Review & Reconcile POA | Display content related to extra amount being paid through new reconciliation debit charges for both POAs on POA drill down pages

danielmaltz1:


app/controllers/ChargeSummaryController.scala
```diff
@@ -184,6 +186,23 @@ class ChargeSummaryController @Inject()(val auth: AuthenticatorPredicate,
 
   }
 
+  private def checkForPoaExtraChargeLink(chargeDetailsForTaxYear: FinancialDetailsModel, documentDetailWithDueDate: DocumentDetailWithDueDate, isAgent: Boolean): Option[String] = {
+    val desiredMainTransaction: String = documentDetailWithDueDate.documentDetail.getDocType match { //TODO: Replace with new mainTransaction method
```
avelx, line 190: I would assume you are still working on this segment of code?

danielmaltz1: This code will need to be refactored once Owen's ticket is merged and we can use mainTransaction to catergorise charges. I didn't realise I'd left the TODO in there though, will remove it

danielmaltz1: This will need to be done for a couple other tickets as well, and I'm happy to do it, just didn't want it to hold everything up

avelx: I think in this case we need to merge @owen-gilfellon  story first, as this is a bit redundant manipulation

PR #2534, opened 2024-09-25
MISUV-8379 Review & Reconcile POA - Feature 2 | Build - Charge drill down page for new reconciliation debit charges

ayubAz:


app/controllers/ChargeSummaryController.scala
```diff
@@ -154,28 +167,44 @@ class ChargeSummaryController @Inject()(val auth: AuthenticatorPredicate,
...
+                isAgent = isAgent,
+                isMFADebit = isMFADebit,
+                isReviewAndReconcilePoaOneDebit = isReviewAndReconcilePoaOneDebit,
+                isReviewAndReconcilePoaTwoDebit = isReviewAndReconcilePoaTwoDebit,
+                documentType = documentDetailWithDueDate.documentDetail.getDocType,
+                adjustmentHistory = chargeHistoryService.getAdjustmentHistory(chargeHistory, documentDetailWithDueDate.documentDetail),
+                poaOneChargeUrl = poaOneChargeUrl,
+                poaTwoChargeUrl = poaTwoChargeUrl,
+              )
+              mandatoryViewDataPresent(isLatePaymentCharge, viewModel.documentDetailWithDueDate) match {
```
avelx, line 203: this is a second time validation, looks a bit strange, I guess there is no reason to create viewModel in the first place if data is not valid, this is a bit confusing to read, I think you can look at this as the lowest priority

app/models/chargeSummary/ChargeSummaryViewModel.scala
```diff
@@ -41,8 +41,12 @@ case class ChargeSummaryViewModel(
                                    origin: Option[String] = None,
                                    gatewayPage: Option[GatewayPage] = None,
                                    isMFADebit: Boolean,
+                                   isReviewAndReconcilePoaOneDebit: Boolean,
```
avelx, line 44: we really need to create an enum something like DebitType or similar, as there can be only one debit selected for this model: Mfa, ReviewAndReconcilePoaOne or ReviewAndReconcilePoaTwo

app/views/ChargeSummary.scala.html
```diff
@@ -324,7 +339,7 @@ <h1 class="govuk-heading-xl">
         }
     }
 
-    @if(viewModel.noIsAgentAndRemainingToPayWithNoCodingOutEnabledAndIsPayeSelfAssessment){
+    @if(viewModel.noIsAgentAndRemainingToPayWithNoCodingOutEnabledAndIsPayeSelfAssessment && !viewModel.isReviewAndReconcileDebit){
```
avelx, line 342: I guess we can resolve all logic on the level of view model? so extra condition added can be moved within noIsAgentAndRemainingToPayWithNoCodingOutEnabledAndIsPayeSelfAssessment property ?

avelx (approved):
approved with comments : tech debt stories to be created

PR #2536, opened 2024-09-25
Review & Reconcile POA 'Overdue' tag to be displayed on 'What you Owe' and 'Tax Year Summary -> Charges tab' for new reconciliation debit charges after 31 Jan (due date)

danielmaltz1:


avelx (approved):

avelx:
As discussed tech debt story required to cover added properties for DocumentDetail model;

PR #2541, opened 2024-09-27
MISUV-8274 Upgrade ITSAStatusUpdateConnector to HttpClientV2 and some cleanup

mywyau:
At the time didn't realise the change was just a single connector, can split out the changes into a smaller pr when we can get the tests and connector in a good spot. Also was doing some cleaning up which has made the pr larger than it should be...

**Summary of changes**

- clean up build.sbt
- move out dependencies to own object
- move out code coverage to help clean up build.sbt
- add run all tests script
- refactor AddressLookupConnector to new http client
- refactor BtaNavBarPartialConnector to new http client
- refactor ITSAStatusUpdateConnector to new http client
- clean up refactored connector unit tests
- try to clean up some syntax warnings mainly ()
- fix run.sh sbt command to run testonly custom login route locally
- updated all the copyrights

tried to reduce the number of exclusions from code coverage but then coverage drops below minimum, will leave for now.


mywyau:
So can hit the ReportingFrequencyPageController which is using the optInService but need to write the integration tests really. 

Screenshot 2024-09-27 at 20 59 04

ATs appear to run clean but unsure if the code in the connector is ever tested in anyway there.

PR #2552, opened 2024-09-30
MISUV-8315 Attempt 2: Display content related to extra amount being paid through new reconciliation debit charges for both POAs on POA drill down pages

danielmaltz1:


app/controllers/ChargeSummaryController.scala
```diff
@@ -222,6 +223,27 @@ class ChargeSummaryController @Inject()(val auth: AuthenticatorPredicate,
     }
   }
 
+  private def checkForPoaExtraChargeLink(chargeDetailsForTaxYear: FinancialDetailsModel, documentDetailWithDueDate: DocumentDetailWithDueDate, isAgent: Boolean)(implicit user: MtdItUser[_]): Option[String] = {
+    val chargeItem: Option[ChargeItem] = ChargeItem.tryGetChargeItem(isEnabled(CodingOut), isEnabled(ReviewAndReconcilePoa))(
+      chargeDetailsForTaxYear.financialDetails)(documentDetailWithDueDate.documentDetail)
+
+    val desiredMainTransaction: String = chargeItem match {
```
avelx, line 230: worth to move this code as a partial implementation
under ChargeType, as opposite to ChargeType.fromCode function;
something like fromChargeType etc

app/models/financialDetails/ChargeItem.scala
```diff
@@ -142,6 +142,12 @@ case class ChargeItem (
         Logger("application").error(s"Missing or non-matching charge type: $error found")
         "unknownCharge"
     }
+
+  def poaLinkForDrilldownPage: String = transactionType match {
```
avelx, line 146: naming convention => toTransactonCode ?

avelx (approved):
approve with comments: more refactoring would be required

PR #2562, opened 2024-10-03
MISUV-8477 Review & Reconcile POA - Feature 1 & 2 | Display new reconciliation credit types on 'Payment, credit and refund history' page

ayubAz:


app/models/financialDetails/TransactionType.scala
```diff
@@ -145,6 +153,9 @@ object CreditType {
   private val mfaCredit = Range.inclusive(4004, 4025)
     .filterNot(_ == 4010).filterNot(_ == 4020).map(_.toString)
     .toList
+  private val paymentOnAccountOneReviewAndReconcileCredit = "4912"
```
avelx, line 156: if we can move this up around line 99, to keep transaction codes close to each other

avelx (approved):

PR #2565, opened 2024-10-04
MISUV-8344 Filter PoA Charges if Coded Out

danielmaltz1:


app/controllers/TaxYearSummaryController.scala
```diff
@@ -149,6 +149,13 @@ class TaxYearSummaryController @Inject()(taxYearSummaryView: TaxYearSummary,
     }
   }
 
+  private val notCodedOutPoa: TaxYearSummaryChargeItem => Boolean = chargeItem => {
```
avelx, line 152: if we can move this under ChargeItem and add some tests around this logic?

app/services/WhatYouOweService.scala
```diff
@@ -43,6 +43,13 @@ class WhatYouOweService @Inject()(val financialDetailsService: FinancialDetailsS
     }
   }
 
+  val notCodedOutPoa: ChargeItem => Boolean = chargeItem => {
```
avelx, line 46: same as above, looks like a duplicate code which we can avoid by moving it in ChargeItem?

app/controllers/TaxYearSummaryController.scala
danielmaltz1: TaxYearSummaryChargeItem is a different type to ChargeItem, but I can move both of these bits of code into the relevant type

app/models/financialDetails/ChargeItem.scala
```diff
@@ -80,6 +80,13 @@ case class ChargeItem (
     subTransactionType.exists(subType => codingOutSubTypes.contains(subType))
   }
 
+  def notCodedOutPoa: Boolean = {
```
avelx, line 83: Please move this method from [ChargeItem] and [TaxYearSummaryChargeItem] to base trait => TransactionItem
here is example:
https://github.com/hmrc/income-tax-view-change-frontend/pull/2567/files

avelx (approved):

PR #2568, opened 2024-10-07
MISUV-8185 Charge drill down page | Removal of 'Make payment' button and related content

danielmaltz1:


app/views/ChargeSummary.scala.html
```diff
@@ -381,18 +357,6 @@ <h1 class="govuk-heading-xl">
         }
     }
 
-    @if(viewModel.remainingToPayWithNoCodingOutEnabledAndIsPayeSelfAssessment){
```
avelx, line 384: :recycle: - Please drop this properly from the viewModel as this is not required anymore;

avelx (approved):

PR #2569, opened 2024-10-07
Misuv 8467

Antony100:


conf/messages.cy
```diff
@@ -1112,8 +1112,6 @@ paymentHistory.paymentToHmrc                                    = Taliad roeddec
 paymentHistory.cutOver                                          = Credyd o flwyddyn dreth gynharach
 paymentHistory.mfa                                              = Credyd o ganlyniad i addasiad gan CThEF
 paymentHistory.repaymentInterest                                = Credyd o log ar ad-daliadau
-paymentHistory.POA1RR-credit                                    = Taliad cyntaf ar gyfrif: credyd o’ch Ffurflen Dreth
```
avelx, line 1115: why we drop these two lang resources ? they been added this morning

Antony100: This is the result of a pull from main. Currently checking if this is correct

avelx (approved):

avelx (approved):

PR #2574, opened 2024-10-08
MISUV-8434 - User Adjusts POA Audit event

Daveo-G:


app/services/claimToAdjustPoa/RecalculatePoaHelper.scala
```diff
@@ -44,16 +47,32 @@ trait RecalculatePoaHelper extends ClientConfirmedController with FeatureSwitchi
...
       case PoAAmendmentData(Some(poaAdjustmentReason), Some(amount), _) =>
         ctaCalculationService.recalculate(nino, poa.taxYear, amount, poaAdjustmentReason) map {
           case Left(ex) =>
             Logger("application").error(s"POA recalculation request failed: ${ex.getMessage}")
+            auditingService.extendedAudit(AdjustPaymentsOnAccountAuditModel(
+              isSuccessful = false,
+              previousPaymentOnAccountAmount = poa.totalAmountOne,
+              requestedPaymentOnAccountAmount = amount,
+              adjustmentReasonCode = poaAdjustmentReason.code,
+              adjustmentReasonDescription = Messages(poaAdjustmentReason.messagesKey),
```
avelx, line 63: as I understand you are planning to switch from English to Welsh, depending on the context?
we only use one single language in audit events;

Daveo-G: The Adjustment reason text only appears in the messages file.
I didn't want to hard code the text for the adjustment reason so decided to use the case objects in this file: https://github.com/hmrc/income-tax-view-change-frontend/blob/main/app/models/claimToAdjustPoa/SelectYourReason.scala 
All the descriptions use a key from the messages file.

avelx: * This would not work, I've just tested this locally;
* There will be switch to use Welsh if user selected Welsh, we expect all audit messages to be in English;

Screenshot 2024-10-08 at 13 41 16
Screenshot 2024-10-08 at 13 40 19

Daveo-G: Ok, let me fix that.
Looks like I need to rebase as well!!!

PR #2583, opened 2024-10-14
MISUV-8441 Content change for CY+1 Mandated CY Annual

jmountain-hmrc:


app/controllers/optIn/OptInCompletedController.scala
```diff
@@ -70,7 +69,8 @@ class OptInCompletedController @Inject()(val view: OptInCompletedView,
                   optInTaxYear = optInTaxYear,
                   showAnnualReportingAdvice = proposition.showAnnualReportingAdvice(optInTaxYear),
                   isCurrentYear = proposition.isCurrentTaxYear(optInTaxYear),
-                  optInIncludedNextYear = proposition.nextTaxYear.status == Voluntary
+                  optInIncludedNextYear = proposition.nextTaxYear.status == Voluntary,
```
mywyau, line 72: not sure if pr is ready for review since it's still in draft but wonder if it worth moving this logic to a new method in the OptInProposition case class model

something like:
```
def annualWithFollowingYearMandated(): Boolean = currentTaxYear.status == Annual && nextTaxYear.status == Mandated
```
know the logic is simple but then we can test the logic in the unit tests rather than just test the flag being set to false or true in the view model.

mywyau: Think somewhere like here: 
https://github.com/hmrc/income-tax-view-change-frontend/blob/7aaab58b6c99482299eb15a31140c94de58c1b3e/test/services/optIn/OptInPropositionSpec.scala#L143

not sure on the nesting/scope but something along these lines. The last test case isn't really a valid scenario but yeah can choose to add it or not.
 
then in the OptInPropositionSpec we can add test cases:
```
    ".annualWithFollowingYearMandated()" when {

      "current year ITSA status is Annual and next year itsa status is Mandated" should {
        "return true" in {

          val result =
            createOptInProposition(
              currentYear = currentTaxYear,
              currentYearItsaStatus = Annual,
              nextYearItsaStatus = Mandated
            ).annualWithFollowingYearMandated()

          result shouldBe true
        }
      }

      "current year ITSA status is Annual and next year itsa status is any other status" should {
        "return false" in {

          val result =
            createOptInProposition(
              currentYear = currentTaxYear,
              currentYearItsaStatus = Annual,
              nextYearItsaStatus = Voluntary
            ).annualWithFollowingYearMandated()

          result shouldBe false
        }
      }

      "current year ITSA status is any other status and next year itsa status is Mandated" should {
        "return false" in {

          val result =
            createOptInProposition(
              currentYear = currentTaxYear,
              currentYearItsaStatus = Voluntary,
              nextYearItsaStatus = Mandated
            ).annualWithFollowingYearMandated()

          result shouldBe false
        }
      }
    }
```

app/views/optIn/OptInCompletedView.scala.html
```diff
@@ -96,30 +96,48 @@ <h2 id="optin-completed-view-heading" class="govuk-heading-m" >@messages("optin.
...
+            }
+            </p>
+            <ul class="govuk-list govuk-list--bullet">
+                <li>@messages("optin.completedOptIn.futureReporting.p1")</li>
+                <li>@messages("optin.completedOptIn.futureReporting.p2")</li>
+            </ul>
+        } else {
+            <h2 class="govuk-heading-m" >@messages("optin.completedOptIn.nymandated.futureReporting")</h2>
+            <p id="optin-completed-view-p5" class="govuk-body">
+                @messages("optin.completedOptIn.cy.futureReporting.desc1", model.startYear, model.endYear)
```
mywyau, line 115: not sure if I've gotten them the wrong way round now. But think this key needs to be 
```
optin.completedOptIn.nymandated.futureReporting.desc1
```

to change the text to be 
"You have just chosen to voluntarily report quarterly from the {0} to {1} tax year."

when it is CY = Annual CY+1 = Mandated, to match the ticket.

mywyau (approved):
LGTM can see the changes running locally CY = Annual CY+1 = Mandated


Screenshot 2024-10-15 at 15 44 25

conf/messages
```diff
@@ -2315,8 +2315,14 @@ optin.completedOptIn.submitUpdates.desc1.b.url  = https://www.gov.uk/guidance/fi
...
+optin.completedOptIn.nymandated.futureReporting.desc1   = You have just chosen to voluntarily report quarterly from the {0} to {1} tax year.
+optin.completedOptIn.nymandated.futureReporting.desc2   = This could be because:
+optin.completedOptIn.nymandated.futureReporting.bullet.p1   = HMRC lowered the income threshold for Making Tax Digital for Income Tax
+optin.completedOptIn.nymandated.futureReporting.bullet.p2   = you reported an increase in your qualifying income in a tax return
+optin.completedOptIn.nymandated.inset   = From {0}, you’ll be required to send quarterly updates through compatible software.
 optin.completedOptIn.futureReporting.p1         = HMRC lowers the income threshold for Making Tax Digital for Income Tax
 optin.completedOptIn.futureReporting.p2         = you report an increase in your qualifying income in a tax return
 optin.completedOptIn.futureReporting.scenario   = For example, if your income from self-employment or property, or both, exceeds the threshold in the {0} to {1} tax year, you would have to report quarterly from 6 April {2}.
```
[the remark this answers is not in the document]
mywyau: note: nvm need to double check what we're doing going forward with welsh changes.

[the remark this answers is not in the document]
mywyau: nvm we're proceeding with out welsh

chandraAndem (approved):

mywyau (approved):

PR #2584, opened 2024-10-17
[MISUV-8594] Removing controller refs from the models and viewmodels

Mbrookesy:


TomRafferty (approved):
code looks good,
tests pass,
couldn't get ATs to run locally (some issue with my machine at the minute I believe)

happy to approve

PR #2589, opened 2024-10-18
MISUV-8257 Decommission Feature Switches: IV Uplift, Payment Allocation, Coding Out

dnyuar:


app/controllers/HomeController.scala
```diff
@@ -82,7 +82,7 @@ class HomeController @Inject()(val homeView: views.html.Home,
   private def buildHomePage(nextUpdatesDueDates: Seq[LocalDate], isAgent: Boolean, origin: Option[String])
                            (implicit user: MtdItUser[_]): Future[Result] =
     for {
-      unpaidCharges <- financialDetailsService.getAllUnpaidFinancialDetails(isEnabled(CodingOut))
+      unpaidCharges <- financialDetailsService.getAllUnpaidFinancialDetails(isCodingOutEnabled = true)
```
danielmaltz1, line 85: Here again, if the parameter will always be true, we don't need it

danielmaltz1:
Lots of places where you're passing something like (codingOutEnabled = true) as a parameter. Since this is always true can you just remove the logic for false from the called method, and then we don't need the parameter at all

danielmaltz1:
There are methods that take an isCodedOutEnabled parameter that you haven't changed, such as getChargeTypeKey in DocumentDetail and getDueDateForNonZeroBalancingCharge in ChargeItem and TaxYearSummaryChargeItem

danielmaltz1:
Also tests that have scenarios where coding out is false, that are no longer valid scenarios. Cmd+shift+F for "coding out is disabled"

PR #2591, opened 2024-10-21
Misuv 8585

Antony100:


conf/messages
```diff
@@ -1136,7 +1136,7 @@ paymentAllocation.tableHead.allocated-date                      = Date
 paymentAllocation.tableHead.amount                              = Amount
 paymentAllocation.taxYear										= Tax year {0} to {1}
 paymentAllocation.creditOnAccount                               = Credit on account
-paymentAllocation.moneyOnAccount                                = Money in your account
+paymentAllocation.moneyOnAccount                                = Remaining amount (added to available credit)
 paymentAllocation.moneyOnAccount.agent                          = Money in your client’s account
```
Antony100, line 1140: This message was not requested to be changed. However it seems like it should as it is the agent counterpart to the message that was requested.

danielmaltz1: Yeah I agree, we do display this content "money in your client's account" so worth going back to the BA and asking if it should be replaced

Antony100: It has been agreed to make the agent message the same as `moneyOnAccount`. Will make the change now.

app/views/partials/paymentAllocations/PaymentAllocationsCreditAmount.scala.html
```diff
@@ -35,21 +35,11 @@
 
 @getLink = @{
     if (creditsRefundsRepayEnabled) {
-        if(isAgent) {
```
Antony100, line 38: Now that the there's only the `moneyOnAccount` message, the check if the user is an agent is no longer required. The message is the same regardless of user type.

danielmaltz1 (approved):

PR #2598, opened 2024-10-28
MISUV-3816: PROTO

avelx:
This story is continuation of 1553 refactoring; in current iteration:
  * hide DocumentDetails within FinancialDetails, as we are planning to use ChargeItem / other TransactionItem subtypes applicable;


avelx:
New version of PROTOTYPE created;

PR #2601, opened 2024-10-30
MISUV-8478 Review & Reconcile POA - Feature 1 & 2 | New POA reconciliation credits being allocated to original POAs shown in POA history table

ayubAz:


app/models/financialDetails/ReviewAndReconcileCredit.scala
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.financialDetails
+
+import java.time.LocalDate
+
+case class ReviewAndReconcileCredit(taxYear: Int,
```
avelx, line 21: if we can use TaxYear type instead of Int to represent taxYear

app/services/ChargeHistoryService.scala
```diff
@@ -68,4 +69,32 @@ class ChargeHistoryService @Inject()(chargeHistoryConnector: ChargeHistoryConnec
     }
   }
 
+  def getReviewAndReconcileCredit(chargeItem: ChargeItem,
+                                  chargeDetailsforTaxYear: FinancialDetailsModel,
+                                  reviewAndReconcileEnabled: Boolean): Option[ReviewAndReconcileCredit] = {
+
+    if (reviewAndReconcileEnabled)
+      chargeDetailsforTaxYear
+        .getPairedDocumentDetails()
```
avelx, line 78: * if you can add comment/tag(?) around method getPairedDocumentDetails in FinancialDetailsModel as deprecated, as we are planning to hide DocumentDetail within FinancialDetailsModel and rely fully on ChargeItem in our domain instead: here is a story for reference => https://jira.tools.tax.service.gov.uk/browse/MISUV-3816;
Please add similar method to FinancialDetailsModel which will return ChargeItem and rely on this method in this part of the logic;

test/controllers/ChargeSummaryControllerSpec.scala
```diff
@@ -310,6 +312,21 @@ class ChargeSummaryControllerSpec extends MockAuthenticationPredicate
         JsoupParse(result).toHtmlDocument.select("h1").text() shouldBe successHeadingForRAR1(startYear.toString, endYear.toString)
         JsoupParse(result).toHtmlDocument.select("#payment-processing-bullets").isEmpty shouldBe true
       }
+
+      "display the Review & Reconcile credit for POA1 when present in the user's financial details" in new Setup(
+        financialDetailsModelWithPoaOneAndTwoWithRarCredits()) {
+        enable(ReviewAndReconcilePoa)
+        val result: Future[Result] = controller.show(testTaxYear, id1040000125)(fakeRequestWithNinoAndOrigin("PTA"))
+
+        println(s"\n${JsoupParse(result).toHtmlDocument}\n")
```
avelx, line 321: Please remove debug code

app/services/ChargeHistoryService.scala
avelx: also I am not entirely understand how chargeItem is related to charges in chargeDetailsforTaxYear; as I see they are joined purely on the fact that we are filtering out POA_credits from chargeDetailsforTaxYear

avelx (approved):

PR #2602, opened 2024-10-31
MISUV-8658 Potential fix for Feature Switch deployments in Staging/QA 

danielmaltz1:


danielmaltz1:
Will need to change feature-switches { read-from-mongo } back to false before merging

conf/application.conf
```diff
@@ -182,7 +182,7 @@ feature-switch {
 #  In the Staging Environment, this will be flipped to true
 #  Therefore changing the state of FS(s) will be easier during demos
 feature-switches {
-  read-from-mongo = false
+  read-from-mongo = true
```
avelx, line 185: I would prefer this flag to stag false in local as its in PROD, plus there will be quite a lot of failing UT/ITs as far as I know, we might need to wait for EMA team to complete they work around replacing current auth infrastructure with a new approach, before fixing these tests;

app/testOnly/controllers/FeatureSwitchController.scala
```diff
@@ -54,7 +54,7 @@ class FeatureSwitchController @Inject()(featureSwitchView: FeatureSwitchView,
 
   def show(): Action[AnyContent] = Action.async { implicit user =>
     featureSwitchService.getAll.flatMap { featureSwitches =>
-      val fss = featureSwitches.map(x => {
+      val fss = featureSwitches.filter(_.name.name != InvalidFS.name).map(x => {
```
avelx, line 57: should be fine to add this temp FS to staging, this is our internal tooling, we just need to drop this FS before next Show and Tell;

danielmaltz1: It doesn't show up on the test-only feature switch page, so they don't need to ever know about it

avelx: FeatureSwitch is out internal tool, not in prod for sure; should be fine to show it; but I guess really up to you

PR #2605, opened 2024-11-01
MISUV-8653 - Remove coded out POAs from the UI based on a feature switch

danielmaltz1:


app/services/WhatYouOweService.scala
```diff
@@ -43,18 +47,19 @@ class WhatYouOweService @Inject()(val financialDetailsService: FinancialDetailsS
...
-      } yield getWhatYouOweChargesList(unpaidChanges, isCodingOutEnabled, isReviewAndReconcile)
+      } yield getWhatYouOweChargesList(unpaidChanges, isCodingOutEnabled, isReviewAndReconcile, isFilterCodedOutPoas)
     }.flatten
   }
 
   def getWhatYouOweChargesList(unpaidCharges: List[FinancialDetailsResponseModel],
                                isCodingOutEnabled: Boolean,
-                               isReviewAndReconciledEnabled: Boolean)
+                               isReviewAndReconciledEnabled: Boolean,
+                               isFilterCodedOutPoas: Boolean)
```
avelx, line 62: not a big issue, but if we can use the same naming convention for isFilterCodedOutPoas?
ie. you can name it as isFilterCodedOutPoasEnabled or similar

conf/application.conf
```diff
@@ -174,6 +174,7 @@ feature-switch {
   enable-time-machine = false
   enable-session-data-storage = true
   enable-review-and-reconcile-poa = true
+  enable-filter-coded-out-poas = true
```
avelx, line 177: I would suspect that default should be set to false ? as we are not showing this logic at the moment

danielmaltz1: We are using this logic at the moment, but it was supposed to be behind a feature switch so we could remove it later. We are already filtering out coded out Poas

avelx (approved):
approved with comments

PR #2606, opened 2024-11-04
MISUV-8390 Continue ChargeHistoryModel refactoring

dnyuar:


app/models/chargeHistory/ChargeHistoryModel.scala
```diff
@@ -33,14 +33,14 @@ case class ChargeHistoryModel(taxYear: String,
...
-      AdjustmentReversalReason
+      Right(AdjustmentReversalReason.value)
     else
       reversalReason match {
-        case "amended return" => AmendedReturnReversalReason
-        case "Customer Request" => CustomerRequestReason
-        case _ => UnknownReversalReason
+        case "amended return" => Right(AmendedReturnReversalReason.value)
+        case "Customer Request" => Right(CustomerRequestReason.value)
+        case _ => Left(throw new Exception("Unknown reversal reason"))
```
avelx, line 43: no need to throw exception, just a create one instead; I would be surprised if you can create a UT for this change;

avelx: also should be ok to return Either[Throwable, ReversalReason] and consume .value where it is actually required, this way it should be easy to write required UTs;

app/models/chargeHistory/AdjustmentHistoryModel.scala
```diff
@@ -24,5 +24,5 @@ case class AdjustmentHistoryModel(creationEvent: AdjustmentModel,
                                   adjustments: List[AdjustmentModel])
 
 case class AdjustmentModel(amount: BigDecimal,
-                      adjustmentDate: Option[LocalDate],
-                           reasonCode: ReversalReason)
+                           adjustmentDate: Option[LocalDate],
+                           reasonCode: String)
```
avelx, line 28: if we can keep reasonCode to be of type ReversalReason, as in this case we can rely on the compile to spot obvious mistakes in the future, ie if we would have to apply refactoring etc

avelx (approved):
approved

avelx (approved):

PR #2617, opened 2024-11-07
MISUV-8553 Remove mandated following content

jmountain-hmrc:


test/views/optIn/OptInCompletedViewSpec.scala
```diff
@@ -137,7 +138,9 @@ class OptInCompletedViewSpec extends TestSupport {
       pageDocument.getElementsByClass(Selectors.pagePanelBodyClass).text() shouldBe messages("optin.completedOptIn.followingVoluntary.heading.desc", "2022", "2023")
 
       val expectedText: String = "From 6 April 2023, you’ll be required to send quarterly updates through compatible software."
-      pageDocument.getElementById(Selectors.warningInsetId).text() shouldBe expectedText
+      pageDocument.getElementById(Selectors.warningInsetAnnualFollowingId).text() shouldBe expectedText
+
+      pageDocument.getElementById(Selectors.warningInsetId) shouldBe null
```
avelx, line 143: If we can avoid use of null please, and stick with None / isDefined instead, thanks

PR #2619, opened 2024-11-07
MISUV-8500 - Re-implement dynamic inset text

TomRafferty:


app/views/optOut/OptOutChooseTaxYear.scala.html
```diff
@@ -98,6 +102,10 @@
         @messages("optout.chooseOptOutTaxYear.desc")
     }
 
+    @if(submissionCounts.isQuarterlyUpdatesMade) {
```
mywyau, line 105: Small comment, can we get a unit test in the OptOutChooseTaxYearViewSpec for the warning inset. 
The test setup scenario at first glance looks like it should have a quarterly update so it should be present but it may need a new test scenario. not completely sure tho.

TomRafferty: yup I can look into that
the change was tested to ensure it was dynamic but it would probably be a good idea to have an actual written test for it - especially given the scoverage readings

test/views/optOut/OptOutChooseTaxYearViewSpec.scala
```diff
@@ -83,4 +101,16 @@ class OptOutChooseTaxYearViewSpec extends TestSupport {
...
+
+    "show the warning block (for no submissions) user: non-agent" in new SetupNoSubmissions(false) {
+      pageDocument.getElementById("warning-inset") shouldBe null
+    }
+
+    "show the warning block (for no submissions) user: agent" in new SetupNoSubmissions(true) {
+      pageDocument.getElementById("warning-inset") shouldBe null
+    }
+  }
 }
```
mywyau, line 115: Think these tests can go, generally what I see is we can just ignore or not write tests if the component is not present. Also wording is off these tests I'm thinking would say  "not show"/"not present".

Also avoid using nulls in Scala there is almost never a need to use them. 
If we had to add a test for a element not being present we can likely do either:

1. Wrap the element in an Option and then check it is None, tbh it's kinda samey as throwing null.
2. Use css selectors think .select("#selector-goes-here") or some other methods like .getElements... return a Seq/List and we can check for an empty List()/Seq().

But yeah would go with deleting the test.

TomRafferty: happy to delete them, are we sure we want to go that route for dynamic text though? Just thinking if the logic gets tampered with in the future and starts showing all the time for whatever reason, we may not know as a test wouldn't flag - just a thought

mywyau (approved):

PR #2620, opened 2024-11-07
MISUV-8687 Create new Reporting Frequency View

mywyau:
opt out does not have a journey start page.

The logic for the links are off but we have tickets to update the link logic and have dynamic changing links for the page based on the user statuses and journey eligibility.

For now placing some logic as a placeholder and for demoing the page.

Had git history issues, cleaned up and using this pr/branch. For cleaner history.


TomRafferty (approved):
LGTM nice one

PR #2624, opened 2024-11-09
MISUV-8584: Dev - Apply Non Breaking Space on Dates Presented in Tables

ayubAz:


app/implicits/HtmlFormatter.scala
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package implicits
+
+import play.twirl.api.Html
+
+object HtmlFormatter {
+  implicit class NbspString(val str: String) extends AnyVal {
```
avelx, line 22: might be a good idea to add UT for this logic

avelx (approved):

PR #2626, opened 2024-11-11
MISUV-8776 Address issues with R&R Debits on Tax Year Summary Page

danielmaltz1:


app/models/financialDetails/ChargeItem.scala
```diff
@@ -140,11 +140,10 @@ object ChargeItem {
 
   implicit val format: Format[ChargeItem] = Json.format[ChargeItem]
 
-  def filterChargeWithOutstandingAmount(isChargeTypeEnabled:Boolean, chargeType: ChargeType*)
-                  (chargeItem: TransactionItem): Boolean = {
+  def filterAllowedCharges(isChargeTypeEnabled:Boolean, chargeType: ChargeType*)
```
avelx, line 143: as I understand we don't have any ITs coverage for filtering?

danielmaltz1: No it doesn't look like it, will add some

test/models/financialDetails/ChargeItemSpec.scala
```diff
@@ -365,4 +366,31 @@ class ChargeItemSpec extends UnitSpec with ChargeConstants  {
         }
       }
   }
+
+  "filterAllowedCharges" should {
+    "filter out FS related charges" when {
+      "relevant FS is disabled" in {
```
avelx, line 372: I would suggest to have one UT to cover scenario with empty List

avelx (dismissed):
approved with comments

PR #2638, opened 2024-11-14
MISUV-8750: FE: migration to income-tax-session-data service AC4

danielmaltz1:


app/controllers/agent/predicates/ClientConfirmedController.scala
```diff
@@ -17,22 +17,26 @@
...
 import uk.gov.hmrc.auth.core.retrieve.Name
 import uk.gov.hmrc.http.{HeaderCarrier, InternalServerException}
 
 import scala.concurrent.Future
 
-trait ClientConfirmedController extends BaseAgentController {
+trait ClientConfirmedController extends BaseAgentController with FeatureSwitching {
+
+  val appConfig: FrontendAppConfig
+  val sessionDataService: SessionDataService
```
avelx, line 39: * no need to expose this on the level of trait, should be fine to add this as implicit for the methods consuming this service, this would reduce number of controllers affected by this change;

PR #2639, opened 2024-11-14
[MISUV-8641] Implement Welsh/English Checker

Mbrookesy:


it/test/utils/WelshMessagesSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package utils
+
+import helpers.ComponentSpecBase
+
+class WelshMessagesSpec extends ComponentSpecBase {
```
avelx, line 21: We already have one as script:
income-tax-view-change-frontend/check_messages.sh
we would need to decide to use one of these I guess

Mbrookesy: The difference between this and the script is that this one is run automatically with the other tests, so it will always be checked.

Also, the script currently appears to be flagging comments as well, which is a bit odd

avelx: I think we can drop this old script, its was more of aesthetic choice rather then logical one to create this script;

it/test/utils/WelshMessagesSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  "all english messages should have a welsh translation" in {
+    val realMessages = messagesAPI.messages
+    val englishMessages = realMessages("default")
+    val welshMessages = realMessages("cy")
+
+    val missingWelshKeys = englishMessages.keySet.filterNot(key => welshMessages.keySet(key))
```
avelx, line 28: also this might be easy to compare Set's on this occasion; as there can be also missing English keys?
something like:
https://docs.scala-lang.org/overviews/collections-2.13/sets.html
`
xs diff ys or xs &~ ys | The set difference of xs and ys.

`

mywyau (approved):
lgtm runs fine and throws errors when there is a missing key

Screenshot 2024-11-20 at 12 41 19

PR #2651, opened 2024-11-21
MISUV-8125: Refactor SessionService.getMongo to take parameter 'JourneyType'

ayubAz:


app/enums/JourneyType/JourneyType.scala
```diff
@@ -37,3 +39,19 @@ case object Manage extends Operation {
...
+
+case class Opt(optJourney: OptJourney) extends JourneyType {
+  override def toString: String = optJourney.choice
+}
+
+sealed trait OptJourney extends JourneyType {
+  val choice: String
+}
+
+case object In extends OptJourney {
```
avelx, line 51: OptInJourney

app/enums/JourneyType/JourneyType.scala
```diff
@@ -37,3 +39,19 @@ case object Manage extends Operation {
...
+  override val choice = "OPTIN"
+}
+
+case object Out extends OptJourney {
```
avelx, line 55: OptOutJourney

app/enums/JourneyType/JourneyType.scala
```diff
@@ -37,3 +39,19 @@ case object Manage extends Operation {
 case object Cease extends Operation {
   override val operationType = "CEASE"
 }
+
+case class Opt(optJourney: OptJourney) extends JourneyType {
```
avelx, line 43: OptInOutJourney ? not sure, check the context

app/enums/JourneyType/JourneyType.scala
```diff
@@ -18,7 +18,9 @@ package enums.JourneyType
 
 import enums.IncomeSourceJourney.IncomeSourceType
 
-case class JourneyType(operation: Operation, businessType: IncomeSourceType) {
+sealed trait JourneyType
+
+case class IncomeSources(operation: Operation, businessType: IncomeSourceType) extends JourneyType {
```
avelx, line 23: this naming convention would create confusion, if you planning to generalise this
you can use something like: IncomeSourceJourneyType or similar to preserver meaning

app/enums/JourneyType/JourneyType.scala
```diff
@@ -18,7 +18,9 @@ package enums.JourneyType
 
 import enums.IncomeSourceJourney.IncomeSourceType
 
-case class JourneyType(operation: Operation, businessType: IncomeSourceType) {
+sealed trait JourneyType
```
avelx, line 21: You can name is BaseJourneyType or similar,
as we have to preserver original naming convention which is

app/enums/JourneyType/JourneyType.scala
Antony100: The changes you've suggested make it clearer and will implement. `Opt(optJourney)` may be clearer as a result? When this class is called with the proposed changes, we would see `Opt(OptInJourney)` or `Opt(OptOutJourney)`

avelx: I suspect you can leave it as is, its just a bit confusing when I try to read this separately from the actual usage;

avelx:
quite a few imports need to be dropped / optimised, but this can be fixed across the project by running ScalaFix import optimise rules / so no action: https://scalacenter.github.io/scalafix/docs/rules/OrganizeImports.html

avelx (approved):

PR #2656, opened 2024-11-22
[MISUV-8824] Upgrading AddressLookupConnector to use HttpClientV2

Mbrookesy:


test/connectors/AddressLookupConnectorSpec.scala
```diff
@@ -130,7 +127,7 @@ class AddressLookupConnectorSpec extends TestSupport with FeatureSwitching with
           enable(IncomeSources)
           beforeEach()
 
-          setupMockHttpPost(TestAddressLookupConnector.addressLookupInitializeUrl, addressJson(individualContinueUrl, individualFeedbackUrl, individualEnglishBanner, individualWelshBanner))(HttpResponse(status = OK,
+          setupMockHttpV2Post(TestAddressLookupConnector.addressLookupInitializeUrl)(HttpResponse(status = OK,
```
TomRafferty, line 130: Do we lose anything by dropping out parameters such as `individualWelshBanner` ?

TomRafferty (commented):
just a general question
other that that LGTM

test/connectors/AddressLookupConnectorSpec.scala
Mbrookesy: Not really, the only reason they are there is that the setup files are very specific on the argument matchers for the mocks so the mocks use these variables to only happen on them specifically. 

They can be added back but personally i felt the desired result would be the same regardless if you used a more specific mock or a generic one like we have done with most our other tests.

TomRafferty (approved):
LGTM

PR #2661, opened 2024-11-25
MISUV-3816  FE: 1553 refactoring / FinancialDetailsModel -> documentDetails

dnyuar:


dnyuar:
- Setting documentDetails parameter in FinancialDetailsResponseModel to a private val
- Adding poaRelevantAmount parameter to the ChargeItem model
- Refactoring methods which used the documentDetails parameter
- Updating unit and IT tests
- Removing unused methods from FinancialDetailsResponseModel

avelx (approved):

PR #2666, opened 2024-11-27
Removed Confidence Level check and test

danielmaltz1:


avelx (approved):

avelx:
agent should not be check against CL

PR #2686, opened 2024-12-05
MISUV-8901 Session-data service: dynamically load client details

danielmaltz1:


app/auth/authV2/actions/RetrieveClientData.scala
```diff
@@ -47,23 +49,29 @@ class RetrieveClientData @Inject()(sessionDataService: SessionDataService,
...
+        clientDetailsService.checkClientDetails(sessionData.utr).map {
+          case Right(name) => Right(ClientDataRequest(
+              sessionData.mtditid,
+              name.firstName,
+              name.lastName,
+              sessionData.nino,
+              sessionData.utr,
+              getBooleanFromSession(SessionKeys.isSupportingAgent),
+              confirmed = {
+                r.session.get(SessionKeys.confirmedClient) match {
```
avelx, line 63: my understanding is that we can fully rely on session-data service in this case, as client is confirmed right after we write to this service? not sure why we still need to have a split for this operation in two steps; might be something we need to discuss

app/auth/authV2/actions/RetrieveClientData.scala
```diff
@@ -47,23 +49,29 @@ class RetrieveClientData @Inject()(sessionDataService: SessionDataService,
...
-        r.session.get(SessionKeys.clientLastName),
-        sessionData.nino,
-        sessionData.utr,
-        getBooleanFromSession(SessionKeys.isSupportingAgent),
-        confirmed = {
-          r.session.get(SessionKeys.confirmedClient) match {
-            case Some(value) if value != "" => value.toBoolean
-            case _ => useSessionDataService
-          }
+    sessionDataService.getSessionData(useCookie = !useSessionDataService).flatMap {
```
avelx, line 52: :thinking: - ideally if we can refactor this change to use for-comprehension, so this code would be flexible for changes in the future / EitherT / OptionT use really up to you to decide;

it/test/controllers/agent/EnterClientsUTRControllerISpec.scala
```diff
@@ -513,7 +513,7 @@ class EnterClientsUTRControllerISpec extends ComponentSpecBase with FeatureSwitc
       "there was a failed future when posting data to the income tax session data service" in {
         val validUTR: String = "1234567890"
         stubAgentAuthorisedUser(true)
-        CitizenDetailsStub.stubGetCitizenDetails(validUTR)(
+        CitizenDetailsStub.stubGetCitizenDetails(validUTR,
```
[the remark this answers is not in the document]
danielmaltz1: We have these tests in that same file:
"an empty utr string is submitted"
"a utr containing non-digits is submitted"
"a utr which has less than 10 digits is submitted"
"a utr which has more than 10 digits is submitted"

app/auth/authV2/actions/RetrieveClientData.scala
danielmaltz1: I'm not sure, are you saying we should just set confirmed to true in all cases? I'm still not sure what ConfirmedClient means to be honest

avelx: we starting session/calling income-tax-session-data service in EnterClientsUTRController, but actual session should not start for the agent, until agent confirmed that its ok to continue for the client selected -> which is ConfirmClientUTRController: essentially starting a session for the agent is a two step operation;
 we did not discussed this originally and it seems that we are starting session for the agent in the wrong place;

danielmaltz1: Ok, what would you like me to do in this PR?

app/auth/authV2/actions/AsMtdUser.scala
```diff
@@ -72,26 +91,27 @@ class AsMtdUser @Inject()
 
     implicit val r = request
 
-    val (optMtdId, optClientName) = request.affinityGroup match {
-      case Some(Agent) => (getClientMtdid, getClientName)
-      case _ => (Future.successful(request.mtdId), None)
+    for {
+      mtdAndUtr <- getClientMtdidAndUtr
+      clientName <- getClientName(mtdAndUtr._2)
```
avelx, line 96: if we can avoid using tuple in this way: ._2 / _1 as it hard to read this code later;

danielmaltz1: I'm trying to make it so that we only have to call the SessionDataService once, but yeah I'll try to make this more readable

app/auth/authV2/actions/RetrieveClientData.scala
avelx: no, I've created a separate story for this: https://jira.tools.tax.service.gov.uk/browse/MISUV-8926

avelx (approved):

PR #2689, opened 2024-12-06
MISUV-8724 added new auth solution for supporting agent controllers

Chrissystk:


test/services/optout/OptOutPropositionSpec.scala
```diff
@@ -99,7 +99,6 @@ class OptOutPropositionSpec extends UnitSpec {
 
     rowsWithoutHeaders.foreach(line => {
       val cells = line.split("\t").take(noOfColumns)
-//      println(cells.mkString("  --  "))
       println(s"(${
```
avelx, line 102: looks like we still got **_println_** in our code base

test/mocks/auth/MockAuthActions.scala
```diff
@@ -53,10 +61,15 @@ trait MockAuthActions extends
   override def beforeEach(): Unit = {
     super.beforeEach()
     disableAllSwitches()
-    Play.stop(fakeApplication())
+//    Play.stop(fakeApplication())
```
avelx, line 64: if this code is not required, we can drop it?

it/test/helpers/GenericStubMethods.scala
```diff
@@ -22,10 +22,8 @@ trait GenericStubMethods extends CustomMatchers {
 
   def isAuthorisedUser(authorised: Boolean): Unit = {
     if (authorised) {
-      Given("I wiremock stub an isAuthorisedUser user response")
```
avelx, line 25: not sure why we are dropping Given statements here, its a small thing, but I don't see reason to drop these

conf/incomeSources.routes
```diff
@@ -283,12 +283,12 @@ GET          /agents/income-sources/manage/your-details-uk-property            c
...
-GET          /income-sources/manage/uk-property-will-report                     controllers.incomeSources.manage.ManageObligationsController.showUKProperty(changeTo: String, taxYear: String)
-GET          /agents/income-sources/manage/uk-property-will-report              controllers.incomeSources.manage.ManageObligationsController.showAgentUKProperty(changeTo: String, taxYear: String)
-GET          /income-sources/manage/foreign-property-will-report                controllers.incomeSources.manage.ManageObligationsController.showForeignProperty(changeTo: String, taxYear: String)
-GET          /agents/income-sources/manage/foreign-property-will-report         controllers.incomeSources.manage.ManageObligationsController.showAgentForeignProperty(changeTo: String, taxYear: String)
+GET          /income-sources/manage/business-will-report                        controllers.incomeSources.manage.ManageObligationsController.show(changeTo: String, taxYear: String, incomeSourceType: IncomeSourceType = SelfEmployment)
```
avelx, line 286: this change would imply changes in ATs required, and this is strange that current set of ATs not failing

avelx (approved):
approved with comments

conf/incomeSources.routes
Chrissystk: The url hasn't changed here and just the function in the controller is now just one shared controller rather than 3 separate ones.

test/mocks/auth/MockAuthActions.scala
Chrissystk: This was the fix for the tests length. I will remove it in my next PR.

test/services/optout/OptOutPropositionSpec.scala
Chrissystk: This is one of the weird files that isn't part of my code change. I think it needs looking at separately.

PR #2698, opened 2024-12-13
MISUV-8926 - Move session creation to ConfirmClientUTRController

danielmaltz1:
AgentHasClientDetails: At the moment this always happens after RetrieveClientData, which will fail the auth if the session data flag is true and there is not a valid entry in the service. So we can assume that if the flag is true and we have reached AgentHasClientDetails, there must be a valid entry in the service. So if the flag is true, ClientConfirmed should be true
RetrieveClientData: This code is within the success case of the session data service call, so we know there is a valid client, so it's true if the flag is true
EnterClientsUTR and ConfirmClientUTR controllers: The session data service call has been moved from Enter to Confirm, but setting the cookies has been left in Enter
SessionCookieUtil has been deleted and incorporated into ConfirmClientUTRController, as that was the only place that used it


app/auth/authV2/actions/RetrieveClientDataFromCookies.scala
```diff
@@ -0,0 +1,68 @@
...
+
+    implicit val r: Request[A] = request
+    implicit val hc: HeaderCarrier = HeaderCarrierConverter
+      .fromRequestAndSession(request, request.session)
+
+    sessionDataService.getSessionData(useCookie = true).map {
+      case Right(sessionData) => Right(ClientDataRequest(
+        sessionData.mtditid,
+        r.session.get(SessionKeys.clientFirstName),
+        r.session.get(SessionKeys.clientLastName),
```
Chrissystk, line 52: Should you still be using the session cookie for storing the clientName? Every other request (requiring an enrolled user) calls Citizen details to get the client Name and also looking at the confluence page linked in your ticket it has clientName as something that shouldn't be stored in the cookie.

avelx: the only place where we use this predicate is within asMTDAgentWithUnconfirmedClient / ConfirmClientUTRController;
at this stage we don't know yet client details as client not yet selected, so old logic is trying to read details from existing browser cookies, which will not be available after migration/but at this stage we just fallback to use cookies and we not yet able to call service at this stage;

danielmaltz1: We'll have to change to using Citizen Details eventually, so I'm happy to switch over to using it now if you're ok with that @avelx

danielmaltz1: I've now changed RetrieveClientDetails so we don't need RetrieveClientDetailsFromCookies anymore, so both methods will now use the citizen details service

PR #2702, opened 2024-12-18
MISUV-8931-CS added new auth solutions to controllers that support individuals and primary agents only

Chrissystk:


test/testConstants/PropertyDetailsTestConstants.scala
```diff
@@ -120,7 +120,7 @@ object PropertyDetailsTestConstants {
   val uKPropertyDetails2 = PropertyDetailsModel(
     incomeSourceId = testPropertyIncomeId2,
     accountingPeriod = Some(testPropertyAccountingPeriod),
-    firstAccountingPeriodEndDate = None,
+    firstAccountingPeriodEndDate = Some(LocalDate.of(year2018, Month.APRIL, 5)),
```
avelx, line 123: looks like something you can re-use from TaxYear type => 
```
  def toFinancialYearEnd: LocalDate = {
    LocalDate.of(this.endYear, April, Fifth)
  }
```

test/controllers/PaymentControllerSpec.scala
```diff
@@ -16,157 +16,98 @@
...
+    s"${if (isAgent) "agentP" else "p"}aymentHandoff" when {
+      s"the $mtdUserRole is authenticated" should {
+        if (mtdUserRole == MTDSupportingAgent) {
+          testSupportingAgentDeniedAccess(action)(fakeRequest)
+        } else {
+          "redirect to the payment api provided by payAPI" when {
+            "a successful payments journey is started" in {
+              setupMockSuccess(mtdUserRole)
+              mockSingleBusinessIncomeSource()
+              when(mockPayApiConnector.startPaymentJourney(ArgumentMatchers.eq(testSaUtrId), ArgumentMatchers.eq(BigDecimal(10000)),
```
avelx, line 64: not sure if this is the way calling API / service consuming this amount, but I would prefer to provide amount in pounds and pence, ie: BigDecimal(10000.19)

app/controllers/TaxYearsController.scala
```diff
@@ -17,49 +17,55 @@
...
+        isAgent = isAgent,
+        utr = user.saUtr,
+        itsaSubmissionIntegrationEnabled = isEnabled(ITSASubmissionIntegration),
+        earliestSubmissionTaxYear = earliestSubmissionTaxYear,
+        btaNavPartial = user.btaNavPartial,
+        origin = origin)
+    } match {
+      case Success(taxView) => Future.successful(Ok(taxView))
+      case Failure(_) =>
+        val errorHandler = if(isAgent) agentItvcErrorHandler else itvcErrorHandler
```
avelx, line 63: if we can log error as well

app/controllers/PaymentHistoryController.scala
```diff
@@ -125,32 +116,15 @@ class PaymentHistoryController @Inject()(val paymentHistoryView: PaymentHistory,
...
-              backUrl = "",
-              itvcErrorHandler = itvcErrorHandler,
-              isAgent = false
-            )
+  def refundStatus: Action[AnyContent] =
+    authActions.asMTDIndividual.async { implicit user =>
+      if (isEnabled(PaymentHistoryRefunds)) {
+        repaymentService.view(user.nino).map {
+          case Right(nextUrl) => Redirect(nextUrl)
+          case Left(_) => itvcErrorHandler.showInternalServerError()
```
avelx, line 124: I understand that this is not your code, but if we can add error logging as well

conf/app.routes
```diff
@@ -150,7 +150,7 @@ GET         /language/:lang                                               contro
 #Payment history
 GET         /payment-refund-history                                       controllers.PaymentHistoryController.show(origin: Option[String] ?= None)
 GET         /agents/payment-refund-history                                controllers.PaymentHistoryController.showAgent
-GET         /refund-status                                                controllers.PaymentHistoryController.refundStatus(isAgent: Boolean = false)
+GET         /refund-status                                                controllers.PaymentHistoryController.refundStatus
```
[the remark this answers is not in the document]
Chrissystk: Yes the code was written so it would give the internal error page if called as an agent. The view file which contains the link to this page only calls for an individual user as well so it didn't make sense to call it with an isAgent flag and instead just restrict the endpoint to Individuals only.

PR #2705, opened 2024-12-18
MISUV-8931-CS added new auth solutions to controllers that don't support supporting agents

Chrissystk:


test/controllers/optOut/OptOutCancelledControllerSpec.scala
```diff
@@ -44,15 +44,15 @@ class OptOutCancelledControllerSpec extends MockAuthActions with MockOptOutServi
...
+  override lazy val app: Application = applicationBuilderWithAuthBindings
     .overrides(
       api.inject.bind[OptOutService].toInstance(mockOptOutService)
     )
     .configure(config)
     .build()
 
-  val testController = fakeApplication().injector.instanceOf[OptOutCancelledController]
-  val optOutCancelledView = fakeApplication().injector.instanceOf[OptOutCancelledView]
+  val testController = app.injector.instanceOf[OptOutCancelledController]
```
avelx, line 54: might worth to change these two vals to be lazy? as we are using lazy app right away

test/controllers/optOut/ConfirmedOptOutControllerSpec.scala
```diff
@@ -35,12 +35,12 @@ import scala.concurrent.Future
...
   with MockOptOutService {
 
-  override def fakeApplication(): Application = applicationBuilderWithAuthBindings()
+  override lazy val app: Application = applicationBuilderWithAuthBindings
     .overrides(
       api.inject.bind[OptOutService].toInstance(mockOptOutService)
     ).build()
 
-  val testController = fakeApplication().injector.instanceOf[ConfirmedOptOutController]
+  val testController = app.injector.instanceOf[ConfirmedOptOutController]
```
avelx, line 43: same thing, we are using lazy val right away

avelx:
Next pattern looks a bit strange to me as we use lazy val right away:
```
override lazy val app: Application = ... 
val testController = app.injector.instanceOf
```

it/test/controllers/CalculationPollingControllerISpec.scala
```diff
@@ -16,146 +16,245 @@
...
+              "calculation service returns retryable response back initially and then returns success response before interval time completed" in {
+                stubAuthorised(mtdUserRole)
+                IncomeTaxViewChangeStub.stubGetIncomeSourceDetailsResponse(testMtditid)(OK, multipleBusinessesAndUkProperty)
+                IncomeTaxCalculationStub.stubGetCalculationErrorResponseByCalcId(testNino, "idFour", testTaxYear)(NO_CONTENT,
+                  LiabilityCalculationError(NO_CONTENT, "no content"))
+
+                val res = buildGETMTDClient(finalCalcPath, additionalCookies ++ Map(SessionKeys.calculationId -> "idFour"))
+
+                //After 1.75 seconds responding with success message
+                Thread.sleep(1750)
```
avelx, line 190: Please don't use Thread.sleep anywhere in our code base

Chrissystk: This is a preexisting tests and not something I introduced. In this case as well they are testing polling so the thread.sleep is delaying the successful stub call not waiting for a result so the thread.sleep approach is not wrong in this scenerio.

test/controllers/optOut/ConfirmedOptOutControllerSpec.scala
Chrissystk: I have updated to use lazy val but I don’t believe it was necessary as yes the “val” was called straight away but the nature of classes mean that they are only executed at run time. The issue seemed to be more around the fakeApplication() function. I expect under the covers it was being called several times and even though we was stopping the play application after every test class some files where remaining open.

app/controllers/TaxYearsController.scala
```diff
@@ -17,49 +17,55 @@
...
+        isAgent = isAgent,
+        utr = user.saUtr,
+        itsaSubmissionIntegrationEnabled = isEnabled(ITSASubmissionIntegration),
+        earliestSubmissionTaxYear = earliestSubmissionTaxYear,
+        btaNavPartial = user.btaNavPartial,
+        origin = origin)
+    } match {
+      case Success(taxView) => Future.successful(Ok(taxView))
+      case Failure(_) =>
+        val errorHandler = if(isAgent) agentItvcErrorHandler else itvcErrorHandler
```
avelx, line 63: if we can log error here

it/test/controllers/CalculationPollingControllerISpec.scala
avelx: Next story created to address this issue:
https://jira.tools.tax.service.gov.uk/browse/MISUV-8977

avelx (approved):
approved with comments

PR #2716, opened 2025-01-08
MISUV-8991: Addressed CreditService

danielmaltz1:


app/services/CreditService.scala
```diff
@@ -34,6 +34,7 @@ import scala.util.Try
 class CreditService @Inject()(val financialDetailsConnector: FinancialDetailsConnector,
                               implicit val dateService: DateServiceInterface)
                              (implicit ec: ExecutionContext, implicit val appConfig: FrontendAppConfig) {
+  @deprecated("Migrating as part of MISUV-8845")
```
avelx: if we can change description a bit: "Use getAllCreditsV2 instead" or similar

avelx:
As discussed, please add more tests to CreditServiceSpec

avelx (approved):

PR #2717, opened 2025-01-09
[MISUV-9009] Ceased business warning section for Reporting Frequency

Mbrookesy:


Mbrookesy:
please test again

app/controllers/ReportingFrequencyPageController.scala
```diff
@@ -70,13 +70,16 @@ class ReportingFrequencyPageController @Inject()(
             }
           }
 
+          val isCeased = user.incomeSources.businesses.forall(_.isCeased)
```
chandraAndem, line 73: Could we move this to IncomeSourceDetailsResponseModel please? It helps to reuse in other places. Something like:

`def areAllBusinessesCeased: Boolean = businesses.forall(_.isCeased) `

chandraAndem (changes requested):
Small comment on PR and also this ticket is missing AT's subtask.

chandraAndem (approved):

PR #2721, opened 2025-01-10
MISUV-8991 financial details service

danielmaltz1:


app/connectors/FinancialDetailsConnector.scala
```diff
@@ -191,6 +191,45 @@ class FinancialDetailsConnector @Inject()(
       }
   }
 
+  def getFinancialDetails(taxYearFrom: Int, taxYearTo: Int, nino: String)
```
avelx, line 194: worth to use TaxYear type for any new code added as there are method to transform it to YYYY-MM-DD format

app/connectors/FinancialDetailsConnector.scala
```diff
@@ -191,6 +191,45 @@ class FinancialDetailsConnector @Inject()(
       }
   }
 
+  def getFinancialDetails(taxYearFrom: TaxYear, taxYearTo: TaxYear, nino: String)
```
avelx, line 194: We might need to introduce a new type to represent range of tax years: say something like TaxYearsRange ? or similar not sure if we do have something at the moment in our domain but worth to add it (in the scope of this PR I think this would be optional)

avelx (approved):

app/connectors/FinancialDetailsConnector.scala
avelx: next tech story created: https://jira.tools.tax.service.gov.uk/browse/MISUV-9063

PR #2722, opened 2025-01-10
MISUV-8397 Opt in cancelled

jmountain-hmrc:


app/controllers/optIn/OptInCancelledController.scala
```diff
@@ -0,0 +1,89 @@
...
+  extends FrontendController(mcc) with I18nSupport with FeatureSwitching {
+
+  private def pageRender(isAgent: Boolean = false)
+                        (implicit user: MtdItUser[AnyContent]): Future[Result] = {
+    for {
+      proposition <- optInService.fetchOptInProposition()
+      availableOptInYears = proposition.availableOptInYears
+      isOneYearOptIn = proposition.isOneYearOptIn
+    } yield {
+      if (isOneYearOptIn) {
```
michael-hmrc, line 52: wonder if we can simplify this in the pattern match as a if guard?

michael-hmrc (approved):

PR #2728, opened 2025-01-15
MISUV-9024 - Reporting frequency page entry point

TomRafferty:


app/views/nextUpdates/NextUpdatesOptOut.scala.html
```diff
@@ -62,42 +64,69 @@
...
 
-				case m:OptOutOneYearViewModel =>
-					if (m.showWarning) {
+					case m:OptOutOneYearViewModel =>
 						Html(
 							s"""
 							<span id="one-year-opt-out-message">
-								${messages("nextUpdates.optOutOneYear-1", m.startYear, m.endYear)}
-								${link(id = Some("single-year-opt-out-warning-link"), link = controllers.optOut.routes.SingleYearOptOutWarningController.show(isAgent).url, messageKey = "nextUpdates.optOutOneYear-2")}
+								${messages("nextUpdates.withReportingFrequencyContent.optOutOneYear-1", m.startYear, m.endYear)}
```
michael-hmrc, line 74: think we can simplify this a bit and move some logic out of the views.

We can we use pattern match guards to determine the most of the html and link redirects etc. 

Then just pass like a model of everything into a single

```
Html(
     s"""
            <span id="one-year-opt-out-message">
                 ${messages("nextUpdates.optOutOneYear-1", m.startYear, m.endYear)}
                 ${link(id = Some("single-year-opt-out-warning-link"), link = controllers.optOut.routes.SingleYearOptOutWarningController.show(isAgent).url, messageKey = "nextUpdates.optOutOneYear-2")}
           </span>
      """
)
 ```

Comment (michael-hmrc) [reply app/views/nextUpdates/NextUpdatesOptOut.scala.html:74]: formatting is a bit of a pain in github comments sorry.

File: app/controllers/NextUpdatesController.scala [hunk 2]
```diff
@@ -59,7 +59,7 @@ class NextUpdatesController @Inject()(NoNextUpdatesView: NoNextUpdates,
     }
   }
 
-  def getNextUpdates(backUrl: Call, isAgent: Boolean, errorHandler: ShowInternalServerError, origin: Option[String] = None)
+  def getNextUpdates(backUrl: Call, isAgent: Boolean, errorHandler: ShowInternalServerError, origin: Option[String] = None, showReportingFrequencyContent: Boolean = false)
```
michael-hmrc, line 62: think we don't need a boolean flag here as a parameter and can just call the feature switch inside this method? 

```
val shouldShowReportingContent: Boolean = isEnabled(ReportingFrequencyPage)
```

also I try to not set default values for params unless it's for a  good reason or massive syntax convenience. Helps with spotting issues sooner with the code. Since the value is not defaulted, if the value needs to be passed around a lot it explicitly defined at each point.

TomRafferty: will look into all the comments ta 😄

michael-hmrc: wonder what you think about something like this?

https://github.com/hmrc/income-tax-view-change-frontend/compare/main...MISUV-9024b

app/views/nextUpdates/NextUpdatesOptOut.scala.html
```diff
@@ -127,6 +156,7 @@ <h2 class="govuk-heading-m">@messages("nextUpdates.sub-heading")</h2>
     }
 }
 
+@import models.admin.ReportingFrequencyPage
```
michael-hmrc, line 159: this import is not needed

it/test/controllers/NextUpdatesControllerISpec.scala
```diff
@@ -301,9 +301,7 @@ class NextUpdatesControllerISpec extends ControllerISpecHelper {
 
         Then("the quarterly updates info sections")
         res should have(
-          elementTextBySelector("#one-year-opt-out-message")(expectedValue = "You are currently reporting quarterly on a " +
-            "voluntary basis for the 2021 to 2022 tax year. You can choose to opt out of quarterly updates and " +
-            "report annually instead.")
+          elementTextBySelector("#one-year-opt-out-message")(expectedValue = "You are currently reporting quarterly on a voluntary basis for the 2021 to 2022 tax year. You can choose to opt out of quarterly updates and report annually instead.")
```
michael-hmrc, line 304: I know this was pre-existing but think we should be testing content in the views instead of the IT tests tbh. 

We can move these checks to the views instead and for more scenarios, easier to test there I think

test/views/nextUpdates/NextUpdatesOptOutViewSpec.scala
```diff
@@ -17,6 +17,7 @@
 package views.nextUpdates
```
michael-hmrc, line 17: There is a 

`obligationsMessages` object can this be capitalised and moved into the messages package in the test/views/messages directory?

conf/messages
```diff
@@ -769,11 +769,16 @@ nextUpdates.propertyIncome            							   = Property income
...
+
 nextUpdates.optOutOneYear-1                                        = You are currently reporting quarterly on a voluntary basis for the {0} to {1} tax year. You can choose to
 nextUpdates.optOutOneYear-2                                        = opt out of quarterly updates and report annually instead.
+nextUpdates.withReportingFrequencyContent.optOutOneYear-1          = Depending on your circumstances, you may be able to
+nextUpdates.withReportingFrequencyContent.optOutOneYear-2          = view and change your reporting frequency.
 
 nextUpdates.optOutMultiYear-1                                      = You are currently reporting quarterly on a voluntary basis. You can choose to
 nextUpdates.optOutMultiYear-2                                      = opt out of quarterly updates and report annually instead.
+nextUpdates.withReportingFrequencyContent.optOutMultiYear-1        = Depending on your circumstances, you may be able to
+nextUpdates.withReportingFrequencyContent.optOutMultiYear-2        = view and change your reporting frequency.
```
Mbrookesy, line 781: Currently i'm doing the ATs for this and I noticed that the added messages (Specifically "view and change your reporting frequency") aren't being used within the views you edited

Mbrookesy (changes requested):

conf/messages
Mbrookesy: Also can you do me a favour and add IDs to the opt in/out links in the reporting frequency page (https://github.com/hmrc/income-tax-view-change-frontend/blob/main/app/views/ReportingFrequencyView.scala.html#L61-L79) so i can get selenium to click them

PR #2738, opened 2025-01-23
MISUV-9007 Adding scenario based content for optout confirmed page

michael-hmrc:
Requirements and scenarios covered in the ticket is a bit weird, basically we showed content for the submit your tax return with no logic originally.

However, we are now according to the ticket displaying logic based on itsa statuses & CY & CY+1 choices.

The scenarios covered in the ticket miss out single tax year choice scenarios and possible CY-1 choices. This has been raised with Bethrand and also other combinations of itsa statuses.

After speaking with Bethrand it is likely we ultimately only really care about the user's chosen intent tax year and not the combination of itsa statuses. But atm not sure so coded out the ticket logic. A comment has been raised to raise a ticket for other scenarios.

**Update:** 
Have a feeling we may need to for now add content for the previous years and some combinations not mentioned in the ticket. Added some logic to display content for various CY-1 scenarios and only for a multi year if the user chooses the most recent/in the future year possible it will display the 2nd paragraph.

### DIFF
--- a/app/config/FrontendAppConfig.scala
+++ b/app/config/FrontendAppConfig.scala
@@ -59,7 +59,6 @@ class FrontendAppConfig @Inject()(val servicesConfig: ServicesConfig, val config
   //SA for Agents Online Service
   lazy val saForAgents: String = "https://www.gov.uk/guidance/self-assessment-for-agents-online-service"
 
-
   //GG Sign In via BAS Gateway
   lazy val signInUrl: String = servicesConfig.getString("base.sign-in")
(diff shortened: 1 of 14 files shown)


Mbrookesy:
Opt out of CY when both CY and CY+1 are Voluntary

image

Mbrookesy:
Opt out of CY+1 when both CY and CY+1 are voluntary

image

Mbrookesy:
Opt out of CY onwards when CY+1 is annual

image

Mbrookesy (dismissed):

Mbrookesy (approved):

PR #2742, opened 2025-01-27
MISUV-8845 Refactor claim refund to retrieve data from API 1553 in a single call

danielmaltz1:


app/controllers/TaxYearSummaryController.scala
```diff
@@ -152,15 +153,16 @@ class TaxYearSummaryController @Inject()(authActions: AuthActions,
   private def withTaxYearFinancials(taxYear: Int, isAgent: Boolean)(f: List[TaxYearSummaryChargeItem] => Future[Result])
                                    (implicit user: MtdItUser[_]): Future[Result] = {
 
-    financialDetailsService.getFinancialDetails(taxYear, user.nino) flatMap {
+    val taxYearTyped = TaxYear(taxYear - 1, taxYear)
+    financialDetailsService.getFinancialDetails(taxYearTyped, taxYearTyped, user.nino) flatMap {
```
avelx, line 157: looks a bit strange with respect to taxYear we are passing, looks like we are passing the same value

danielmaltz1: We only want the data for that tax year, so the range is the start of taxYearTyped to the end of taxYearTyped, so we pass the same one twice. I could make a new method in financialDetailsService to just get financial details for a single year, if that makes it clearer

test/controllers/ChargeSummaryControllerSpec.scala
```diff
@@ -390,18 +390,19 @@ class ChargeSummaryControllerSpec extends ChargeSummaryControllerHelper {
...
-                setupMockSuccess(mtdUserRole)
-                mockBothIncomeSources()
-
-                val result: Future[Result] = testController.show(2020, "CODINGOUT01")(fakeRequest)
-
-                status(result) shouldBe Status.INTERNAL_SERVER_ERROR
-                JsoupParse(result).toHtmlDocument.select("h1").text() shouldBe errorHeading
-              }
-            }
+            //TODO: Check validity of this test
```
avelx, line 393: do we really need this commented code?

danielmaltz1: Oh yep sorry, still need to investigate this test. Doing now

app/services/WhatYouOweService.scala
```diff
@@ -56,33 +56,37 @@ class WhatYouOweService @Inject()(val financialDetailsService: FinancialDetailsS
...
           case Some(outstandingChargesModel) => whatYouOweChargesList.copy(outstandingChargesModel = Some(outstandingChargesModel))
           case _ => whatYouOweChargesList
         }
+      case None =>
+        val whatYouOweChargesList = WhatYouOweChargesList(BalanceDetails(0.00, 0.00, 0.00, None, None, None, None, None))
+        callOutstandingCharges(mtdUser.saUtr, mtdUser.incomeSources.yearOfMigration, dateService.getCurrentTaxYearEnd).map {
+          case Some(outstandingChargesModel) => whatYouOweChargesList.copy(outstandingChargesModel = Some(outstandingChargesModel))
+          case _ => whatYouOweChargesList
+        }
+      case _ =>
```
avelx, line 88: if we can add error details from FinancialDetailsErrorModel pls
`case Some(error: FinancialDetailsErrorModel) => .. `

app/services/PaymentAllocationsService.scala
```diff
@@ -86,14 +86,14 @@ class PaymentAllocationsService @Inject()(financialDetailsConnector: FinancialDe
...
-        case (_, financialDetails: FinancialDetailsModel) =>
+        case financialDetails: FinancialDetailsModel =>
           financialDetails.documentDetailsWithLpiId(paymentCharge.allocations.head.chargeReference).map {
             documentDetailsWithLpiId =>
               LatePaymentInterestPaymentAllocationDetails(documentDetailsWithLpiId,
                 documentDetailsWithFinancialDetails.documentDetails.head.originalAmount)
           }
         case _ => None
-      }.headOption
+      }
```
avelx, line 96: we might need to be explicit on the error response + need to log error if any

app/controllers/TaxYearSummaryController.scala
danielmaltz1: Have added getFinancialDetailsSingleYear

test/controllers/ChargeSummaryControllerSpec.scala
danielmaltz1: Not needed as we don't group by tax year anymore

app/controllers/ChargeSummaryController.scala
```diff
@@ -79,25 +79,21 @@ class ChargeSummaryController @Inject()(val authActions: AuthActions,
...
-      matchingYear.headOption match {
+      financialResponses match {
         case Some(fdmForTaxYear: FinancialDetailsModel) if fdmForTaxYear.documentDetailsExist(id) =>
           doShowChargeSummary(taxYear, id, isInterestCharge, fdmForTaxYear, paymentsFromAllYears, isAgent, origin, isMFADebit(fdmForTaxYear, id))
         case Some(_: FinancialDetailsModel) =>
           Future.successful(onError(s"Transaction id not found for tax year $taxYear", isAgent, showInternalServerError = false))
         case Some(error: FinancialDetailsErrorModel) =>
           Future.successful(onError(s"Financial details error :: $error", isAgent, showInternalServerError = true))
-        case None =>
+        case _ =>
```
avelx, line 96: worth to keep like " case None" so we can catch any other future changes/ extensions of this pattern matching as compilation errors

app/controllers/HomeController.scala
```diff
@@ -190,9 +187,9 @@ private def mergePaymentsDue(paymentsDue: List[LocalDate], outstandingChargesDue
     .sortWith(_ isBefore _)
     .headOption
 
-private def hasDunningLock(financialDetails: List[FinancialDetailsResponseModel]): Boolean =
+private def hasDunningLock(financialDetails: Option[FinancialDetailsResponseModel]): Boolean =
   financialDetails
-    .collectFirst { case fdm: FinancialDetailsModel if fdm.dunningLockExists => true }
+    .collect { case fdm: FinancialDetailsModel if fdm.dunningLockExists => true }
```
avelx, line 192: .find ( ~) .isDefined

danielmaltz1: I've changed to .exists

app/services/PaymentAllocationsService.scala
```diff
@@ -84,16 +84,18 @@ class PaymentAllocationsService @Inject()(financialDetailsConnector: FinancialDe
...
+        financialDetails.documentDetailsWithLpiId(paymentCharge.allocations.head.chargeReference).map {
+          documentDetailsWithLpiId =>
+            LatePaymentInterestPaymentAllocationDetails(documentDetailsWithLpiId,
+              documentDetailsWithFinancialDetails.documentDetails.head.originalAmount)
+        }
+      case Some(errorModel: FinancialDetailsErrorModel) =>
+        Logger("application").error(s"error when getting financial details - code: ${errorModel.code}, message: ${errorModel.message}")
+        None
+      case None =>
+        None
```
avelx, line 98: can we log the fact that we have no response as a result as well

avelx (dismissed):
approved with comment

app/controllers/HomeController.scala
```diff
@@ -166,56 +166,54 @@ class HomeController @Inject()(val homeView: views.html.Home,
...
-  ) map {
-    case WhatYouOweChargesList(_, _, Some(OutstandingChargesModel(outstandingCharges)), _) =>
-      outstandingCharges.filter(_.isBalancingChargeDebit)
-        .filter(_.relevantDueDate.isDefined)
-    case _ => Nil
-  }
+  private def getDueDates(unpaidCharges: Option[FinancialDetailsResponseModel]): List[LocalDate] =
+    (unpaidCharges collect {
+      case fdm: FinancialDetailsModel => fdm.validChargesWithRemainingToPay.getAllDueDates
+    }).getOrElse(List.empty)
```
avelx, line 176: its looks like we are dropping sorting for some reason, unless this is part of the requirements for current story, its worth to double check

app/controllers/TaxYearsController.scala
```diff
@@ -48,9 +48,10 @@ class TaxYearsController @Inject()(taxYearsView: TaxYears,
                     isAgent: Boolean,
                     origin: Option[String] = None)
                    (implicit user: MtdItUser[_]): Future[Result] = {
+    val maxTaxYears = 5
     Try {
       taxYearsView(
-        taxYears = user.incomeSources.orderedTaxYearsByAccountingPeriods.reverse,
+        taxYears = user.incomeSources.orderedTaxYearsByAccountingPeriods.reverse.take(maxTaxYears),
```
avelx, line 54: I don't think we should move maxTaxYears into config, but worth define a separate private val as we have on line 46 in this controller =>
`private val earliestSubmissionTaxYear = 2023`
and maybe create a separate private method to extract this range of tax years

app/services/FinancialDetailsService.scala
```diff
@@ -60,64 +61,58 @@ class FinancialDetailsService @Inject()(val financialDetailsConnector: Financial
...
-  @deprecated("Use getUnpaidFinancialDetailsV2 instead", "MISUV-8845")
-  def getAllUnpaidFinancialDetails()(implicit user: MtdItUser[_], hc: HeaderCarrier, ec: ExecutionContext): Future[List[FinancialDetailsResponseModel]] = {
-    getAllFinancialDetails.map { chargesWithYears =>
-      chargesWithYears.flatMap {
-        case (_, errorModel: FinancialDetailsErrorModel) => Some(errorModel)
-        case (_, financialDetails: FinancialDetailsModel) =>
-          val unpaidDocDetails: List[DocumentDetail] = financialDetails.unpaidDocumentDetails()
-          if (unpaidDocDetails.nonEmpty) Some(financialDetails.copy(documentDetails = unpaidDocDetails)) else None
-      }
+      val maxYears = 5
```
avelx, line 73: we might need to move this value to the level of FrontendAppConfig, still not sure if we need to make this configurable

app/services/FinancialDetailsService.scala
```diff
@@ -60,64 +61,58 @@ class FinancialDetailsService @Inject()(val financialDetailsConnector: Financial
...
         if (unpaidDocDetails.nonEmpty) Some(financialDetails.copy(documentDetails = unpaidDocDetails)) else None
       case _ => None
     }
   }
+
+  def combineTwoResponses(response1: Option[FinancialDetailsResponseModel],
+                                  response2: Option[FinancialDetailsResponseModel]): Option[FinancialDetailsResponseModel] = {
+    (response1, response2) match {
+      case (Some(validModel1: FinancialDetailsModel), Some(validModel2: FinancialDetailsModel)) => Some(validModel1.mergeLists(validModel2))
+      case (Some(validModel: FinancialDetailsModel), _) => Some(validModel)
```
avelx, line 109: its looks like based on this change we can merge valid and not valid/error model?

app/services/PaymentHistoryService.scala
```diff
@@ -38,40 +38,31 @@ class PaymentHistoryService @Inject()(repaymentHistoryConnector: RepaymentHistor
...
     Logger("application").debug(s"Getting payment history for TaxYears: ${from} - ${to}")
 
-    for {
-      response <- financialDetailsConnector.getPayments(TaxYear.forYearEnd(from), TaxYear.forYearEnd(to))
-    } yield response match {
-      case Payments(payments) => Right(payments.distinct)
-      case PaymentsError(_, _) => Left(PaymentHistoryError)
-    }
+    val maxYears = 5
+    val listOfCalls = orderedTaxYears.grouped(maxYears).toList
```
avelx, line 49: worth to add function which will return list of TaxYearRanges, something like:
user.incomeSources.getOrderedTaxYears[Window/Chunks] or similar

app/services/FinancialDetailsService.scala
danielmaltz1: If we get multiple calls back and one of them is an error/not valid, we should just take the valid one. I think if for example:
We request for 2020-2025 and 2017-2019. 2020-2025 returns data but 2017-2019 returns 404. 
I think we should just take the 2020-2025 response and we don't need to worry about the 404

app/services/CreditService.scala
```diff
@@ -34,40 +34,15 @@ import scala.util.Try
...
-  }
-
-  def getAllCreditsV2(implicit user: MtdItUser[_],
-                      hc: HeaderCarrier): Future[CreditsModel] = {
-
     Logger("application").debug(
       s"Requesting Financial Details for all periods for mtditid: ${user.mtditid}")
 
-    val (from, to) = (user.incomeSources.orderedTaxYearsByYearOfMigration.min, user.incomeSources.orderedTaxYearsByYearOfMigration.max)
+    val from = Math.max(user.incomeSources.orderedTaxYearsByYearOfMigration.min, dateService.getCurrentTaxYearEnd - 4)
```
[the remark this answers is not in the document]
danielmaltz1: This is just saying there's a maximum of 5 years, so if the user migrated less than 5 years ago, it'll choose the year of migration (which is orderedTaxYearsByYearOfMigration.min), but if the user migrated more than 5 years ago it'll choose 5 years ago. I guess this could just be simplified by taking the 5 highest years in the list, I'll simplify it to that

app/models/incomeSourceDetails/IncomeSourceDetailsResponse.scala
```diff
@@ -60,6 +61,11 @@ case class IncomeSourceDetailsModel(nino: String,
     taxYears
   }
 
+  def orderedTaxYearsInWindows(windowWidth: Int)(implicit dateService: DateServiceInterface): List[List[Int]] = {
```
avelx, line 64: you can directly read from appConfig.api1553MaxYears
no actually pass windowWidth as external, this way data will be always looks the same etc

app/services/FinancialDetailsService.scala
```diff
@@ -60,64 +61,57 @@ class FinancialDetailsService @Inject()(val financialDetailsConnector: Financial
...
-          case _ => None
-        }
-    }).map(_.flatten)
-  }
-
-  def getAllFinancialDetailsV2(implicit user: MtdItUser[_],
-                               hc: HeaderCarrier, ec: ExecutionContext): Future[Option[FinancialDetailsResponseModel]] = {
-    Logger("application").debug(
-      s"Requesting Financial Details for all periods for mtditid: ${user.mtditid}")
+    val maxYears = appConfig.api1553MaxYears
```
avelx, line 69: same thing, new variable is not requried

app/services/WhatYouOweService.scala
```diff
@@ -56,33 +57,40 @@ class WhatYouOweService @Inject()(val financialDetailsService: FinancialDetailsS
...
         callOutstandingCharges(mtdUser.saUtr, mtdUser.incomeSources.yearOfMigration, dateService.getCurrentTaxYearEnd).map {
           case Some(outstandingChargesModel) => whatYouOweChargesList.copy(outstandingChargesModel = Some(outstandingChargesModel))
           case _ => whatYouOweChargesList
         }
+      case Some(errorModel: FinancialDetailsErrorModel) =>
+        Logger("application").error(s"error model in unpaid charges - code: ${errorModel.code}, message: ${errorModel.message}")
+        throw new Exception("Error response while getting Unpaid financial details")
+      case None =>
+        val whatYouOweChargesList = WhatYouOweChargesList(BalanceDetails(0.00, 0.00, 0.00, None, None, None, None, None))
+        callOutstandingCharges(mtdUser.saUtr, mtdUser.incomeSources.yearOfMigration, dateService.getCurrentTaxYearEnd).map {
```
avelx, line 88: passing optional UTR and then used it like: -> saUtr.get, maybe not a good idea, same for passing year of migration
yearOfMigration.get... 
callOutstandingCharges method need to be re-design, you can create a tech debt for this or fix in scope of this change

app/services/PaymentHistoryService.scala
```diff
@@ -85,6 +70,16 @@ class PaymentHistoryService @Inject()(repaymentHistoryConnector: RepaymentHistor
       }
     else Future(Right(Nil))
   }
+
+  def combineTwoResponses(response1: Either[PaymentHistoryError.type, Seq[Payment]],
```
avelx, line 74: we can move thins to response level, ie as an function in companion object, but should be fine for now to keep it here

app/services/PaymentHistoryService.scala
```diff
@@ -35,43 +35,28 @@ import scala.concurrent.{ExecutionContext, Future}
...
-        case PaymentsError(status, _) if status == NOT_FOUND => false
-        case PaymentsError(_, _) => true
+    Future.sequence(listOfCalls.map { years =>
+      val (from, to) = (years.min, years.max)
+      Logger("application").debug(s"Getting payment history for TaxYears: ${from} - ${to}")
+
+      for {
+        response <- financialDetailsConnector.getPayments(TaxYear.forYearEnd(from), TaxYear.forYearEnd(to))
+      } yield response match {
+        case Payments(payments) => Right(payments.distinct)
```
avelx, line 52: not sure if we should do any data processing for payments, how "distinct" function would work??

app/services/FinancialDetailsService.scala
avelx: if these are agree business ruleS specified in the story, I am ok with that

avelx (approved):
approved with comments

PR #2743, opened 2025-01-27
MISUV-9068: Build - Penalties and appeals tile

ayubAz:


app/models/homePage/HomePageViewModel.scala
```diff
@@ -63,4 +64,6 @@ case class ReturnsTileViewModel(currentTaxYear: TaxYear, iTSASubmissionIntegrati
 case class YourBusinessesTileViewModel(displayCeaseAnIncome: Boolean, incomeSourcesEnabled: Boolean,
                                        incomeSourcesNewJourneyEnabled: Boolean)
 
-case class AccountSettingsTileViewModel(currentTaxYear: TaxYear, reportingFrequencyEnabled: Boolean, currentYearITSAStatus: ITSAStatus)
\ No newline at end of file
+case class AccountSettingsTileViewModel(currentTaxYear: TaxYear, reportingFrequencyEnabled: Boolean, currentYearITSAStatus: ITSAStatus)
+
+case class PenaltiesAndAppealsTileViewModel(penaltiesAndAppealsEnabled: Boolean)
```
avelx, line 69: Please add "Penalty points" to this view model, should be fine to create a fake service to return fixed value while integration task still in progress;

avelx: also I don't think we need penaltiesAndAppealsEnabled field in the first place, we can make penalties points optional and and create different model depending on the FS status etc

app/controllers/HomeController.scala
```diff
@@ -128,6 +130,8 @@ class HomeController @Inject()(val homeView: views.html.Home,
 
       val accountSettingsTileViewModel = AccountSettingsTileViewModel(currentTaxYear, isEnabled(ReportingFrequencyPage), currentITSAStatus)
 
+      val penaltiesAndAppealsTileViewModel = PenaltiesAndAppealsTileViewModel(maybePenaltyPoints)
```
avelx, line 133: should be fine to return a viewModel right from the service, but it is really your choice how you like to implement this

app/views/helpers/injected/home/PenaltiesAndAppealsTile.scala.html
```diff
@@ -0,0 +1,53 @@
...
+        id = Some("sa-penalties-and-appeals-link"),
+        link = "",
+        messageKey = "home.penaltiesAndAppeals.view"
+    )
+}
+
+@if(penaltiesAndAppealsTileViewModel.penaltyPoints.isDefined) {
+    @home_tile("penalties-and-appeals-tile", messages("home.penaltiesAndAppeals.heading")) {
+        @(penaltiesAndAppealsTileViewModel.penaltyPoints match {
+            case Some(2) =>
```
avelx, line 39: not sure why we are doing this match here, this is threshold which we need be check depending on "Submission frequency" , so you can create a field in the viewModel like:

`case class PenaltiesAndAppealsTileViewModel(penaltyPoints: Option[Int], submittingFrequency: String) {
  val finalPoints: Option[Int] = (submittingFrequency, penaltyPoints) match {
    case ("Annual", Some(points)) if points > 2 => penaltyPoints
    case ("Quarterly",  Some(points)) if points > 4 => penaltyPoints
    case _ => None
  }
}`

avelx (approved):

avelx (approved):

PR #2746, opened 2025-01-29
MISUV-9040 updated auth solution and request models

Chrissystk:
PR contains a new auth solution - the old solution involved (for agents) calling session-data-service and citizen details to get the clients data then calling auth (with a predicate generated based on the client details). The new solution involves calling auth first and checking the user is an agent with arn, then getting the client details as above then making further calls to auth (first one to check if primary and if that fails check if supporting agent).

In addition to this change I have updated the request models so they represent more where in the action flow we are. A lot of the unit tests had to be updated due to this change as we was creating the same model over and over. I have created some default mtdIt users and common audit details in the tests now which to reduce the repetition and reduce the number of places that need to be updated when future changes are made.


app/auth/authV2/actions/AuthoriseAndRetrieveMtdAgent.scala
```diff
@@ -16,85 +16,102 @@
...
+        logger.warn("Bearer Token Timed Out.")
+        Future.successful(Left(Redirect(controllers.timeout.routes.SessionTimeoutController.timeout)))
+      case _: InsufficientEnrolments =>
+        logger.error(s"missing delegated enrolment. Redirect to agent error page.")
+        Future.successful(Left(Redirect(controllers.agent.routes.ClientRelationshipFailureController.show)))
+      case authorisationException: AuthorisationException =>
+        logger.error(s"Unauthorised request: ${authorisationException.reason}. Redirect to Sign In.")
+        Future.successful(Left(Redirect(controllers.routes.SignInController.signIn)))
+      case ex =>
+        logger.error(s"Unexpected error from Auth.")
```
avelx, line 112: if we can log actual error here pls

PR #2747, opened 2025-01-30
MISUV-9040 updated auth solution and request models

Chrissystk:


app/auth/authV2/models/AuthorisedUserRequest.scala
```diff
@@ -0,0 +1,38 @@
...
+case class AuthorisedUserRequest[A](authUserDetails: AuthUserDetails)
+                                   (implicit request: Request[A]) extends WrappedRequest[A](request)
+
+
+case class AuthorisedAgentWithClientDetailsRequest[A](authUserDetails: AuthUserDetails,
+                                                      clientDetails: AgentClientDetails)
+                                                     (implicit request: Request[A]) extends WrappedRequest[A](request)
+
+
+case class AuthorisedAndEnrolledRequest[A](mtditId: String,
```
avelx, line 31: Ideally if we can introduce value class for mtditId and not use String type, but I think we can refactor this later

avelx (approved):

chandraAndem (approved):

PR #2754, opened 2025-02-04
MISUV-9099 Investigate: WhatYouOwe view: remove exceptions

Antony100:


app/models/financialDetails/WhatYouOweChargesList.scala
```diff
@@ -49,12 +50,17 @@ case class WhatYouOweChargesList(balanceDetails: BalanceDetails, chargesList: Li
...
   def interestOnOverdueCharges: Boolean =
     if (overdueChargeList.exists(_.interestOutstandingAmount.isDefined)
       && overdueChargeList.exists(_.latePaymentInterestAmount.getOrElse[BigDecimal](0) <= 0)) true
     else false
 
+  def getRelevantDueDate: LocalDate =
+    outstandingChargesModel.get.bcdChargeType.get.relevantDueDate.getOrElse(throw MissingFieldException("documentRelevantDueDate"))
+
+  def getAciChargeWithTieBreakerChargeAmount: BigDecimal =
+    outstandingChargesModel.get.getAciChargeWithTieBreaker.getOrElse(throw MissingFieldException("documentAciChargeWithTieBreaker")).chargeAmount
```
avelx, line 62: similar situation here, we need to handle situation where there is no value in `getAciChargeWithTieBreaker`

avelx:
@Antony100 - few minor comments, but were are on the right parth

app/models/financialDetails/WhatYouOweChargesList.scala
```diff
@@ -49,12 +50,17 @@ case class WhatYouOweChargesList(balanceDetails: BalanceDetails, chargesList: Li
       && overdueChargeList.exists(_.lpiWithDunningLock.getOrElse[BigDecimal](0) > 0)) true
     else false
 
-
   def interestOnOverdueCharges: Boolean =
     if (overdueChargeList.exists(_.interestOutstandingAmount.isDefined)
       && overdueChargeList.exists(_.latePaymentInterestAmount.getOrElse[BigDecimal](0) <= 0)) true
     else false
 
+  def getRelevantDueDate: LocalDate =
```
avelx, line 58: can we wrap this statement into Try block and throw another exception if there is no value of `bcdChargeType`

app/models/financialDetails/WhatYouOweChargesList.scala
```diff
@@ -49,12 +50,30 @@ case class WhatYouOweChargesList(balanceDetails: BalanceDetails, chargesList: Li
...
     else false
 
-
   def interestOnOverdueCharges: Boolean =
     if (overdueChargeList.exists(_.interestOutstandingAmount.isDefined)
       && overdueChargeList.exists(_.latePaymentInterestAmount.getOrElse[BigDecimal](0) <= 0)) true
     else false
 
+  def getRelevantDueDate: LocalDate = {
+    try {
```
avelx, line 59: we don't have a format agreement in this project what we are going to use, either stick to use Try or we can also use try from time to time; so I guess should be to use this approach

avelx (approved):

PR #2758, opened 2025-02-06
MISUV-5422 Same message with different amounts getting displayed on WYO page

danielmaltz1:
Current mappings (see MISUV-8845) say that both of these values should come from balanceDetails.availableCredit, so have changed both to that

### DIFF
--- a/app/views/WhatYouOwe.scala.html
+++ b/app/views/WhatYouOwe.scala.html
@@ -437,10 +437,9 @@
             @saNoteParagraph
         }
         @whatYouOweChargesList.availableCredit.map { availableCreditAmt =>
-            @if(availableCreditAmt > 0) {
-                @unallocatedCreditParagraphContent(availableCreditAmt.toCurrencyString)
-            }
+            @unallocatedCreditParagraphContent(availableCreditAmt.toCurrencyString)
         }
+
         @if(utr.isDefined && !isAgent && whatYouOweChargesList.getEarliestTaxYearAndAmountByDueDate.isDefined) {
             <div id="payment-button">
             @link(link = controllers.routes.PaymentController.paymentHandoff(whatYouOweChargesList.getEarliestTaxYearAndAmountByDueDate.get._2.toPence, origin).url,
--- a/app/views/WhatYouOwe.scala.html
+++ b/app/views/WhatYouOwe.scala.html
@@ -436,9 +436,9 @@
         @if(!whatYouOweChargesList.isChargesListEmpty) {
             @saNoteParagraph
         }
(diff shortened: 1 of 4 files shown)


app/views/WhatYouOwe.scala.html
```diff
@@ -436,9 +436,9 @@
         @if(!whatYouOweChargesList.isChargesListEmpty) {
             @saNoteParagraph
         }
-        @whatYouOweChargesList.balanceDetails.getAbsoluteUnAllocatedCreditAmount.map { unallocatedCreditAmt =>
-            @if(unallocatedCreditAmt > 0) {
-                @unallocatedCreditParagraphContent(unallocatedCreditAmt.toCurrencyString)
+        @whatYouOweChargesList.balanceDetails.availableCredit.map { availableCreditAmt =>
```
avelx, line 439: same suggestion as per @Daveo-G PR early this week, we should avoid dealing with API structure directly, use what we have already define in the WhatYouOweChargeList 
image

app/views/WhatYouOwe.scala.html
```diff
@@ -436,9 +436,9 @@
...
+            @if(availableCreditAmt > 0) {
+                @unallocatedCreditParagraphContent(availableCreditAmt.toCurrencyString)
```
avelx, line 441: image

avelx: as long as we have separation between success response data and viewModel, it should be fine, there still more refactoring need to be done on this page

danielmaltz1: Oh was this just saying to call the whatYouOweChargesList.availableCredit instead of getting it from the balanceDetails? I've made that change and pushed it

avelx: correct, its a fix and refactoring at the same time

avelx (dismissed):

app/views/WhatYouOwe.scala.html
```diff
@@ -436,9 +436,9 @@
         @if(!whatYouOweChargesList.isChargesListEmpty) {
             @saNoteParagraph
         }
-        @whatYouOweChargesList.balanceDetails.getAbsoluteUnAllocatedCreditAmount.map { unallocatedCreditAmt =>
-            @if(unallocatedCreditAmt > 0) {
-                @unallocatedCreditParagraphContent(unallocatedCreditAmt.toCurrencyString)
+        @whatYouOweChargesList.availableCredit.map { availableCreditAmt =>
+            @if(availableCreditAmt > 0) {
```
avelx, line 440: no need to check if value is greater than zero

avelx (approved):

PR #2759, opened 2025-02-06
MISUV-8352 - CreditSummaryView: remove exception

Antony100:


app/models/financialDetails/BalanceDetails.scala
```diff
@@ -35,6 +36,10 @@ case class BalanceDetails(balanceDueWithin30Days: BigDecimal,
     total - firstPendingAmountRequested.getOrElse(0) - secondPendingAmountRequested.getOrElse(0)
   }
 
+  def getAvailableCredit: BigDecimal = {
```
avelx, line 39: I don't think we need this change, we already have these defined in WhatYouOweChargesList or we should define in other viewModels as a part of future refactoring;

test/models/financialDetails/BalanceDetailsSpec.scala
```diff
@@ -107,5 +108,20 @@ class BalanceDetailsSpec extends UnitSpec {
         balanceDetailsModel(unallocatedCredit = Some(-1000.toDouble)).getAbsoluteUnAllocatedCreditAmount shouldBe  Some(1000.toDouble)
       }
     }
+
```
avelx, line 111: no need to test this method as its not required

app/models/financialDetails/BalanceDetails.scala
```diff
@@ -16,6 +16,7 @@
 
 package models.financialDetails
 
+import exceptions.MissingFieldException
```
avelx, line 19: can we drop this line as this import is not required

test/models/financialDetails/BalanceDetailsSpec.scala
```diff
@@ -16,6 +16,7 @@
 
 package models.financialDetails
 
+import exceptions.MissingFieldException
```
avelx, line 19: can we drop this line as this import is not required

app/views/CreditsSummary.scala.html
```diff
@@ -120,7 +120,7 @@
         @if(isAgent) {
             @p(id = Some("p-money-in-your-account")) {
                 @messages("credits.money-in-your-account-section.agent.content", availableCredit.toCurrencyString)
-                @* @messages("credits.money-in-your-account-section.agent.content", balanceDetails.availableCredit.getOrElse(throw MissingFieldException("Available Credit")).toCurrencyString)) *@
+                @* @messages("credits.money-in-your-account-section.agent.content", whatYouOweChargesList.availableCredit.toCurrencyString)) *@
```
avelx, line 123: we are actually need to drop this line as its commented code

avelx (approved):

PR #2760, opened 2025-02-10
MISUV-8819 updated and added new audits for supporting agents

Chrissystk:


app/audit/models/ConfirmClientDetailsAuditModel.scala
```diff
@@ -32,7 +36,9 @@ case class ConfirmClientDetailsAuditModel(clientName: String, nino: String,
     "mtditid" -> mtditid,
     "saUtr" -> saUtr,
     "agentReferenceNumber" -> arn,
-    "userType" -> "Agent") ++
+    "userType" -> "Agent",
```
avelx, line 39: worth to bundle thins into one single field, but depends on the requirements I guess

avelx (approved):
approved with comments

PR #2767, opened 2025-02-13
MISUV-9139 Move Phase Banner into HMRC Header component/<header> to fix a11y issues 

michael-hmrc:


TomRafferty (approved):
looks good, all tests pass and we are down 1 violation 😄 ticket updated with a comment and pics

PR #2787, opened 2025-03-05
MISUV-9307 - Refactoring RepaymentHistoryUtils.creditEntry

BAmiriDev:


app/models/repaymentHistory/RepaymentHistoryUtils.scala
```diff
@@ -133,14 +133,22 @@ object RepaymentHistoryUtils {
     )
   }
 
-  private def creditEntry(payment: Payment, isAgent: Boolean, isPoaReconciliationCredit: Boolean = false)(implicit messages: Messages, dateServiceInterface: DateServiceInterface): PaymentHistoryEntry = {
-    val creditType = payment.creditType.getOrElse(throw MissingFieldException("Credit type"))
-    PaymentHistoryEntry(
-      date = payment.dueDate.getOrElse(throw MissingFieldException(s"Payment Due Date - ${creditType.getClass.getSimpleName}")),
+  private def creditEntry(payment: Payment, isAgent: Boolean, isPoaReconciliationCredit: Boolean = false)(implicit messages: Messages, dateServiceInterface: DateServiceInterface): Either[Throwable, PaymentHistoryEntry] = {
+    for {
+      creditType <- payment.creditType.toRight(MissingFieldException("Credit type"))
```
avelx, line 138: * We are suppressing anything in the Left side in the calling method => 
**filterPayment =**> 
creditEntry(payment, isAgent).toOption
* What we should do instead, is to convert **filterPayment** to return **Either[Throwable, PaymentHistoryEntry]** as well
and handle result in combinePaymentHistoryData (log error and throw error)

app/models/repaymentHistory/RepaymentHistoryUtils.scala
```diff
@@ -94,20 +100,20 @@ object RepaymentHistoryUtils {
...
-      case (false, _, true, Some(PaymentType))                                 => Some(paymentToHMRCEntry(payment, isAgent))
-      case (_, _, _, _)                                                        => None
+      case (true, _, _, Some(MfaCreditType))                                   => Right(mfaCreditEntry(payment, isAgent))
+      case (true, _, _, Some(CutOverCreditType))                               => creditEntry(payment, isAgent)
+      case (true, true, _, Some(PoaOneReconciliationCredit))  => creditEntry(payment, isAgent, true)
+      case (true, true, _, Some(PoaTwoReconciliationCredit))  => creditEntry(payment, isAgent, true)
+      case (true, _, _, Some(BalancingChargeCreditType))                       => creditEntry(payment, isAgent)
+      case (true, _, _, Some(RepaymentInterest))                               => creditEntry(payment, isAgent)
+      case (false, _, true, Some(PaymentType))                                 => Right(paymentToHMRCEntry(payment, isAgent))
+      case (_, _, _, _)                                                        => Left(MissingFieldException("Invalid Payment Data"))
```
avelx, line 116: In case we are promoting this case to be an error,  I guess worth to log actual data / payment entity like on line 121;
but not necessarily we have to use prettyPrint:
`Logger("application").info("json:" + Json.prettyPrint(Json.toJson(payment))) `

avelx (approved):

PR #2793, opened 2025-03-11
MISUV-7919: Investigate: fix failing UTs / ITs when read-from-mongo key is on

ayubAz:


test/testUtils/TestSupport.scala
```diff
@@ -290,8 +294,21 @@ trait TestSupport extends UnitSpec with GuiceOneAppPerSuite with BeforeAndAfterA
     }
   }
 
-  def disableAllSwitches(): Unit = {
-    allFeatureSwitches.foreach(switch => disable(switch))
-  }
-
+  def disableAllSwitches(): Unit =
+    if (appConfig.readFeatureSwitchesFromMongo)
+      Await.result(featureSwitchRepository.setFeatureSwitches(allFeatureSwitches.map(_ -> false).toMap), 5.seconds)
```
avelx, line 299: :+1:

avelx:
@ayubAz  - I would suggest to discuss with QAs how we are going to check this in the pipeline from time to time, maybe worth to assign value to read-from-mongo key randomly? with [75/25 => false|true] distribution , up to discussion;

conf/application.conf
```diff
@@ -188,7 +188,7 @@ feature-switch {
 #  In the Staging Environment, this will be flipped to true
 #  Therefore changing the state of FS(s) will be easier during demos
 feature-switches {
-  read-from-mongo = false
+  read-from-mongo = true
```
avelx, line 191: if we can keep this key set to false as before

avelx (approved):

avelx (approved):

PR #2798, opened 2025-03-12
MISUV-9322 using h1 component rather than <h1

Chrissystk:


app/views/incomeSources/add/IncomeSourceReportingMethod.scala.html
```diff
@@ -113,7 +114,7 @@
         <li id="bullet-3">@Html(messages("incomeSources.add.incomeSourceReportingMethod.description4.bullet3"))</li>
     </ul>
 
-    @h1(messages("incomeSources.add.incomeSourceReportingMethod.chooseReport"), classes = "govuk-heading-m")
+    @h2(messages("incomeSources.add.incomeSourceReportingMethod.chooseReport"))
```
avelx, line 117: any reason why we are changing H1 -> H2 ?

Chrissystk: h1 should only be for the main heading this is a subheading

avelx (approved):
approved with comments

PR #2807, opened 2025-03-17
MISUV-9338 - fixed incorrect message on home page next updates tile

TomRafferty:


chandraAndem (changes requested):
We need to fix ATs too which are breaking because of this change.

chandraAndem (approved):

PR #2810, opened 2025-03-17
MISUV-9201: Summary pages for each of the charge types

ayubAz:


app/views/partials/yourSelfAssessmentCharges/LatePaymentInterestChargeMessage.scala.html
```diff
@@ -0,0 +1,62 @@
...
+        val link = if(viewModel.isAgent) {
+            controllers.routes.PaymentAllocationsController.viewPaymentAllocationAgent(matchingPayment.get).url
+        } else {
+            controllers.routes.PaymentAllocationsController.viewPaymentAllocation(matchingPayment.get, None).url
+        }
+        val linkText = if (viewModel.chargeItem.transactionType == MfaDebitCharge) messages("chargeSummary.paymentAllocations.mfaDebit") else
+            allocation.getPaymentAllocationTextInChargeSummary.map(msg => {
+                messages(msg, viewModel.taxYearFromCodingOut, viewModel.taxYearToCodingOut)
+            }).getOrElse(
+                throw MissingFieldException("Allocation link text missing")
```
avelx, line 56: Please raise a tech debt for any exception thrown, so we can cover this at some point

app/views/partials/yourSelfAssessmentCharges/PaymentHistoryTable.scala.html
```diff
@@ -0,0 +1,173 @@
...
+            allocation <- viewModel.paymentAllocations
+            payment    <- allocation.allocations
+            if !viewModel.latePaymentInterestCharge
+        } yield {
+            Seq(
+                TableRow(
+                    content = HtmlContent(latePaymentInterestChargeMessage(viewModel, allocation, payment).toString)
+                ),
+                TableRow(
+                    content = HtmlContent(payment.dueDate.getOrElse(throw MissingFieldException("Payment Date")).toLongDate.toNonBreaking)
```
avelx, line 162: same here, if we can create a story to address these exceptions thrown

app/views/partials/yourSelfAssessmentCharges/PaymentHistoryTable.scala.html
```diff
@@ -0,0 +1,173 @@
...
+                content = HtmlContent(messages("chargeSummary.chargeHistory.date"))
+            ),
+            HeadCell(
+                content = Text(messages("chargeSummary.chargeHistory.amount")),
+                format = Some("numeric")
+            )
+        )
+    ),
+    rows = Seq(
+        if(viewModel.noInterestChargeAndNoCodingOutEnabledWithIsPayeSelfAssessment)
```
avelx, line 79: might worth to refactor this large segment of code where we joining results with ++;
or I guess to create a tech debt story if we are in the rush to push this part of the work

app/views/partials/yourSelfAssessmentCharges/InterestOnYourBalancingPaymentTable.scala.html
```diff
@@ -0,0 +1,91 @@
...
+@import java.time.LocalDate
+
+@this(
+    govukTable: GovukTable,
+    implicitDateFormatter: ImplicitDateFormatterImpl,
+    chargeSummaryPaymentAllocation: ChargeSummaryPaymentAllocation,
+    h2: h2
+)
+
+@(viewModel: ChargeSummaryViewModel)(implicit messages: Messages)
```
avelx, line 34: might need to have a separate view model + with a different consistent name? I would assume we are sharing ChargeSummaryViewModel with a the old view?

app/views/partials/yourSelfAssessmentCharges/LatePaymentInterestChargeMessage.scala.html
```diff
@@ -0,0 +1,62 @@
...
+@import implicits.ImplicitDateFormatterImpl
+@import implicits.HtmlFormatter.NbspString
+@import _root_.implicits.ImplicitCurrencyFormatter.CurrencyFormatter
+
+@this(
+        implicitDateFormatter: ImplicitDateFormatterImpl,
+        p: p
+)
+
+@(viewModel: ChargeSummaryViewModel, allocation: PaymentHistoryAllocations, payment: PaymentHistoryAllocation)(implicit messages: Messages)
```
avelx, line 34: same here, if we should rely on the different view model here

app/views/partials/yourSelfAssessmentCharges/PaymentHistoryTable.scala.html
```diff
@@ -0,0 +1,173 @@
...
+@import models.financialDetails.PoaTwoDebit
+
+@this(
+        govukTable: GovukTable,
+        implicitDateFormatter: ImplicitDateFormatterImpl,
+        latePaymentInterestChargeMessage: LatePaymentInterestChargeMessage,
+        h2: h2
+)
+
+@(viewModel: ChargeSummaryViewModel)(implicit messages: Messages)
```
avelx, line 37: same here / should be create a new viewModel ?

avelx (approved):
approved with comments
more UTs/ ITs need to be added

app/controllers/ChargeSummaryController.scala
```diff
@@ -191,9 +192,19 @@ class ChargeSummaryController @Inject()(val authActions: AuthActions,
           poaOneChargeUrl = poaOneChargeUrl,
           poaTwoChargeUrl = poaTwoChargeUrl
         )
+
+        def isChargeAllowed(chargeItem: ChargeItem) =
```
avelx, line 196: this is something we can move on the ChargeItem level I guess

app/views/helpers/ChargeName.scala
```diff
@@ -0,0 +1,34 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package views.helpers
+
+import models.financialDetails.{BalancingCharge, PoaOneDebit, PoaTwoDebit, TransactionType}
+import play.api.i18n.Messages
+
```
avelx, line 21: I think we should start unifying Lang specific converters: Type -> String etc
Maybe name like  ChargeNameLangHelper would be more intuitive, not sure, worth to create a tech story to start moving similar code under one namespace/helper?

app/views/partials/yourSelfAssessmentCharges/PaymentHistoryTable.scala.html
avelx: we can do this before passing view model into the scope of the view ie; or say create a new property in the view model etc; the main reason to mention this was that we have to reduce this complexity in order to make future maintenance much easier + we have to consider that we would need to extend this view level logic if requirements would change; might worth to create a tech debt ? really up to you

avelx (approved):

avelx (approved):
approve with comments

app/views/partials/yourSelfAssessmentCharges/InterestOnYourChargeTable.scala.html
```diff
@@ -0,0 +1,92 @@
...
+    rows = Seq(
+        Seq(
+            TableRow(
+                content = Text(viewModel.chargeItem.outstandingAmount.toCurrencyString)
+            ),
+            TableRow(
+                content = Text(getMessage("daysOverdue", viewModel.chargeItem.getInterestFromDate.toLongDate, viewModel.chargeItem.getInterestEndDate.toLongDate))
+            ),
+            TableRow(
+                content = Text(viewModel.chargeItem.getInterestRate.toString + "%"),
```
avelx, line 52: if we can create a formatting method Or use existing one to avoid add "%" at the end; really optionalI would say;
similar to what we have for "toCurrencyString" etc

avelx (approved):
approved with comments

conf/messages.cy
```diff
@@ -2518,4 +2518,50 @@ reporting.frequency.latency.details.section.text.para2.bullet3      = hyd yn oed
 reporting.frequency.latency.details.section.text.para3.heading      = Sut i newid amlder eich adroddiadau ar gyfer ffynhonnell incwm newydd
 reporting.frequency.latency.details.section.text.para3.text.1       = Neges testun. Gallwch wneud hyn ar unrhyw adeg o’r adran
 reporting.frequency.latency.details.section.text.para3.link.text    = pob busnes
-reporting.frequency.latency.details.section.text.para3.text.2       =
\ No newline at end of file
+reporting.frequency.latency.details.section.text.para3.text.2       =
+
+yourSelfAssessmentChargeSummary.tax-year                            = {0} to {1} tax year
```
avelx, line 2523: I would assume we are still awaiting Welsh translation ?

test/controllers/ChargeSummaryControllerSpec.scala
```diff
@@ -60,6 +60,130 @@ class ChargeSummaryControllerSpec extends ChargeSummaryControllerHelper {
           testSupportingAgentDeniedAccess(action(id1040000123))(fakeRequest)
         } else {
           "render the charge summary page" when {
+            "charge history & your self assessment charges feature switch is enabled and there is a user" that {
+              "provided with an id associated to a POA1 Debit" in new Setup(
+                financialDetailsModelWithPoaOneAndTwo()) {
+                enable(ReviewAndReconcilePoa)
```
avelx, line 66: would like nice to have a enable method where we can pass variable number of argument, as this would be a step forward I think

avelx (approved):
approve with comments

PR #2813, opened 2025-03-20
MISUV-9284 - Charges due in 30 days tab

BAmiriDev:


app/controllers/YourSelfAssessmentChargesController.scala
```diff
@@ -66,8 +66,10 @@ class YourSelfAssessmentChargesController @Inject()(val authActions: AuthActions
 
       val hasOverdueCharges: Boolean = whatYouOweChargesList.chargesList.exists(_.isOverdue()(dateService))
       val hasAccruingInterestReviewAndReconcileCharges: Boolean = whatYouOweChargesList.chargesList.exists(_.isNotPaidAndNotOverduePoaReconciliationDebit()(dateService))
+      val hasChargesDueWithin30Days: Boolean = whatYouOweChargesList.chargesDueWithin30DaysList.nonEmpty
```
avelx, line 69: should be fine to move this one on the level of WhatYouOweChargeList type, but really up to you if you keen to make this more concise

app/models/financialDetails/YourSelfAssessmentChargesViewModel.scala
```diff
@@ -23,6 +23,7 @@ import services.DateServiceInterface
 import java.time.LocalDate
 
 case class YourSelfAssessmentChargesViewModel(hasOverdueOrAccruingInterestCharges: Boolean,
+                                              hasChargesDueWithin30Days: Boolean,
```
avelx, line 26: this can be resolved by what we have in  `whatYouOweChargesList: WhatYouOweChargesList ` ?

app/views/YourSelfAssessmentCharges.scala.html
```diff
@@ -348,16 +393,60 @@ <h2 class="govuk-tabs__title">
...
+                                    </div>
+                                    }
+                                </div>
+                                }
+
                                 @if(ChargeItem.overdueOrAccruingInterestChargeList(viewModel.whatYouOweChargesList).nonEmpty || viewModel.whatYouOweChargesList.overdueOutstandingCharges.nonEmpty) {
                                     <div class="govuk-tabs__panel" id="charges-due-now">
                                         <table class="govuk-table" id="charges-due-now-table">
-                                            @tableHead("over-due-payments-table-head")
+                                            @tableHead("over-due-payments-table-head", true)
```
avelx, line 449: in case of `showInterest` flag is always true, maybe we can drop it all as its not required? I don't see any case were its set externally

app/views/YourSelfAssessmentCharges.scala.html
```diff
@@ -285,8 +308,17 @@
                 target = Some("_blank"),
                 outerMessage = ".")
         </p>
-    }
 
+    @if(viewModel.claimToAdjustViewModel.adjustPaymentsOnAccountFSEnabled) {
+        <p class="govuk-body" id="adjust-payments-on-account">
+            @link(
+                link = controllers.claimToAdjustPoa.routes.AmendablePoaController.show(user.isAgent()).url,
+                messageKey = getMessage("adjust-poa", "2024", "2025"),
```
avelx, line 316: not sure if this is a good idea for message to be resolved by specific tax years / calendar years

app/views/YourSelfAssessmentCharges.scala.html
```diff
@@ -348,16 +393,60 @@ <h2 class="govuk-tabs__title">
...
+                                        </div>
+                                        <br>
+                                        <div>
+                                            @getMessage("overdue-inset-text-2")
+                                        </div>
+                                    </div>
+
+                                    @if(user.saUtr.isDefined && !user.isAgent() && viewModel.whatYouOweChargesList.getEarliestTaxYearAndAmountByDueDate.isDefined) {
+                                    <div id="payment-button">
+                                        @link(link = controllers.routes.PaymentController.paymentHandoff(viewModel.whatYouOweChargesList.getEarliestTaxYearAndAmountByDueDate.get._2.toPence, origin).url,
```
avelx, line 438: Please avoid use of .get and/o .head in the views when resolving data provided by viewModel;
I would suggest to add next property to the viewModel and throw and error on the viewModel construction phase:
`val earliestTaxYearAndAmountByDueDate : (Int, BigDecimal) `

+ _might worth to use TaxYear type instead of the Int?_

conf/messages.cy
```diff
@@ -899,6 +899,7 @@ selfAssessmentCharges.important                                    = Important
 selfAssessmentCharges.important-p1                                 = This is the total amount of overdue tax you need to pay. Any interest will increase daily until the overdue amount is paid.
 selfAssessmentCharges.important-p2                                 = You can
 selfAssessmentCharges.important-p2-link-text                       = pay the total amount needed now.
+selfAssessmentCharges.charges-due-in-30-days                       = Charges due in 30 days
```
avelx, line 902: looks like we missed Welsh at some point, as this is an Welsh lang resource file; Or we are still awaiting for Welsh team reply??

avelx:
@BAmiriDev  - I would suspect we would need a bit more test coverage around ViewSpect / ITs tests; if you can double check this part pls;

conf/messages.cy
BAmiriDev: Rob is still finalising content; I believe Welsh translations will be handled in a future story. I added these more as a placeholder so the test passes

avelx: In this case we need to create a sub-task related to current story , in case this part would slip into the next sprint we would have to create a story to track this change in the future

app/views/YourSelfAssessmentCharges.scala.html
BAmiriDev: The showInterest flag is used in @tableHead("due-in-30-days-payments-table-head", false) because the "Due in 30 days" tab doesn’t have estimated interest

app/models/financialDetails/YourSelfAssessmentChargesViewModel.scala
```diff
@@ -29,6 +29,7 @@ case class YourSelfAssessmentChargesViewModel(hasOverdueOrAccruingInterestCharge
                                               dunningLock: Boolean,
                                               reviewAndReconcileEnabled: Boolean,
                                               creditAndRefundEnabled: Boolean,
+                                              earliestTaxYearAndAmountByDueDate: (TaxYear, BigDecimal),
```
avelx, line 32: if we can avoid using tuples in this fashion as this would generate "not readable code";
feel free to create a new case class if you need to in order to improve code readability

app/controllers/YourSelfAssessmentChargesController.scala
```diff
@@ -66,13 +67,19 @@ class YourSelfAssessmentChargesController @Inject()(val authActions: AuthActions
 
       val hasOverdueCharges: Boolean = whatYouOweChargesList.chargesList.exists(_.isOverdue()(dateService))
       val hasAccruingInterestReviewAndReconcileCharges: Boolean = whatYouOweChargesList.chargesList.exists(_.isNotPaidAndNotOverduePoaReconciliationDebit()(dateService))
+      val earliestTaxYearAndAmount: (TaxYear, BigDecimal) =
```
avelx, line 70: same here, fee free to create a new case class if you need to as tuples  use is generally discouraged; if this is something used within a function and not exposed outside of it, than this should be fine

app/controllers/YourSelfAssessmentChargesController.scala
```diff
@@ -66,13 +68,18 @@ class YourSelfAssessmentChargesController @Inject()(val authActions: AuthActions
 
       val hasOverdueCharges: Boolean = whatYouOweChargesList.chargesList.exists(_.isOverdue()(dateService))
       val hasAccruingInterestReviewAndReconcileCharges: Boolean = whatYouOweChargesList.chargesList.exists(_.isNotPaidAndNotOverduePoaReconciliationDebit()(dateService))
+      val earliestTaxYearAndAmount: Option[TaxYearWithAmount] =
```
avelx, line 71: :+1:

avelx (dismissed):
approved with comments

avelx (approved):

PR #2815, opened 2025-03-21
MISUV-8598 Amendment for quarterly+quarterly case

jmountain-hmrc:


test/views/manageBusinesses/add/IncomeSourceAddedObligationsViewSpec.scala
```diff
@@ -886,10 +886,9 @@ class IncomeSourceAddedObligationsViewSpec extends ViewSpec {
 
       "It is reporting quarterly" in new Setup(validCurrentTaxYearQuarterlyCallNoOverdue) {
         Option(document.getElementById("change-frequency")) match {
-          case Some(changeFrequency) =>
-            changeFrequency.text() shouldBe "Depending on your circumstances, you may be able to view and change your reporting frequency for all your businesses."
-            changeFrequency.select("a").attr("href") shouldBe nextUpdatesUrl
-          case None => fail("No upcoming updates link was found")
+          case Some(_) =>
+            fail("Upcoming updates link was found")
```
chandraAndem, line 890: Could you update the test case to fail("view or change your reporting frequency link is found") as this test is not specific to Upcoming Updates link? 
This applies to other old test in this scenario as well as it is misleading.

chandraAndem (changes requested):

PR #2826, opened 2025-03-27
MISUV-9294 Build - Summary pages Late Submission Penalty

danielmaltz1:


app/audit/models/ChargeSummaryAudit.scala
```diff
@@ -34,7 +34,9 @@ case class ChargeSummaryAudit(mtdItUser: MtdItUser[_],
                               paymentAllocations: List[PaymentHistoryAllocations],
                               isLatePaymentCharge: Boolean,
                               isMFADebit: Boolean = false,
-                              taxYear: Int
+                              taxYear: Int,
```
avelx, line 37: If we can use TaxYear type here instead of Int?

app/audit/models/ChargeSummaryAudit.scala
```diff
@@ -34,7 +34,9 @@ case class ChargeSummaryAudit(mtdItUser: MtdItUser[_],
...
+                              reviewAndReconcileEnabled: Boolean,
```
avelx, line 38: this is the first ever audit event which we driver/execute by providing a feature switch status, why we need to pass it to audit event level?

test/controllers/ChargeSummaryControllerSpec.scala
```diff
@@ -183,6 +177,28 @@ class ChargeSummaryControllerSpec extends ChargeSummaryControllerHelper {
                 document.getElementById("interest-on-your-charge-table").getAllElements.size().equals(0) shouldBe false
                 document.getElementById("payment-history-table").getElementsByTag("caption").text() shouldBe "Balancing payment history"
               }
+              "provided with an id associated to a Late Submission Penalty" in new Setup(testValidFinancialDetailsModelWithLateSubmissionPenalty) {
+                enable(ReviewAndReconcilePoa)
```
avelx, line 181: if we can create an overload version of enable method where we can pass variable list of arguments

app/audit/models/ChargeSummaryAudit.scala
danielmaltz1: This audit model uses ChargeItem.getChargeTypeKey for some of the content, which takes feature switches. Otherwise if we have the reviewAndReconcile/penalties FS turned on and go to the Charge Summary for one of those types of charges, the audit will say "unknown charge"

avelx: I think these charges would not be available on relevant page based on the FS i, so I would think that these check / flags not required at all

app/models/financialDetails/TransactionItem.scala
```diff
@@ -47,7 +47,7 @@ trait TransactionItem {
   }
 
   // TODO: duplicate logic, in scope of => https://jira.tools.tax.service.gov.uk/browse/MISUV-8557
-  def getChargeTypeKey(reviewAndReconcileEnabled: Boolean = false, penaltiesEnabled: Boolean = false): String =
+  def getChargeTypeKey: String =
```
avelx, line 50: Just wondering if we can move this under ChargeNameLangHelper as a separate method?

app/views/partials/yourSelfAssessmentCharges/PaymentHistoryTable.scala.html
```diff
@@ -54,6 +55,7 @@
         case PoaOneDebit     => getMessage("chargeHistory.paymentOnAccount1.heading")
         case PoaTwoDebit     => getMessage("chargeHistory.paymentOnAccount2.heading")
         case BalancingCharge => getMessage("chargeHistory.balancingPayment.heading")
+        case LateSubmissionPenalty => getMessage("chargeHistory.lateSubmissionPenalty.heading")
```
avelx, line 58: same here, if we can move this under something like 
ChargeNameLangHelper.scala

app/models/financialDetails/TransactionItem.scala
danielmaltz1: It is used in quite a few places, and would need to be called like this in the views:
ChargeNameLangHelper.getChargeTypeKey(viewModel.chargeItem.transactionType, viewModel.chargeItem.subTransactionType)
I'm not sure if this is worth it

avelx: as discussed: will be covered by 
https://jira.tools.tax.service.gov.uk/browse/MISUV-8557

avelx (dismissed):

avelx (approved):

dnyuar (approved):

PR #2839, opened 2025-04-10
[MISUV-9583] Adding Accounting Journey FS

Mbrookesy:


conf/application.conf
```diff
@@ -166,6 +166,7 @@ feature-switch {
   enable-adjust-payments-on-account = true
   enable-charge-history = false
   enable-credits-refunds-repay = false
+  enable-display-accounting-method-row = false
```
avelx, line 169: If we can update confluence as well 
https://confluence.tools.tax.service.gov.uk/pages/viewpage.action?spaceKey=MISUV&title=Feature+Switches+State

Mbrookesy: Sure i'll get that updated now

avelx: we might need to introduce some kind of the sbt ~ plugging to keep this documentation in code? as this is getting tedious to track all these changes

PR #2842, opened 2025-04-14
[MISUV-9594] Adding guidance link to supporting agents front page

Mbrookesy:
The link itself isn't live yet so clicking it currently just leads to a 404.

### DIFF
--- a/conf/messages
+++ b/conf/messages
@@ -179,6 +179,7 @@ home.agent.supporting.cannotAccess.payments                     = payments, cred
 home.agent.supporting.cannotAccess.returns                      = returns
 home.agent.supporting.cannotAccess.nextChargesDue               = next charges due
 home.agent.supporting.readMore                                  = Read more about the difference between main and supporting agents on GOV.UK (opens in new tab).
+home.agent.supporting.readMore.link                             = https://www.gov.uk/guidance/choose-agents-for-making-tax-digital-for-income-tax
 
 home.penaltiesAndAppeals.heading                                = Penalties and appeals
 home.penaltiesAndAppeals.view                                   = Check Self Assessment penalties and appeals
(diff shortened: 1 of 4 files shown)


Mbrookesy:
rebuild

conf/messages
```diff
@@ -179,6 +179,7 @@ home.agent.supporting.cannotAccess.payments                     = payments, cred
 home.agent.supporting.cannotAccess.returns                      = returns
 home.agent.supporting.cannotAccess.nextChargesDue               = next charges due
 home.agent.supporting.readMore                                  = Read more about the difference between main and supporting agents on GOV.UK (opens in new tab).
+home.agent.supporting.readMore.link                             = https://www.gov.uk/guidance/choose-agents-for-making-tax-digital-for-income-tax
```
michael-hmrc, line 182: can this be moved to application config and passed in / handled like the compatible software guidance url? Helps keep the messages cleaner and we have can update the url via config changes instead of raising prs for changes to the message file + deploy etc.

Mbrookesy: How would we do that when we have both a welsh and english link

michael-hmrc: okay not 100% sure this would work but would we be able to swap between welsh and english urls based on logic like this? As in we have 2 urls in config?

```
        <div id="language-switch">
            @languageSwitch(LanguageSelect(
                if (messages.lang.code == "cy") Cy else En,
                (En, controllers.routes.LocalLanguageController.switchToLanguage("english").url),
                (Cy, controllers.routes.LocalLanguageController.switchToLanguage("cymraeg").url)
            ))
        </div>
```

specifically the 
```
                if (messages.lang.code == "cy") Cy else En,
```

Or do you think that would be more messy?

michael-hmrc: nevermind please ignore above comments about moving to config, decision to leave as is.

michael-hmrc (approved):

PR #2862, opened 2025-04-24
MISUV-9599: Update AddBusinessName view and spec to reflect revised hint text

IoannisLazaridis-oc:
- [x] Updated business name guideline from paragraph to hint text based on new requirements
~~Renamed related variables for clarity across codebase~~

Screenshot:
image

---
Ticket: [MISUV- 9599](https://jira.tools.tax.service.gov.uk/browse/MISUV-9599)

### DIFF
--- a/app/views/manageBusinesses/add/AddBusinessName.scala.html
+++ b/app/views/manageBusinesses/add/AddBusinessName.scala.html
@@ -71,13 +71,13 @@ btaNavPartial = user.btaNavPartial) {
         @h1(messages("add-business-name.heading1"))
 
         @p(classes="govuk-body") {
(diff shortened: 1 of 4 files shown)


Mbrookesy (changes requested):
Would it be possible to adjust the size of the "Business Name" heading so the h1 is using the "l" argument for the size instead of "xl". The heading looks big compared to the rest of the questions in the journey plus the official component for complex questions uses an "l" heading https://design-system.service.gov.uk/patterns/question-pages/explanatory-text/ 

Also your changes have broken the old income source journey which use the same message keys
image

Mbrookesy (approved):

PR #2882, opened 2025-05-14
MISUV-9586: Save mandation status is session cookie

nemanjatomanovic:


app/controllers/HomeController.scala
```diff
@@ -71,8 +72,10 @@ class HomeController @Inject()(val homeView: views.html.Home,
   }
 
   def handleShowRequest(origin: Option[String] = None)
-                       (implicit user: MtdItUser[_], hc: HeaderCarrier): Future[Result] =
-    nextUpdatesService.getDueDates().flatMap {
+                       (implicit user: MtdItUser[_], hc: HeaderCarrier): Future[Result] = {
+    val test = nextUpdatesService.getDueDates()
```
avelx, line 76: if we can revert this name change to original or more meaningful

nemanjatomanovic: All done. Missed it while testing, good catch mate !

app/controllers/HomeController.scala
```diff
@@ -227,7 +234,7 @@ private def handleErrorGettingDueDates(isAgent: Boolean)(implicit user: MtdItUse
 }
 
   private def getCurrentITSAStatus(taxYear: TaxYear)(implicit user: MtdItUser[_]): Future[ITSAStatus.ITSAStatus] = {
-    ITSAStatusService.getStatusTillAvailableFutureYears(taxYear.previousYear).map(_.view.mapValues(_.status)
+    ITSAStatusService.getStatusTillAvailableFutureYears(taxYear.previousYear, user.nino).map(_.view.mapValues(_.status)
```
avelx, line 237: I would suspect we need to save mandation status in session cookie for supporting agents as well?

app/controllers/agent/ConfirmClientUTRController.scala
```diff
@@ -39,7 +39,8 @@ import scala.concurrent.{ExecutionContext, Future}
 class ConfirmClientUTRController @Inject()(confirmClient: confirmClient,
                                            val authActions: AuthActions,
                                            val auditingService: AuditingService,
-                                           val sessionDataService: SessionDataService)
+                                           val sessionDataService: SessionDataService,
+                                           val ITSAStatusService: ITSAStatusService)
```
avelx, line 43: if we can revert this to original version as we don't need ITSAStatusService within this controller as its not in use?

app/controllers/agent/EnterClientsUTRController.scala
```diff
@@ -45,7 +46,8 @@ class EnterClientsUTRController @Inject()(enterClientsUTR: EnterClientsUTR,
                                           clientDetailsService: ClientDetailsService,
                                           val authorisedFunctions: FrontendAuthorisedFunctions,
                                           val authActions: AuthActions,
-                                          val auditingService: AuditingService)
+                                          val auditingService: AuditingService,
+                                          val ITSAStatusService: ITSAStatusService)
```
avelx, line 50: if we can revert change to original as we are not using ITSAStatusService in this controller

app/controllers/incomeSources/add/IncomeSourceReportingMethodController.scala
```diff
@@ -115,7 +115,7 @@ class IncomeSourceReportingMethodController @Inject()(val authActions: AuthActio
           errorHandler(isAgent).showInternalServerError()
         }
       case true =>
-        itsaStatusService.hasMandatedOrVoluntaryStatusCurrentYear.flatMap {
+        itsaStatusService.hasMandatedOrVoluntaryStatusCurrentYear(user.nino).flatMap {
```
avelx, line 118: these changes need to be confirm with someone from Solar team, as we not own this part of sub-domain

app/models/sessionData/SessionCookieData.scala
```diff
@@ -35,8 +35,7 @@ case class SessionCookieData(mtditid: String,
       SessionKeys.clientMTDID -> mtditid,
       SessionKeys.clientNino -> nino,
       SessionKeys.clientUTR -> utr,
-      SessionKeys.isSupportingAgent -> isSupportingAgent.toString
-    ) ++ clientFirstName.map(SessionKeys.clientFirstName -> _) ++ clientLastName.map(SessionKeys.clientLastName -> _)
+      SessionKeys.isSupportingAgent -> isSupportingAgent.toString) ++ clientFirstName.map(SessionKeys.clientFirstName -> _) ++ clientLastName.map(SessionKeys.clientLastName -> _)
```
avelx, line 38: I would assume we can revert this file as I would expect this is just a formatting ?

app/services/ITSAStatusService.scala
```diff
@@ -32,10 +31,10 @@ import scala.concurrent.{ExecutionContext, Future}
 class ITSAStatusService @Inject()(itsaStatusConnector: ITSAStatusConnector,
                                   dateService: DateService,
                                   implicit val appConfig: FrontendAppConfig) extends FeatureSwitching {
-  private def getITSAStatusDetail(taxYear: TaxYear, futureYears: Boolean, history: Boolean)
-                                 (implicit user: MtdItUser[_], hc: HeaderCarrier, ec: ExecutionContext): Future[List[ITSAStatusResponseModel]] = {
+  private def getITSAStatusDetail(taxYear: TaxYear, futureYears: Boolean, history: Boolean, nino: String)
```
avelx, line 34: I think we are limiting ourself here, not sure why we have to switch from user: MtdItUser[_] to use nino ? 
anyway, this is something you can confirm with Solar team; buts its fine with me

chandraAndem (commented):
Minor comment.

chandraAndem (approved):

PR #2884, opened 2025-05-20
MISUV-9715

Farnooshmo:


conf/application.conf
```diff
@@ -127,6 +127,11 @@ microservice {
         port = 9057
       }
 
+      set-up-a-payment-plan {
```
nemanjatomanovic, line 130: Do these changes need to be added into config files of other environments? 

⭐ All config can be found [here](https://catalogue.tax.service.gov.uk/whats-running-where)

⭐ [This](https://github.com/hmrc/app-config-base) is base config, but there are one specific for each environment if you need to override it.

it/test/connectors/SelfServeTimeToPayConnectorISpec.scala
```diff
@@ -0,0 +1,100 @@
...
+            SelfServeTimeToPayJourneyResponseModel("journeyId", "nextUrl")
+
+          WiremockHelper.stubPostWithRequest(url, requestBody, CREATED, json)
+
+          val result: SelfServeTimeToPayJourneyResponse = connector.startSelfServeTimeToPayJourney().futureValue
+
+          result shouldBe expectedResponse
+
+          WiremockHelper.verifyPost(
+            uri = s"/essttp-backend/sa/itsa/journey/start"
```
nemanjatomanovic, line 58: 🌟 Since we are verifying the POST call here, it would be great to verify body as well. Something like  `optBody = Some(Json.stringify(requestBody))` would fit perfectly.  This would apply for other cases as well.

test/connectors/SelfServeTimeToPayConnectorSpec.scala
```diff
@@ -0,0 +1,83 @@
...
+      "a 201 response is received with valid json" in {
+        when(mockHttpClientV2.post(any())(any())).thenReturn(mockRequestBuilder)
+
+        when(mockRequestBuilder.withBody(any())(any(), any(), any()))
+          .thenReturn(mockRequestBuilder)
+
+        when(mockRequestBuilder.execute(any[HttpReads[HttpResponse]], any()))
+          .thenReturn(Future(successResponse))
+        val result = TestSelfServeTimeToPayConnector.startSelfServeTimeToPayJourney().futureValue
+        result shouldBe SelfServeTimeToPayJourneyResponseModel("journeyId", "http://www.redirect-url.com")
```
nemanjatomanovic, line 50: ⚡ ⚡ Can you add verify for a number of calls we make towards mocked instances. This will ensure we do not have unnecessary calls to other services.

Good example can be found [here](https://github.com/hmrc/income-tax-view-change-frontend/blob/437ce1bcfba25a1db37e6f76b47a8f918df49a8a/test/controllers/agent/EnterClientsUTRControllerSpec.scala#L85)

This applies to other cases as well.

nemanjatomanovic (approved):

PR #2888, opened 2025-05-22
Making CashAccrualsFlag mandatory for HIP migration

chandraAndem:


chandraAndem:
Closing PR as we don't need to do this anymore

paulanderson1234:
Not required to be fixed for R16 - issue addressed via feature switch for accounting method

PR #2891, opened 2025-05-27
Fixed bug with class 2 Nics

danielmaltz1:
Class 2 Nics was being shown even with 0 outstanding amount, as it was being treated as a coded out charge, but Class 2 Nics can't be coded out

### DIFF
--- a/app/models/financialDetails/DocumentDetail.scala
+++ b/app/models/financialDetails/DocumentDetail.scala
@@ -106,7 +106,7 @@ case class DocumentDetail(taxYear: Int,
       case _ => false
     }
 
-  def isNotCodingOutDocumentDetail: Boolean = !isClass2Nic && !isPayeSelfAssessment && !isCancelledPayeSelfAssessment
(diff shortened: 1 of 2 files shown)


app/models/financialDetails/DocumentDetail.scala
```diff
@@ -100,14 +100,13 @@ case class DocumentDetail(taxYear: Int,
...
-      case (_, true, _) => true
-      case (_, _, true) => true
+    (isPayeSelfAssessment, isCancelledPayeSelfAssessment) match {
+      case (true, _) => true
+      case (_, true) => true
       case _ => false
     }
 
-  def isNotCodingOutDocumentDetail: Boolean = !isClass2Nic && !isPayeSelfAssessment && !isCancelledPayeSelfAssessment
+  def isNotCodingOutDocumentDetail: Boolean = !isPayeSelfAssessment && !isCancelledPayeSelfAssessment
```
avelx, line 109: bug fix must have a tests around it

PR #2895, opened 2025-05-28
MISUV-9644 - Content update on completion page for sign up/Opt out of MTD

IoannisLazaridis-oc:
- [x] Injected `showOptInOptOutContentUpdateR17` and `isCurrentTaxYear` into the controller and passed them to the view.
- [x] Conditionally rendered new panel in the view when the feature switch is enabled.
- [x] Added a paragraph for overdue updates, shown only when the change is to quarterly and the tax year is the current tax year.
- [x] Updated ManageObligationsViewSpec to cover feature switch ON scenarios
- [x] Updated integration tests to cover all user roles and journey types with feature switch enabled/disabled


app/views/manageBusinesses/manage/ManageObligations.scala.html
```diff
@@ -101,74 +137,122 @@ <h2 class="govuk-label-wrapper">
...
+                    link = nextUpdatesControllerURL,
+                    classes = "",
+                    messageKey = "incomeSources.add.manageObligations.OptInOptOutContentUpdateR17.para1.link",
+                    id = None,
+                    target = None,
+                    ariaLabel = None,
+                    outerMessage = "",
+                    role = None,
+                    rel = None,
+                    additionalOpenTabMessage = None
```
Mbrookesy, line 158: Most these arguments don't need to be here as most of them default to None

Mbrookesy (changes requested):
Passes against the ATs currently

app/views/manageBusinesses/manage/ManageObligations.scala.html
```diff
@@ -131,10 +131,8 @@
             )
         )
 
-        <h2 class="govuk-label-wrapper">
-            <label class="govuk-label govuk-label--l govuk-!-margin-top-8">
+        <h2 class="govuk-heading-m">
             @getUniqueMessage("heading")
```
Mbrookesy, line 135: Can we do it so that we aren't changing what the old completion page heading looked like? It only needs to be smaller for the new one

Mbrookesy (changes requested):

PR #2902, opened 2025-06-03
MISUV-9723 Cease business completion page content update

jmountain-hmrc:


IoannisLazaridis-oc (changes requested):
The acceptance test `Agent User is able to add a new Foreign Property income source E2E - Annual Quarterly - Business is in Latency for CY & CY+1` is currently failing because the test expect the user to land on the Reporting Method page but they get redirected to the confirmation page `/foreign-property-added`

PR #2924, opened 2025-06-19
MISUV-9278: Removing of break tags in our code

Farnooshmo:


danielmaltz1 (commented):
There's one more "<br />" in IncomeSourceCeasedObligations.scala.html

PR #2931, opened 2025-06-30
MISUV-9836: Content updates for Next Updates Tile in V&C Homepage for R17

IoannisLazaridis-oc:
Display `Next Quarterly Update` and `Next Tax Return` due dates on Next Updates tile  based on ITSA Status when `OptInOptOutContentUpdateR17` is enabled. 

- [x] Extended NextUpdatesTileViewModel with new optional fields.
- [x] Updated controller and service to fetch conditionally next quarterly update and tax return due dates.
   - The **next quarterly update due date** is fetched from obligations where `obligationType == "Quarterly"` with due dates that are today or in the future.
   - The **next tax return due date** is fetched from obligations where `obligationType == "Crystallisation"` with due dates that are today or in the future. If value is not present then the  **next tax return due date** statically calculated using  the `getCurrentTaxYear.endYear + 1` => `31 January CY+1`.
 - [x] Updated view to show dynamic content based on ITSA status and feature switch.
 - [x] Added tests for individual, agent, and supporting agent scenarios.


app/controllers/HomeController.scala
```diff
@@ -243,4 +257,22 @@ private def handleErrorGettingDueDates(isAgent: Boolean)(implicit user: MtdItUse
...
+    }
+  }
+
+  private def getNextTaxReturnDueDateIfEnabled()
+                                              (implicit hc: HeaderCarrier, user: MtdItUser[_]): Future[Option[LocalDate]] = {
+    if (isEnabled(OptInOptOutContentUpdateR17)) {
+      nextUpdatesService.getNextTaxReturnDueDate()
+    } else {
+      Future.successful(None)
+    }
```
Mbrookesy, line 276: Did we get an idea from design on what exactly we should do in a scenario where the FS is on but they don't get a date from the backend? I feel like leaving it blank looks odd.

Might be worth asking if first that's even possible and second if so, what to put there (or omit)

app/services/NextUpdatesService.scala
```diff
@@ -159,6 +159,50 @@ class NextUpdatesService @Inject()(
 
     processingRes
   }
+
+  def getNextQuarterlyUpdateDueDate()(implicit hc: HeaderCarrier, mtdItUser: MtdItUser[_]): Future[Option[LocalDate]] = {
+    getOpenObligations().map {
+      case ObligationsModel(obligations) =>
+        val allQuarterlies = obligations
+          .flatMap(_.obligations)
+          .filter(o => o.obligationType == "Quarterly" && o.status == StatusOpen)
```
Mbrookesy, line 168: You don't need to filter by `StatusOpen` here as the connector already calls only the open obligations via downstream.

app/controllers/HomeController.scala
```diff
@@ -243,4 +257,22 @@ private def handleErrorGettingDueDates(isAgent: Boolean)(implicit user: MtdItUse
   private def mainChargeIsNotPaidFilter: PartialFunction[ChargeItem, ChargeItem]  = {
     case x if x.remainingToPayByChargeOrInterestWhenChargeIsPaid  => x
   }
+
+  private def getNextQuarterlyUpdateDueDateIfEnabled()
```
Mbrookesy, line 261: Would it be possible to combine these two functions so that we aren't calling the downstream twice for the same set of data that can be individually filtered.

Mbrookesy (changes requested):

app/controllers/HomeController.scala
IoannisLazaridis-oc: From business perspective both values will be always present.

app/models/obligations/NextUpdatesTileViewModel.scala
```diff
@@ -16,9 +16,17 @@
...
+import models.itsaStatus.ITSAStatus.ITSAStatus
+
 import java.time.LocalDate
 
-case class NextUpdatesTileViewModel(dueDates: Seq[LocalDate], currentDate: LocalDate, isReportingFrequencyEnabled: Boolean) {
+case class NextUpdatesTileViewModel(dueDates: Seq[LocalDate],
+                                    currentDate: LocalDate,
+                                    isReportingFrequencyEnabled: Boolean,
+                                    showOptInOptOutContentUpdateR17: Boolean,
+                                    currentYearITSAStatus: Option[ITSAStatus],
```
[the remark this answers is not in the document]
IoannisLazaridis-oc: Used `currentYearITSAStatus: Option[ITSAStatus]` because when `showOptInOptOutContentUpdateR17` is off, the view doesn’t use ITSA Status.

app/services/NextUpdatesService.scala
```diff
@@ -159,6 +159,50 @@ class NextUpdatesService @Inject()(
...
+
+  def getNextTaxReturnDueDate()(implicit hc: HeaderCarrier, mtdUser: MtdItUser[_]): Future[Option[LocalDate]] = {
+    getOpenObligations().map {
+      case ObligationsModel(obligations) =>
+        val openFinalReturns = obligations
+          .flatMap(_.obligations)
+          .filter(o => o.obligationType == "Crystallisation" && o.status == StatusOpen)
+          .sortBy(_.due)
+
+        val nextTaxReturn = openFinalReturns.headOption.map(_.due)
```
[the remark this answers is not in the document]
IoannisLazaridis-oc: removed - logic has changed

app/controllers/HomeController.scala
IoannisLazaridis-oc: Updated the function and removed the API call due to unreliable data - now the tax return due date is calculated based on the current tax year and the current date.

Mbrookesy (changes requested):

app/services/NextUpdatesService.scala
```diff
@@ -159,6 +159,46 @@ class NextUpdatesService @Inject()(
...
+
+      case unexpected =>
+        Logger("application").error(s"[getNextDueDates] Unexpected response: $unexpected")
+        (None, Some(calculateNextTaxReturnDueDate()))
+    }
+  }
+
+  private def calculateNextTaxReturnDueDate(): LocalDate = {
+    val currentDate = LocalDate.now()
+    val currentTaxYearStart = dateService.getCurrentTaxYear.startYear
```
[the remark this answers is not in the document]
IoannisLazaridis-oc: If we change this to `endYear` for the current tax (2025 - 2026)  year it will return tax return due : 31 January 2027 and not 2026

app/services/NextUpdatesService.scala
```diff
@@ -159,6 +159,46 @@ class NextUpdatesService @Inject()(
...
+    var nextTaxReturnDate: LocalDate = LocalDate.of(currentTaxYearStart + 1, 1, 31)
+
+    if (!currentDate.isBefore(nextTaxReturnDate)) {
```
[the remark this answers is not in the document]
IoannisLazaridis-oc: this is a check if the current date is after  the resulted next tax return date for example 
if the initial generated date is 31 January 2026 and the current date is 1 February 2026 that means we should display 31 January 2027 which is the expected behaviour.

chandraAndem (dismissed):

chandraAndem (approved):

PR #2932, opened 2025-06-30
[MISUV-9798] Fixing the user being able to format HTML in the add incom…

Mbrookesy:
…e source address section


Mbrookesy:
rebuild

michael-hmrc:
wanted to rebase and squash commits but due to prior commit history creates horrible conflicts might leave it for now since more trouble than it's worth...

jmountain-hmrc:
Can still reproduce the crash

`		` <- using tab characters

michael-hmrc:
Addin Filtering for emojis displayed only not saved. We still save emojis/full user input in mongo.

Screenshot 2025-07-14 at 11 52 47

PR #2944, opened 2025-07-10
MISUV-9834 Content update to manage income sources CYA page

michael-hmrc:
reviewer notes:

- Tried to uncouple the keys for the old income source manage cya and the new version.
- kept most of the tests but changed some ids and such. moved messages and constants to help modularise the code to keep the view spec a little concise. 
- added the logic as boolean flags but need to test it a little to see if it matches the ask of the ticket.


TomRafferty (approved):
raise ticket to fix bugged link (no fault of this ticket work) - other than that, page looks good, code has been tidied 😄 nice work

PR #2967, opened 2025-07-28
Misuv 9888

chandraAndem:


app/services/manageBusinesses/IncomeSourceRFService.scala
```diff
@@ -92,76 +76,14 @@ class IncomeSourceRFService @Inject()(val sessionService: SessionService,
...
-    data match {
-      case data: Option[IncomeSourceReportingFrequencySourceData] if businessIsInLatency(latencyDetails) =>
-        updateChooseTaxYearsModelInSessionData(sessionData, data)
-          .flatMap(codeBlock(_))
-      case _ =>
-        redirectToIncomeSourceAddedPage(isAgent, incomeSourceType)
-    }
+    onlyOneYearInLatency(latencyDetails) ||
+      isItsaStatusAnnual(currentTyStatus) ||
+      isItsaStatusAnnual(nextTyStatus)
```
Mbrookesy, line 86: Looking at the `isItsaStatusAnnual` functions, if there's a missing ITSA status or the status is of No Status or something other than Annual it flags this as false.

One thought I had was that do we want the user to be able to access the latency reporting frequency pages if their ITSA status is missing or No Status? It looks like that would be the case currently

chandraAndem: Yeah added conditions to handle these scenarios i.e., all non quarterly status are treated as one.

For cases like ITSA status is missing and No Status is also handled. Please review.

Mbrookesy (approved):

Mbrookesy (dismissed):

Mbrookesy (approved):

PR #2968, opened 2025-07-28
MISUV-9688: Update ITSA status models and  fix reporting frequency for unusual statuses

IoannisLazaridis-oc:
The frontend now recognises the `Digitally Exempt` status, and the backend [maps](https://github.com/hmrc/income-tax-view-change/pull/360) the API response code `04` to `DigitallyExempt`. We have also updated the ITSA status handling so that unsupported statuses (`Digitally Exempt`, `Exempt`, `Dormant`, and `NoStatus`) now return a 500 error. The only exception is `NoStatus` for CY+1, which is treated as valid and allowed.


chandraAndem:
test this please

Mbrookesy:
rebuild

test/repositories/ITSAStatusRepositorySupportSpec.scala
```diff
@@ -23,35 +23,46 @@ import testUtils.UnitSpec
...
-        statusToString(Dormant)
+    "allow NoStatus only for CY+1" in {
+      statusToString(status = NoStatus) shouldBe "U"
+    }
+
+    "throw error for other unsupported statuses" in {
+      Seq(Exempt, DigitallyExempt, Dormant).foreach { unsupportedStatus =>
+        assertThrows[RuntimeException] {
+          statusToString(status = unsupportedStatus)
+        }
```
Mbrookesy, line 43: This test fails currently

`Expected exception java.lang.RuntimeException to be thrown, but no exception was thrown`

app/repositories/ITSAStatusRepositorySupport.scala
```diff
@@ -23,12 +23,13 @@ object ITSAStatusRepositorySupport {
 
   def statusToString(status: ITSAStatus): String =
     status match {
-      case ITSAStatus.NoStatus  => "U"
+      //Making all the valid status as Unknown until we get clarity on what to do with these status
+      case ITSAStatus.NoStatus | ITSAStatus.Exempt | ITSAStatus.DigitallyExempt | ITSAStatus.Dormant  => "U"
```
Mbrookesy, line 27: Looks like this is the area where the test is failing from.

it's no longer calling a runtime exception due to those valid statuses being set to unknown

Mbrookesy (changes requested):

it/test/controllers/optOut/ConfirmOptOutControllerISpec.scala
```diff
@@ -100,12 +102,24 @@ class ConfirmOptOutControllerISpec extends ControllerISpecHelper {
...
-              elementTextByID("optOut-warning")(infoMessage),
+              elementTextByID("heading")(expectedTitle(dateService)),
+              elementTextByID("summary")(summary),
+              elementTextByID("info-message")(infoMessage),
             )
           }
+          s"throw an exception for multi-year opt out page when unsupported statuses are present" in {
+            enable(OptOutFs)
+            disable(NavBarFs)
+            stubAuthorised(mtdUserRole)
```
Mbrookesy, line 113: Failed test here too due to the same reasons as the others. It isn't throwing an exception.

Mbrookesy (changes requested):

PR #2989, opened 2025-08-21
[MISUV-10106] updating ceased business warning message and hiding secti…

Mbrookesy:
…on if business is ceased


TomRafferty (approved):
LGTM

PR #2993, opened 2025-08-22
MISUV-10104 Update content to be scenario based

michael-hmrc:


conf/messages
```diff
@@ -2681,6 +2690,42 @@ optout.confirmedOptOut.reportQuarterly.reason-list               = This could be
...
+optout.confirmedOptOut.scenario1.reporting.obligations.bullet2                = you report an increase in your qualifying income in a tax return
+optout.confirmedOptOut.scenario1.reporting.obligations.inset                  = For example, if your income from self-employment or property, or both, exceeds the £50,000 threshold in the 2024 to 2025 tax year, you would have to use Making Tax Digital for Income Tax from 6 April 2026.
+optout.confirmedOptOut.scenario1.reporting.obligations.p3                     = If this happens, we will write to you to let you know.
+optout.confirmedOptOut.scenario1.reporting.obligations.p4                     = You can check the threshold for qualifying income in the
+optout.confirmedOptOut.scenario1.reporting.obligations.p4.link                = criteria for people who will need to sign up for Making Tax Digital for Income Tax
+
+optout.confirmedOptOut.scenario2.reporting.obligations.heading                = Your reporting obligations from the next tax year onwards
+optout.confirmedOptOut.scenario2.reporting.obligations.inset                  = From 6 April 2025, you will be required to use Making Tax Digital for Income tax.
```
Mbrookesy, line 2703: This should be dynamic as its refering to the mandated CY+1

app/views/optOut/ConfirmedOptOutView.scala.html
```diff
@@ -0,0 +1,374 @@
...
+        @scenario3ReportingObligationsSection
+    }
+
+    @if(confirmedOptOutViewScenarios == Scenario4Content) {
+        @greenPanelScenario1And4
+        @revisedDeadlinesBlock
+        <div id="submit-tax" class="govuk-!-margin-bottom-7  govuk-!-margin-top-8">
+            @h2(msg = "optout.confirmedOptOut.submitTax", optId = Some("submit-your-tax-return-heading"))
+            @submitTaxReturnParagraph1
+        </div>
```
Mbrookesy, line 371: I noticed that in the submit tax return paragraph on the ticket, it says scenario 4 should say `As you are opted out, you can find out here how to file your Self Assessment tax return (opens in new tab)`

While it currently shows this
image

Mbrookesy (changes requested):

conf/messages
michael-hmrc: updated think it should be dynamic now

app/views/optOut/ConfirmedOptOutView.scala.html
michael-hmrc: updated to have it's own section to use in scenario4

PR #3045, opened 2025-09-26
MISUV-10238: Fix bug with manage details missing content when crystallisation status not found

IoannisLazaridis-oc:
- [x] Replaced usage of. `showLatencyDetails` with `IsBusinessInLatency` in the view. 
- [x] Removed `showLatencyDetails` and `shouldShowTaxYears` from view model (no longer used).
- [x] Updated specs to cover the case when latency details are absent, making sure the latency paragraph is not rendered.

### DIFF
--- a/it/test/controllers/manageBusinesses/manage/ManageIncomeSourceDetailsUKPropertyControllerISpec.scala
+++ b/it/test/controllers/manageBusinesses/manage/ManageIncomeSourceDetailsUKPropertyControllerISpec.scala
@@ -91,7 +91,8 @@ class ManageIncomeSourceDetailsUKPropertyControllerISpec extends ManageIncomeSou
                 elementTextBySelectorList("#manage-details-table", "div:nth-of-type(2)", "dt")("Accounting method"),
                 elementTextBySelectorList("#manage-details-table", "div:nth-of-type(2)", "dd")(businessAccountingMethod),
                 elementTextByID("change-link-1")(messagesChangeLinkText),
-                elementTextByID("change-link-2")(messagesChangeLinkText)
+                elementTextByID("change-link-2")(messagesChangeLinkText),
+                elementTextByID("up-to-two-tax-years")("Because this is still a new business, you can change how often you report for it for up to 2 tax years. From April 2027, you could be required to report quarterly.")
               )
             }
 
--- a/test/controllers/manageBusinesses/manage/ManageIncomeSourceDetailsForeignPropertySpec.scala
+++ b/test/controllers/manageBusinesses/manage/ManageIncomeSourceDetailsForeignPropertySpec.scala
@@ -158,6 +158,11 @@ class ManageIncomeSourceDetailsForeignPropertySpec extends ManageIncomeSourceDet
               getHeading(document) shouldBe heading
 
               hasInsetText(document) shouldBe true
+
+              val latencyParagraph = document.getElementById("up-to-two-tax-years")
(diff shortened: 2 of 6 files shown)


it/test/controllers/manageBusinesses/manage/ManageIncomeSourceDetailsUKPropertyControllerISpec.scala
```diff
@@ -63,7 +62,8 @@ class ManageIncomeSourceDetailsUKPropertyControllerISpec extends ManageIncomeSou
                 elementTextBySelectorList("#manage-details-table", "div:nth-of-type(1)", "dt")("Date started"),
                 elementTextBySelectorList("#manage-details-table", "div:nth-of-type(1)", "dd")(businessStartDate),
                 elementTextBySelectorList("#manage-details-table", "div:nth-of-type(2)", "dt")("Accounting method"),
-                elementTextBySelectorList("#manage-details-table", "div:nth-of-type(2)", "dd")(businessAccountingMethod)
+                elementTextBySelectorList("#manage-details-table", "div:nth-of-type(2)", "dd")(businessAccountingMethod),
+                elementTextByID("up-to-two-tax-years")("")
```
michael-hmrc, line 66: if I understand this correctly this test is for the scenario for no latency information returned would we want a test for when there is latency information returned and the correct content is returned?

test/controllers/manageBusinesses/manage/ManageIncomeSourceDetailsUkPropertySpec.scala
```diff
@@ -63,6 +63,7 @@ class ManageIncomeSourceDetailsUkPropertySpec extends ManageIncomeSourceDetailsH
               hasGracePeriodInfo(document) shouldBe false
               getManageDetailsSummaryValues(document).get(2).text() shouldBe standard
               document.getElementById("reportingFrequency").text() shouldBe "View and change your reporting frequency for all your businesses"
+              Option(document.getElementById("up-to-two-tax-years")) shouldBe None
```
michael-hmrc, line 66: guess same here for a test where we do return content

test/controllers/manageBusinesses/manage/ManageIncomeSourceDetailsForeignPropertySpec.scala
```diff
@@ -62,6 +62,7 @@ class ManageIncomeSourceDetailsForeignPropertySpec extends ManageIncomeSourceDet
               hasGracePeriodInfo(document) shouldBe false
               getManageDetailsSummaryValues(document).get(2).text() shouldBe calendar
               document.getElementById("reportingFrequency").text() shouldBe "View and change your reporting frequency for all your businesses"
+              Option(document.getElementById("up-to-two-tax-years")) shouldBe None
```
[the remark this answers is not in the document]
IoannisLazaridis-oc: Good point! There is already a test covering the success case in the `test/views/manageBusinesses/manage/ManageIncomeSourceDetailsViewSpec.scala` at line 407 including the rendered content. I didn't add the same check in the controller/integration specs to avoid duplicating assertions across layers and having the view spec to check the actual content. I think the best approach is to add one success path test at the controller level that asserts the element is present but without checking the exact text. Basically adding sanity checks for the elements that should be present.

michael-hmrc (dismissed):

michael-hmrc (approved):

PR #3047, opened 2025-09-30
Enabling Signup and Optout feature R17 switch

chandraAndem:


IoannisLazaridis-oc (dismissed):

conf/application.conf
```diff
@@ -199,7 +199,7 @@ feature-switch {
   enable-time-machine = true
   enable-triggered-migration = false
   enable-your-self-assessment-charges = false
-  opt-in-opt-out-content-update-r17 = false
```
michael-hmrc, line 202: does this need to be enable or is enable- a env based thing?

PR #3050, opened 2025-10-02
MISUV-10263 - Optout/Signup - Fix behaviour regarding updating accoun…

michael-hmrc:
…t itsa statuses when selecting tax years

To try and break up the service, adding new service to handle optout update account itsa status logic. Also trying to untangle the logic

Some small clean up to help me avoid confusion in naming etc.

- Renamed these to submission services, since they handle submissions and update the account as a result.

~~**Note:** We may need to rethink the audit event on submission, this is because we make multiple calls out to update the itsa status for optout, up to 3 times for each year CY-1, CY & CY+1. We also send multiple audit events, one for each request I believe. 
If desired we can refactor the Audit event to only fire once, we can send all outcomes/responses from the ITSAUpdate Connector API which we call to make updates in a single audit event.
The data being send to audit during submissions for optout is off, we repeatedly send the status of all tax years previously before the update and after for all years. So it's I am thinking it is sending multiple audit events slowly updating the statues each time.  But we could send these all together in one go and reduce the data sent by x3 I think.~~

~~If decided not to touch, we can refactor to keep current audit model and up to 3  audit events per submission, one for each request.~~

~~**Note To Self:** Update optout single tax year controller with new trigger method~~ 

**Note:** Was decided to not touch the audit event on submission

### DIFF
--- a/app/connectors/itsastatus/ITSAStatusUpdateConnectorModel.scala
+++ b/app/connectors/itsastatus/ITSAStatusUpdateConnectorModel.scala
@@ -32,8 +32,6 @@ object ITSAStatusUpdateConnectorModel {
 
   sealed trait ITSAStatusUpdateResponse
 
-  sealed trait ITSAStatusUpdateError
-
   case class ErrorItem(code: String, reason: String)
 
   case class ITSAStatusUpdateResponseSuccess(statusCode: Int = NO_CONTENT) extends ITSAStatusUpdateResponse


Mbrookesy (changes requested):
Few comments so far while looking through, will have a further look

app/connectors/itsastatus/ITSAStatusUpdateConnector.scala
```diff
@@ -78,11 +78,53 @@ class ITSAStatusUpdateConnector @Inject()(val httpClient: HttpClientV2, val appC
...
                 valid
               }
             )
         }
       }
   }
+
+  def makeMultipleItsaStatusUpdateRequests(
+                                            taxYears: List[TaxYear],
+                                            userNino: String,
```
Mbrookesy, line 91: Did we ever come up with a solution over what we should do if for say, we update multiple tax years but one happens to fail? It looks like if there's at least one failure itll show an error page but should we be leaving the successful status update as updated?

It may get a bit messy here as if we tried to revert and that failed then it may just end up in a loop.

michael-hmrc: hmm good question yeah having this mechanism for opt out is bit of a headache since some can fail and some could succeed. Guess it would be weird showing the error page when possibly two of the succeeded :l

Mbrookesy: it's not ideal but i feel like we may have to just allow the user to opt out/sign up for the successful year and leave the other one for the user to spot when they get back to the reporting obligations page (feel free to clarify with Chandra however)

In an ideal world the backend would update all the tax years at once and only do the update if all succeed but sadly it doesn't work like that

app/services/optIn/SignUpSubmissionService.scala
```diff
@@ -0,0 +1,77 @@
...
+    } yield {
+      optInSessionData
+    }
+  }
+
+  def triggerOptInRequest()(implicit user: MtdItUser[_], hc: HeaderCarrier, ec: ExecutionContext): Future[ITSAStatusUpdateResponse] = {
+    for {
+      optInSessionData: Option[OptInSessionData] <- getOptInSessionData()
+      selectedOptInYear: Option[TaxYear] = optInSessionData.flatMap(_.selectedOptInYear).flatMap(TaxYear.`fromStringYYYY-YYYY`)
+      optInContextData: Option[OptInContextData] = optInSessionData.flatMap(_.optInContextData)
```
Mbrookesy, line 55: optInContextData doesn't look used here

app/connectors/itsastatus/ITSAStatusUpdateConnector.scala
```diff
@@ -78,11 +78,53 @@ class ITSAStatusUpdateConnector @Inject()(val httpClient: HttpClientV2, val appC
...
 
-                log.error(s"response status: ${response.status}, message: $message")
+                logger.error(s"response status: ${response.status}, message: $message")
                 valid
               }
             )
         }
       }
   }
+
```
Mbrookesy, line 88: If you wanted to, you can probably just use one function for both single and multi year and just put the single year ones in a list on its own. It'll save not having to have 2 functions doing pretty much the same thing

app/connectors/itsastatus/ITSAStatusUpdateConnector.scala
```diff
@@ -78,11 +78,53 @@ class ITSAStatusUpdateConnector @Inject()(val httpClient: HttpClientV2, val appC
...
+  def makeMultipleItsaStatusUpdateRequests(
```
Mbrookesy, line 89: The Sign up journey doesn't appear to be using this

test/services/optIn/SignUpSubmissionServiceSpec.scala
```diff
@@ -0,0 +1,193 @@
...
+          val currentOptInTaxYear =
+            CurrentOptInTaxYear(Voluntary, TaxYear(2025, 2026))
+
+          val nextOptInTaxYear =
+            NextOptInTaxYear(Voluntary, TaxYear(2026, 2027), currentOptInTaxYear)
+
+          val optInProposition =
+            OptInProposition(currentOptInTaxYear, nextOptInTaxYear)
+
+          when(mockUiJourneySessionDataRepository.get(hc.sessionId.get.value, Opt(OptInJourney)))
```
Mbrookesy, line 132: There's no tests/functionality for multi year

michael-hmrc: added some tests and changed some of the logic and validation to handle single and multi years 🤞🏼

app/connectors/itsastatus/ITSAStatusUpdateConnector.scala
michael-hmrc: removed

app/services/optIn/SignUpSubmissionService.scala
```diff
@@ -0,0 +1,77 @@
...
+      optInProposition: OptInProposition <- optInService.fetchOptInProposition()
+      updateResponse <- selectedOptInYear match {
+        case Some(taxYear) =>
+          itsaStatusUpdateConnector.optIn(taxYear = taxYear, user.nino)
+        case None =>
+          Future(ITSAStatusUpdateResponseFailure.defaultFailure())
+      }
```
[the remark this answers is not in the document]
michael-hmrc: doesn signup have a multi year forward looking thing like optout? cannot remember exactly but changed it to validate for single tax year and multi tax year scenarios 🤞🏼 hoping it's there now. Had some issue with adding the recent tests and keep forgetting for single years the user cannot select tax years I think?

michael-hmrc:
ATs run against branch


Screenshot 2025-10-21 at 11 38 06

app/controllers/optOut/newJourney/OptOutTaxYearQuestionController.scala
```diff
@@ -64,19 +71,49 @@ class OptOutTaxYearQuestionController @Inject()(val optOutService: OptOutService
...
+
+    val formResponse = validForm.toFormMap(OptOutTaxYearQuestionForm.response).headOption
+
+    formResponse match {
+      case Some(OptOutTaxYearQuestionForm.responseYes) =>
+        if (redirectToConfirmUpdatesPage) {
+          Future(Redirect(controllers.optOut.newJourney.routes.ConfirmOptOutUpdateController.show(isAgent, taxYear.getOrElse(""))))
+        } else {
+          optOutSubmissionService.updateTaxYearsITSAStatusRequest(ITSAStatus.Voluntary).map {
+            case listOfUpdateRequestsMade if !listOfUpdateRequestsMade.exists(_.isInstanceOf[ITSAStatusUpdateResponseFailure]) || listOfUpdateRequestsMade == List.empty =>
```
Mbrookesy: Same for here regarding an empty list

app/models/ReportingFrequencyViewModel.scala
```diff
@@ -129,13 +129,14 @@ case class ReportingFrequencyViewModel(
...
+    val test = checkIfOnwards.zipWithIndex.map {
       case (MultiYearCard, 0)  => Some("optOut.previousYear.onwards")
       case (SingleYearCard, 0) => Some("optOut.previousYear.single")
       case (MultiYearCard, 1)  => currentYearSuffix.map(_ + ".onwards")
       case (SingleYearCard, 1) => currentYearSuffix.map(_ + ".single")
       case (MultiYearCard, 2)  => nextYearSuffix
       case _ => None
     }
+    test
   }
```
Mbrookesy: There's test code still in here

test/services/optIn/SignUpSubmissionServiceSpec.scala
```diff
@@ -0,0 +1,485 @@
...
+                selectedSignUpYear = Some(currentTaxYear),
+                currentYearItsaStatus = Some(Annual),
+                nextYearItsaStatus = Some(Annual)
+              )
+
+            whenReady(request) { result => result shouldBe ITSAStatusUpdateResponseSuccess(204) } // 204 No Content
+          }
+        }
+
+        "only CY is Annual - single tax year scenario, user is unable to select a yax year" should {
```
Mbrookesy: Tax year is spelt wrong

app/services/optIn/SignUpSubmissionService.scala
```diff
@@ -0,0 +1,129 @@
...
+    val nextTaxYear = currentTaxYear.nextYear
+
+    (selectedSignUpYear, currentYearItsaStatus, nextYearItsaStatus) match {
+      case (Some(selectedTaxYear), Some(Annual), Some(Annual)) =>
+        itsaStatusUpdateConnector.optIn(taxYear = selectedTaxYear, user.nino)
+      case (None, Some(Annual), Some(nextYearStatus)) if nextYearStatus != Annual =>
+        itsaStatusUpdateConnector.optIn(taxYear = currentTaxYear, user.nino)
+      case (None, Some(currentYearStatus), Some(Annual)) if currentYearStatus != Annual =>
+        itsaStatusUpdateConnector.optIn(taxYear = nextTaxYear, user.nino)
+      case (Some(selectedTaxYear), None, None) =>
```
Mbrookesy: I don't really understand why there are scenarios for when they don't pick a tax year? Pretty sure the user has to select a tax year to sign up for/from no matter what

michael-hmrc: it's because we don't populate the db with the selected tax year when it's only a single year. If there is no selected tax year it breaks. The selected tax year field is only for multi years I think. During manual testing it was breaking because of this.

app/services/optIn/SignUpSubmissionService.scala
```diff
@@ -0,0 +1,129 @@
...
+    for {
+      uiSessionData <- uiJourneySessionDataRepository.get(hc.sessionId.get.value, Opt(OptInJourney))
+      optInSessionData: Option[OptInSessionData] = uiSessionData.flatMap(_.optInSessionData)
+    } yield {
+      optInSessionData
+    }
+  }
+
+  def makeUpdateRequest(
+                         selectedSignUpYear: Option[TaxYear],
```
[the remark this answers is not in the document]
Mbrookesy: Last i checked you can do a multi year sign up if both CY and CY+1 are Annual and you sign up from CY

app/controllers/optOut/newJourney/OptOutTaxYearQuestionController.scala
```diff
@@ -64,19 +69,51 @@ class OptOutTaxYearQuestionController @Inject()(val optOutService: OptOutService
...
+    val formResponse = validForm.toFormMap(OptOutTaxYearQuestionForm.response).headOption
+
+    formResponse match {
+      case Some(OptOutTaxYearQuestionForm.responseYes) =>
+        if (redirectToConfirmUpdatesPage) {
+          Future(Redirect(controllers.optOut.newJourney.routes.ConfirmOptOutUpdateController.show(isAgent, taxYear.getOrElse(""))))
+        } else {
+          optOutSubmissionService.updateTaxYearsITSAStatusRequest().map {
+            case Right(List()) =>
+              Redirect(controllers.optOut.oldJourney.routes.OptOutErrorController.show(isAgent))
```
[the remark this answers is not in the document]
michael-hmrc: yeah can probs remove the either since originally thought it could get caught earlier or when speaking with Chandra he was saying it can be caught earlier in reality it can't.

app/services/optout/OptOutSubmissionService.scala
```diff
@@ -0,0 +1,184 @@
...
+
+    val maybeOptOutContextData: Option[OptOutContextData] = maybeSessionData.flatMap(_.optOutContextData)
+    val maybeCurrentTaxYear: Option[TaxYear] = maybeOptOutContextData.map(_.currentYear).flatMap(TaxYear.`fromStringYYYY-YYYY`)
+    val selectedOptOutTaxYear: Option[TaxYear] = maybeSessionData.flatMap(sessionData => sessionData.selectedOptOutYear).flatMap(TaxYear.`fromStringYYYY-YYYY`)
+    val crystallisationStatus = maybeOptOutContextData.exists(_.crystallisationStatus)
+
+    val gatherTaxYearsToUpdateBasedOnUserAnswerChoice =
+      (maybeCurrentTaxYear, selectedOptOutTaxYear) match {
+        case (Some(currentTaxYear), Some(selectedTaxYear)) if selectedTaxYear == currentTaxYear.previousYear && crystallisationStatus =>
+          allYearsToUpdate.filterNot(_.taxYear == currentTaxYear.previousYear)
```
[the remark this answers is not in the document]
michael-hmrc: it's basically filtering out CY and CY+1 to be returned since if it's CY-1 and crystallised the user cannot update CY-1.

I think thats what it is saying with the filterNot.

app/controllers/optOut/newJourney/OptOutTaxYearQuestionController.scala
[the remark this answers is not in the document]
Mbrookesy: Yeah probably a good idea

app/services/optout/OptOutSubmissionService.scala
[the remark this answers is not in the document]
Mbrookesy: Yeah but i dont see a situation where you'd be able to opt out of CY-1 if its crystallised as if it's crystallised the link wouldn't show up

Mbrookesy (approved):

PR #3056, opened 2025-10-07
[MISUV-10296] Fixing opting out of CY+1 not working when CY is not annu…

Mbrookesy:
…al or mandated

### DIFF
--- a/app/services/optout/OptOutService.scala
+++ b/app/services/optout/OptOutService.scala
@@ -302,16 +302,10 @@ class OptOutService @Inject()(
           val currentYearStatus = proposition.currentTaxYear.status
           val nextYearStatus = proposition.nextTaxYear.status
 
-          (checkOptOutStatus, proposition.optOutPropositionType, currentYearStatus) match {
-            case (Some((true, propositionTaxYear)), Some(propositionType), status)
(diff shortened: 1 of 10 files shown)


app/services/optout/OptOutService.scala
```diff
@@ -302,16 +302,10 @@ class OptOutService @Inject()(
...
-          (checkOptOutStatus, proposition.optOutPropositionType, currentYearStatus) match {
-            case (Some((true, propositionTaxYear)), Some(propositionType), status)
-              if propositionType.state().exists(state =>
-                state == MultiYearOptOutDefault ||
-                  state == OneYearOptOutFollowedByMandated ||
-                  state == OneYearOptOutFollowedByAnnual ||
-                  (state == NextYearOptOut && (status == Annual || status == Mandated))
-              ) =>
+          (checkOptOutStatus, proposition.optOutPropositionType) match {
+            case (Some((true, propositionTaxYear)), Some(propositionType)) if propositionType.state().isDefined =>
```
IoannisLazaridis-oc, line 306: Do you think adding an info log here would help to have clearer picture when these generic states are handled and accepted?

PR #3339, opened 2026-04-08
MISUV-11094b Fix AddBusinessTrade error summary

michael-hmrc:


TomRafferty (approved):
LGTM

Small note:
WhatYouOweControllerISpec failed for me locally, I see nothing here that would have any effect on that + all checks pass here, I also recall that to be a bit flakey anyway

PR #3346, opened 2026-04-13
MISUV-11119 - proposed log change and redirect

TomRafferty:
This may not be the way we want to go, I feel as though having some more accurate logging with all the information in one log may help us here - though I'm unsure on a solution right now.

### DIFF
--- a/app/repositories/OptOutSessionDataRepository.scala
+++ b/app/repositories/OptOutSessionDataRepository.scala
@@ -38,7 +38,7 @@ class OptOutSessionDataRepository @Inject()(val repository: UIJourneySessionData
       .map(journeySd => journeySd.copy(optOutSessionData = journeySd.optOutSessionData.map(_.copy(selectedOptOutYear = Some(intent.toString)))))
       .flatMap(journeySd => OptionT.liftF(repository.set(journeySd)))
       .getOrElse({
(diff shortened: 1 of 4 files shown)


app/repositories/OptOutSessionDataRepository.scala
```diff
@@ -36,7 +37,10 @@ class OptOutSessionDataRepository @Inject()(val repository: UIJourneySessionData
     OptionT(repository.get(hc.sessionId.get.value, Opt(OptOutJourney)))
       .map(journeySd => journeySd.copy(optOutSessionData = journeySd.optOutSessionData.map(_.copy(selectedOptOutYear = Some(intent.toString)))))
       .flatMap(journeySd => OptionT.liftF(repository.set(journeySd)))
-      .getOrElse(false)
+      .getOrElse({
+        Logger("application").error(s"Failed to collect session data for sessionId: ${hc.sessionId.get.value} when trying to save opt out intent. Tax year intent was: ${intent.toString}")
```
jmountain-hmrc: Might be to use a .getOrElse instead of .get for sessionId

TomRafferty: good shout

jmountain-hmrc (dismissed):

Mbrookesy (approved):

PR #3358, opened 2026-04-17
MIPR-2571: updated services tests to work when read-from-mongo is true

kSiadat:


Chrissystk (dismissed):
There only seems to be two test suites covered here but when I do a search we are calling enable(...) over more files. if you do a find in files for enable( in the services folder you should be able to see the references.

Chrissystk (approved):

PR #3359, opened 2026-04-17
MISUV-11148 Update to move fullstops out of link text for income sources pages

AdamSmithHMRC:


app/controllers/manageBusinesses/cease/IncomeSourceNotCeasedController.scala
```diff
@@ -55,7 +55,7 @@ class IncomeSourceNotCeasedController @Inject()(val authActions: AuthActions,
     val linkPrefix = Some(messagesApi.preferred(user)("incomeSources.cease.error.notCeased.link.prefix"))
 
     Future.successful {
-      Ok(errorTemplate(pageTitle, heading, message, linkPrefix, linkText, linkUrl, isAgent = isAgent))
+      Ok(errorTemplate(pageTitle, heading, message, linkPrefix, linkText, linkUrl, isAgent = isAgent, optLinkPunct = Some(".")))
```
michael-hmrc, line 58: just wanted to check, why are we adding the param here instead of in the template? is it because it's only this page which needs it?

michael-hmrc (approved):

PR #3365, opened 2026-04-21
MISUV-11158: Add NoIncomeSources page and redirect users without income sources

IoannisLazaridis-oc:
To manually test use the new user `NA000001A`
Stub data PR: https://github.com/hmrc/income-tax-view-change-frontend-acceptance-test/pull/1385

### DIFF
--- a/app/auth/authV2/actions/RedirectIfNoIncomeSourcesAction.scala
+++ b/app/auth/authV2/actions/RedirectIfNoIncomeSourcesAction.scala
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2026 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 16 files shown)


michael-hmrc (dismissed):
lgtm

app/auth/authV2/actions/RedirectIfNoIncomeSourcesAction.scala
```diff
@@ -0,0 +1,50 @@
+/*
```
michael-hmrc, line 1: just wondering think we may not need to create a new income retrieval action

michael-hmrc: we already have an auth action called

IncomeSourceRetrievalAction which could be updated to redirect to the new error page if we cannot find any income sources

michael-hmrc: Image

michael-hmrc:
```
    incomeSourceDetailsService.getIncomeSourceDetails().map {
      case response: IncomeSourceDetailsModel =>
        Right(MtdItUser(req.mtditId, response.nino, req.mtdUserRole, req.authUserDetails, req.clientDetails, response))
      case error: IncomeSourceDetailsError =>
        Left(internalServerErrorFor(request))
    }
```

michael-hmrc: wonder what you think? thinking we update the left? and potentially anywhere else?

michael-hmrc: nvm ignore this

michael-hmrc (approved):

PR #3378, opened 2026-04-23
MISUV-11231 - Content for empty obligation Date submitted fields - Tax Year Summary Submissions tab

TomRafferty:


app/views/partials/taxYearSummary/SubmissionsContent.scala.html
```diff
@@ -49,6 +49,7 @@
 }
 @tableRows(input: Seq[ObligationWithIncomeType]) = @{
     input.map(nextUpdateWithIncomeType => {
+        val dateContent = nextUpdateWithIncomeType.obligation.dateReceived.map(_.toLongDateShort).fold(HtmlContent(""))(value => HtmlContent(value))
```
Mbrookesy: Since this is a fold you can just turn the empty HtmlContent into the not received message so you don't have to do another if statement

example:

content = nextUpdateWithIncomeType.obligation.dateReceived.map(_.toLongDateShort).fold(HtmlContent(messages("submissionsTab.notReceived")))(value => HtmlContent(value))

Mbrookesy (changes requested):

app/views/partials/taxYearSummary/SubmissionsContent.scala.html
TomRafferty: that's cleaned that up nicely, thank you 😄

Mbrookesy (approved):

PR #3383, opened 2026-04-24
[MISUV-11216] removing the search again link from international manual …

Mbrookesy:
…address confirmation page


andriikondiuk (approved):
LGTM

PR #3388, opened 2026-04-28
MISUV-11219 Adjust leading spaces handling for overseas journey

jmountain-hmrc:


app/utils/Converters.scala
```diff
@@ -19,7 +19,7 @@ package utils
 package object Converters {
 
   implicit class OptionExtension(in: Option[String]) {
-    def trim(): Option[String] = {
+    def flattenEmptyString(): Option[String] = {
```
andriikondiuk, line 22: It's good. But if you would like to get rid of **implicit conversion**. Here is an example by using Scala's 3 **extension methods** which are easy to read and native to the Scala 3:
```
extension (in: Option[String])
    def flattenEmptyString: Option[String] =
      in match {
        case Some("") => None
        case _ => in
      }
```

andriikondiuk (approved):
LGTM

PR #3390, opened 2026-04-29
[MISUV-11243] First fix for flakey pipeline tests

Mbrookesy:


it/test/helpers/servicemocks/AuditStub.scala
```diff
@@ -17,10 +17,19 @@
...
+import org.scalatest.time.{Millis, Seconds, Span}
 import play.api.http.Status.NO_CONTENT
 import play.api.libs.json.{JsValue, Json}
 
-object AuditStub extends WiremockMethods {
+object AuditStub extends WiremockMethods with Eventually {
+
+  override implicit val patienceConfig =
+    PatienceConfig(
+      timeout  = Span(5, Seconds),
```
andriikondiuk, line 29: Just a curiosity why 5 Second for `timeout`? As the default values for a timeout: `scaled(150 milliseconds)`.

Mbrookesy: More of a precautionary thing, trying to see what works and if needed I can reduce.

andriikondiuk: Sounds good, thanks

it/test/helpers/servicemocks/AuditStub.scala
```diff
@@ -17,10 +17,19 @@
...
 import audit.models.ExtendedAuditModel
+import org.scalatest.concurrent.Eventually
+import org.scalatest.time.{Millis, Seconds, Span}
 import play.api.http.Status.NO_CONTENT
 import play.api.libs.json.{JsValue, Json}
 
-object AuditStub extends WiremockMethods {
+object AuditStub extends WiremockMethods with Eventually {
+
+  override implicit val patienceConfig =
```
andriikondiuk, line 27: And please add a Type Annotation here. As for Scala 3 [The type of implicit definitions ( val or def ) needs to be given explicitly in Scala 3. They cannot be inferred anymore.](url)

Mbrookesy: Your link leads to a 404

andriikondiuk: Ahh sorry here: https://docs.scala-lang.org/scala3/guides/migration/incompat-contextual-abstractions.html#type-of-implicit-definition

andriikondiuk: We can ignore it as it's only a [ScalaFix](https://scalacenter.github.io/scalafix/docs/rules/ExplicitResultTypes.html) issue

andriikondiuk (approved):
LGTM.

PR #3396, opened 2026-04-30
MISUV-11145: Add conditional logic for help link and hiding help link in NoIncomeSources page

IoannisLazaridis-oc:
The help link is not required on this page, as the Contact us button leads to the same destination.

To manually test this page  you need to log in using the test user : NA000001A - Newly registered user with no income sources

image


paulanderson1234 (dismissed):
Approved subject to the observations we discussed around quotation marks used on the page - does not affect functionality

PR #3402, opened 2026-05-01
MISUV-11273: Updated the ItsaHomePage audit event to match the approved TXM audit change for CY+1

IoannisLazaridis-oc:
Note: The approved Confluence text has a typo saying `userIsCYPlusOneUser`, but the JSON and Jira ticket say `userIsCYPlusOne` which is what I used in this PR.

Confluence page: https://confluence.tools.tax.service.gov.uk/spaces/CIP/pages/245531673/MTD+ITSA+-+View+Change+-+October+2021+plus#MTDITSAView&ChangeOctober2021plus-ItsaHomePage

image

### DIFF
--- a/app/audit/models/HomeAudit.scala
+++ b/app/audit/models/HomeAudit.scala
@@ -42,7 +42,8 @@ case class HomeAudit(mtdItUser: MtdItUser[_],
 
   override val detail: JsValue = Json.obj(
     "mtditid" -> mtdItUser.mtditid,
-    "nino" -> mtdItUser.nino
+    "nino" -> mtdItUser.nino,
+    "userIsCYPlusOne" -> mtdItUser.incomeSources.yearOfMigration.isDefined
   ) ++ userType(mtdItUser.userType, mtdItUser.isSupportingAgent) ++ paymentsInformation ++ updatesInformation ++
     Json.obj("saUtr"-> mtdItUser.saUtr) ++
     Json.obj("credId"-> mtdItUser.credId) ++
--- a/app/audit/models/HomeAudit.scala
+++ b/app/audit/models/HomeAudit.scala
@@ -16,16 +16,17 @@
 
 package audit.models
 
-import audit.Utilities._
+import audit.Utilities.*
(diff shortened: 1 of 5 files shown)


app/audit/models/HomeAudit.scala
```diff
@@ -42,7 +42,8 @@ case class HomeAudit(mtdItUser: MtdItUser[_],
 
   override val detail: JsValue = Json.obj(
     "mtditid" -> mtdItUser.mtditid,
-    "nino" -> mtdItUser.nino
+    "nino" -> mtdItUser.nino,
+    "userIsCYPlusOne" -> mtdItUser.incomeSources.yearOfMigration.isDefined
```
Mbrookesy: We shouldn't be relying on the year of migration here as it's an unreliable field.

It's better to check if a user is a CY+1 user by checking if they have an ITSA status for the current tax year

Mbrookesy (changes requested):

app/audit/models/HomeAudit.scala
IoannisLazaridis-oc: Makes sense, now using the curent itsa status from controller to derive `userIsCYPlusOne`

app/controllers/HomeController.scala
```diff
@@ -207,7 +208,17 @@ class HomeController @Inject()(val homeView: views.html.HomeView,
             if (mandation) SessionKeys.mandationStatus -> "on"
             else SessionKeys.mandationStatus -> "off"
 
-          auditingService.extendedAudit(HomeAudit(user, paymentsDueMerged, overDuePaymentsCount, nextUpdatesTileViewModel))
+          val userIsCYPlusOne = currentITSAStatus != ITSAStatus.NoStatus
```
Mbrookesy: Shouldn't this be inverted? 

A CY+1 user is someone with a CY status of No status or nothing

Mbrookesy (changes requested):

Mbrookesy (approved):

PR #3411, opened 2026-05-07
[MISUV-11251] Adding home audit to your tasks page

Mbrookesy:


app/controllers/newHomePage/HandleYourTasksController.scala
```diff
@@ -74,34 +80,70 @@ class HandleYourTasksController @Inject()(val authActions: AuthActions,
 
   def handleShowRequest(origin: Option[String] = None)
                        (implicit user: MtdItUser[_], hc: HeaderCarrier): Future[Result] = {
-    handleYourTasks(origin, user.isAgent)
+    nextUpdatesService.getDueDates().flatMap {
```
IoannisLazaridis-oc: It looks like we are fetching  `nextUpdatesService.getDueDates` mroe than once when loading the tasks page.

app/controllers/newHomePage/HandleYourTasksController.scala
```diff
@@ -74,34 +80,70 @@ class HandleYourTasksController @Inject()(val authActions: AuthActions,
...
-      updatesAndDeadlinesViewModel <- getNextUpdates()
+      
+      outstandingChargesModel <- getOutstandingChargesModel(unpaidCharges)
+      paymentsDue = getDueDates(unpaidCharges, isEnabled(FilterCodedOutPoas), isEnabled(PenaltiesAndAppeals))
+      outstandingChargeDueDates = getRelevantDates(outstandingChargesModel)
+      overDuePaymentsCount = calculateOverduePaymentsCount(paymentsDue, outstandingChargesModel)
+      paymentsDueMerged = mergePaymentsDue(paymentsDue, outstandingChargeDueDates)
+
+      obligationsResponseModel = nextUpdatesService.getOpenObligations()
+      dueDates <- nextUpdatesService.getDueDates(Some(obligationsResponseModel)).flatMap {
```
IoannisLazaridis-oc, line 105: again fetching nextUpdatesService.getDueDates. Wouldn't be better to reuse the first response?

app/controllers/newHomePage/HandleYourTasksController.scala
```diff
@@ -74,34 +80,70 @@ class HandleYourTasksController @Inject()(val authActions: AuthActions,
...
       val mandationStatus =
         if (mandation) SessionKeys.mandationStatus -> "on"
         else SessionKeys.mandationStatus -> "off"
 
-      val yourTaskCardViewModel = handleYourTasksService.getYourTasksCards(updatesAndDeadlinesViewModel,
-        isAgent, chargeItemList, credits, creditsRefundsRepayEnabled, currentItsaStatus, penaltiesAndAppealsEnabled)
+      val yourTaskCardViewModel = handleYourTasksService.getYourTasksCards(updatesAndDeadlinesViewModel, isAgent, chargeItemList, credits, creditsRefundsRepayEnabled, currentItsaStatus, penaltiesAndAppealsEnabled)
+      
+      val overdueUpdatesCount = nextUpdatesDueDates.count(_.isBefore(dateService.getCurrentDate))
+      val nextUpdateDueDate = nextUpdatesDueDates.sortWith(_ isBefore _).headOption
```
IoannisLazaridis-oc: It seems now the audit values are calculated separately from the `updatesAndDeadlinesViewModel`. I think we could reuse the same values

app/controllers/newHomePage/HandleYourTasksController.scala
Mbrookesy: Good spot, must've left it in during the mess of it all, will remove

app/controllers/newHomePage/HandleYourTasksController.scala
Mbrookesy: We never used the values from updatesAndDeadlinesViewModel, are you talking about the next updates tile model? I'd rather not create and import the tile model just for that purpose

IoannisLazaridis-oc: wouldn't be more accurate to capture the same data what we display in the view?

Mbrookesy: We don't use the tile in the new home page

app/controllers/newHomePage/HandleYourTasksController.scala
```diff
@@ -100,8 +120,22 @@ class HandleYourTasksController @Inject()(val authActions: AuthActions,
         if (mandation) SessionKeys.mandationStatus -> "on"
         else SessionKeys.mandationStatus -> "off"
 
-      val yourTaskCardViewModel = handleYourTasksService.getYourTasksCards(updatesAndDeadlinesViewModel,
-        isAgent, chargeItemList, credits, creditsRefundsRepayEnabled, currentItsaStatus, penaltiesAndAppealsEnabled)
+      val yourTaskCardViewModel = handleYourTasksService.getYourTasksCards(updatesAndDeadlinesViewModel, isAgent, chargeItemList, credits, creditsRefundsRepayEnabled, currentItsaStatus, penaltiesAndAppealsEnabled)
+      
+      val overdueUpdatesCount = dueDates.count(_.isBefore(dateService.getCurrentDate))
+      val nextUpdateDueDate = dueDates.sortWith(_ isBefore _).headOption
```
IoannisLazaridis-oc, line 126: This is more of a question - since this sorts all due dates and takes `headOption`, would this return an overdue date if both overdue and upcoming obligations exist?

Mbrookesy: It would return the earliest due date so It's assumed it would also return an overdue one if that's the earliest open one

This retains the current logic so it's no different from what wouldve been displayed on the next updates tile on the og homepage

IoannisLazaridis-oc (approved):

PR #3422, opened 2026-05-12
MISUV-10441 - Create new page with content logic

nemanjatomanovic:


app/controllers/newHomePage/ProofOfYourIncomeController.scala
```diff
@@ -0,0 +1,116 @@
...
+        handleRequest(controllers.routes.HomeController.showAgent().url, true)
+      } else {
+        Future.successful(Redirect(controllers.newHomePage.routes.HandleYourTasksController.showAgent()))
+      }
+  }
+
+  def handleRequest(backUrl: String, isAgent: Boolean)
+                   (implicit user: MtdItUser[_], hc: HeaderCarrier, ec: ExecutionContext): Future[Result] = {
+
+    if (user.saUtr.isDefined) {
```
jameshforster: This check should be done later, a user without an SAUTR still needs to be able to access this page

app/controllers/newHomePage/ProofOfYourIncomeController.scala
```diff
@@ -0,0 +1,116 @@
...
+  {
+    Future.successful(itvcErrorHandler.showInternalServerError())
+  }
+}
+
+
+  private def assembleProofOfYourIncomeModel (taxYears: List[String], saUtr: String)
+                   (implicit user: MtdItUser[_], hc: HeaderCarrier, ec: ExecutionContext): Seq[Future[ProofOfYourIncomeCardViewModel]] = {
+    taxYears.map { taxYear =>
+      val taxYearFormatted = taxYear.split("-")(0).toInt
```
jameshforster: Tax year should be called by the end year not the start year which this seems to be making

app/controllers/newHomePage/ProofOfYourIncomeController.scala
```diff
@@ -0,0 +1,116 @@
...
+      calcService.getLiabilityCalculationDetail(user.mtditid, user.nino, taxYearFormatted).flatMap {
+        case calcResponse: LiabilityCalculationResponse =>
+          Future.successful(ProofOfYourIncomeCardViewModel(taxYearFormatted, calcResponse.metadata.calculationType,
+            dateService.getCurrentTaxYearStart.getYear, false, saUtr))
+        case calcError: LiabilityCalculationError if calcError.status == NO_CONTENT =>
+          Logger("application").info(s"No data for tax year: ${taxYearFormatted}")
+          Future.failed(new Exception("NO_CONTENT"))
```
jameshforster: Just to check, does this cause the page to fail to load? No data would still be valid as a scenario and we would just not display any element.

jameshforster (changes requested):

app/models/newHomePage/ProofOfYourIncomeCardViewModel.scala
```diff
@@ -0,0 +1,60 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.newHomePage
+
+import play.api.i18n.Messages
+import services.DateServiceInterface
+
+case class ProofOfYourIncomeCardViewModel(taxYearStart: Int,
```
jameshforster, line 22: This model requires unit testing, unit tests in the controller do not cover each scenario in enough detail

jameshforster (changes requested):

jameshforster (approved):

PR #3425, opened 2026-05-12
MISUV-11322 Add  -Werror which is successor of the -Xfatal-warnings and -unchecked compiler options

andriikondiuk:


andriikondiuk (commented):
This PR would mark compile time **_Warnings_** into _**compilation errors**_ in all stages compile / test / IT. It will force a developer to address an issue straight away.

app/obligations/services/reportingObligations/ReportingObligationsAuditService.scala
```diff
@@ -99,7 +98,7 @@ class ReportingObligationsAuditService @Inject()(
     }
   }
 
-  private def tableContentToItsaStatus(content: Option[String])(implicit messages: Messages, user: MtdItUser[_]): ITSAStatus = {
```
IoannisLazaridis-oc, line 102: Not fully udnerstand why we had to remove the MtdItUser[_]. Can we verify that nothing is relying on the on the implicit user details?

andriikondiuk: I've removed it as the implicit parameter `user: MtdItUser[_]` is not used. All Unit/IT and AT's passed.

IoannisLazaridis-oc (approved):

PR #3431, opened 2026-05-13
Misuv 11293: Widening the tab to correct heading 

taradhesi14:
There are additional changes to correct errors that were appearing in html validator for main too that went unnoticed.

### DIFF
--- a/app/views/components/bulletPointList.scala.html
+++ b/app/views/components/bulletPointList.scala.html
@@ -18,8 +18,8 @@
 
 @(itemMsgKeys: Seq[String], id: Option[String]= None, bulletId: Option[String] = None)(implicit messages: Messages)
 
-<ul class="govuk-list govuk-list--bullet" id="@id">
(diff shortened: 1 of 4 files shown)


app/views/components/bulletPointList.scala.html
```diff
@@ -18,8 +18,8 @@
 
 @(itemMsgKeys: Seq[String], id: Option[String]= None, bulletId: Option[String] = None)(implicit messages: Messages)
 
-<ul class="govuk-list govuk-list--bullet" id="@id">
+<ul class="govuk-list govuk-list--bullet" @if(id.isDefined) {id="@id"}>
     @itemMsgKeys.map { msgKey =>
-       <li id="@bulletId">@messages(msgKey)</li>
+        <li @if(bulletId.isDefined) {id="@bulletId"}>@messages(msgKey)</li>
```
TomRafferty, line 23: was the purpose of adding ID here for testing?

TomRafferty (approved):

PR #3450, opened 2026-05-19
[MISUV-11309] Custom login updates for the test data rework

Mbrookesy:


app/testOnly/models/IncomeSourcesUser.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import play.api.libs.json.{Json, OFormat}
+
+case class IncomeSourcesUser(activeSoleTrader: Boolean,
+                             activeUkProperty: Boolean,
+                             activeForeignProperty: Boolean,
+                             ceasedBusiness: Boolean)
+
+object IncomeSourcesUser {
+  implicit val format: OFormat[IncomeSourcesUser] = Json.format[IncomeSourcesUser]
```
michael-hmrc: small one and not super familiar with scala 3 but guess do we change implicit val to given

Mbrookesy: Sure can do

app/testOnly/connectors/DynamicStubConnector.scala
```diff
@@ -127,6 +126,29 @@ class DynamicStubConnector @Inject()(val appConfig: TestOnlyAppConfig,
...
+      .withBody(requestJson)
+      .execute[HttpResponse] map { response =>
+      response.status match {
+        case OK => (): Unit
+        case _ =>
+          Logger("application").error(s" Overwrite unsuccessful. ~ Response status: ${response.status} ~. < Response body: ${response.body} >")
+          throw new Exception(s"Overwrite unsuccessful. ~ Response status: ${response.status} ~. < Response body: ${response.body} >")
+      }
+    }
+    Future(())
```
michael-hmrc: think we would want to throw return the Unit from the http response and not  
return 
Future(())

michael-hmrc: Know it's already a established thing in the stub and custom login type code but typically dislike throwing Units everywhere if we can help it but not a blocker.

TomRafferty (approved):

TomRafferty (approved):

PR #3466, opened 2026-06-02
MISUV-11415 CreditSummary refactor to use GovukTable component

jmountain-hmrc:


TomRafferty (changes requested):
just a small one from me:
we have some unused code (since 2022, lines 57-61) which we could probably chop out as part of routine maintenance 😃 
Other than that, everything looks good to me

TomRafferty (approved):
Thanks for the clean up, all looks good 😄

PR #3473, opened 2026-06-04
MISUV-11438 - What you owe- Charges with interest accruing/overdue/Crystallised interest to pay

nemanjatomanovic:


app/models/financialDetails/WhatYouOweViewModel.scala
```diff
@@ -23,6 +23,7 @@ import financials.controllers.routes as financialsRoutes
 
 case class WhatYouOweViewModel(currentDate: LocalDate,
                                hasOverdueOrAccruingInterestCharges: Boolean,
+                               hasCrystalisedInterest: Boolean,
```
jameshforster: Minor typo, should be spelled "Crystallised"

app/views/WhatYouOweView.scala.html
```diff
@@ -72,14 +72,14 @@
 ) {
     @h1(msg = getHeading, id = Some("page-heading"))
 
-    @if(viewModel.hasOverdueOrAccruingInterestCharges) {
-
+    @if(viewModel.hasOverdueOrAccruingInterestCharges || viewModel.hasCrystalisedInterest) {
```
jameshforster: This section should only show if a charge is accruing interest, if there are only charges with crystallised interest this content should not display

conf/messages
```diff
@@ -458,7 +458,7 @@ tax-year-summary.deductions                                        = Allowances
 tax-year-summary.taxable-income                                    = Total income on which tax is due
 tax-year-summary.tax-due                                           = Self Assessment tax amount
 tax-year-summary.payments.payment-type                             = Payment type
-tax-year-summary.payments.charge-type                              = Charge type
+tax-year-summary.payments.charge-type                              = Tax details
```
jameshforster: This seems to edit a message for a different page that has not been altered by this story. The correct edit has also been made for a different message key. Please undo this change unless someone has specifically requested it from design.

app/views/partials/whatYouOwe/SaNoteParagraph.scala.html
```diff
@@ -46,24 +46,28 @@
 @saNoteParagraphContent(utr: String, ctaTaxYear: Option[TaxYear]) = {
 
     @h2(msg = getMessage("sa-note-1-heading"))
-    @p(id = Some("sa-note-1-migrated-1"))(Html(getMessage("sa-note-1-body-1") +
-            link(link = selfServeTimeToPayStartUrl,
-                outerMessage = ".",
+    @p(id = Some("sa-note-1-migrated-1")) {
```
jameshforster: This paragraph should only appear if balance details shows an overdue charge, please add a conditional check for the paragraph which checks if balanceDetails.overDueAmount is greater than zero

app/views/WhatYouOweView.scala.html
```diff
@@ -93,6 +93,9 @@
                 HeadCell(
                     content = Text(getMessage("tableHead.tax-year"))
                 ),
+                HeadCell(
```
jameshforster: This column should only exist if any charges are accruing interest, otherwise it should not appear.

app/views/WhatYouOweView.scala.html
```diff
@@ -171,6 +184,10 @@
                             TableRow(
                                 attributes = Map("id" -> s"accrued-interest-tax-year-$index"),
                                 content = HtmlContent(taxYearSummaryLink(charge.taxYear.endYear, index, viewModel.taxYearSummaryUrl))
+                            ),TableRow(
```
jameshforster: This cell needs to be left aligned when the not applicable message appears in the same column

jameshforster (changes requested):
Please address the comments below and check out how the page looks on user PA000002A as it does not line up with the visuals on the ticket

jameshforster (approved):

PR #3474, opened 2026-06-04
[MISUV-11423] Updating the link for exempt users to classic SA

Mbrookesy:


app/obligations/views/helpers/injected/NextUpdatesHelperR17.scala.html
```diff
@@ -142,19 +142,26 @@
...
     @p(id = Some("current-year-compatible-software-desc")) {
         @HtmlFormat.fill(Seq(
-            Html(if(Set(ITSAStatus.Exempt, ITSAStatus.DigitallyExempt).contains(currentTaxYearStatus)) {
+            Html(if(ITSAStatus.isExemptionStatus(currentTaxYearStatus)) {
                 getMessage("annual.text2.exempt")
             } else {
                 getMessage("annual.text2")
             }),
-            link(link = appConfig.compatibleSoftwareLink, target = Some("_blank"), messageKey = getMessage("annual.link"), id = Some("annual-compatible-software-link"), outerMessage = ".")
+            link(link = fileLink(currentTaxYearStatus), target = Some("_blank"), messageKey = getMessage("annual.link"), id = Some("annual-compatible-software-link"), outerMessage = ".")
```
IoannisLazaridis-oc, line 164: It seems that the same message key is used for every annual user (exempt or not). For non-exempt users we render: 

- text : `nextUpdates.r17.tab.annual.link                                    = file your Self Assessment tax return`

- url: `compatible-software-link = "https://www.gov.uk/guidance/choose-the-right-software-for-making-tax-digital-for-income-tax"`

These two contradict each other. The link text says `file your Self Assessment tax return` but the URL goes to the MTD software page. I think we should double check if that's the design intent

Mbrookesy: I did check already and bethrand said that it's fine

IoannisLazaridis-oc (approved):

PR #3485, opened 2026-06-09
MISUV-11465 -  Refactored IncomeBreakdown table and removed redundant table component

TomRafferty:


app/returns/views/IncomeBreakdownView.scala.html
```diff
@@ -18,17 +18,48 @@
...
 )
 
 @(incomeBreakdownViewModel: IncomeBreakdownViewModel, taxYear: Int, backUrl: String, isAgent: Boolean = false, serviceNavigationPartial: Option[ServiceNavigation] = None)(implicit request: Request[_], messages: Messages)
 
+@tableRow(messageKey: String, dataItem: Option[BigDecimal], isNegative: Boolean = false): Option[Seq[TableRow]] = @{
+    dataItem.map { data =>
+        Seq(
+            TableRow(
+                attributes = Map("id" -> s"$messageKey-row"),
+                content = HtmlContent(messages(s"income_breakdown.table.$messageKey")),
```
IoannisLazaridis-oc: Not a blocker, but if any of these message values returns characters like `&` or any other special character, using HtmlContent could cause rendering issues since it treats the string as raw HTML. `Text(messages())` might be a better one

IoannisLazaridis-oc (approved):

PR #3487, opened 2026-06-10
[MISUV-11490] Adding new idempotency key FS

Mbrookesy:


app/common/models/admin/FeatureSwitchName.scala
```diff
@@ -236,3 +239,8 @@ case object MortgageEvidence extends FeatureSwitchName {
   override val toString: String = "mortgage-evidence"
 }
 
+case object IdempotencyKeyForCreateIncomeSource extends FeatureSwitchName {
+  override val name: String = "idempotency-key-for-create-income-source"
+  override val toString: String = "idempotency-key-for-create-income-source"
```
TomRafferty: Think this shouldn't be hyphenated? If we're matching the pattern then only name should be I believe

TomRafferty (changes requested):

TomRafferty (approved):

PR #3501, opened 2026-06-12
[MISUV-11487] Adding idempotency key and add income source field to create an income source

Mbrookesy:


app/businessDetails/utils/JourneyCheckerManageBusinesses.scala
```diff
@@ -110,16 +112,28 @@ trait JourneyCheckerManageBusinesses extends IncomeSourcesUtils {
...
+                hc.sessionId.get.value,
+                incomeSources.toString,
+                addIncomeSourceData = idempotencyKey.map(key => AddIncomeSourceData(idempotencyKey = Some(key)))
+              )
+
+              idempotencyKey.foreach { _ =>
+                sessionService.setMongoData(data)
+              }
+
+              codeBlock(data)
```
IoannisLazaridis-oc: This might create some side effects. We never check when or what is returned from `sessionService.setMongoData(data)` and we immediately run `codeBlock(data)`. That means the response could be sent before Mongo has finished writing the key . This could break idempotency in cases where the Mongo write is slow and the next page reads the session faster than the write completes

Maybe we could try something like: `sessionService.setMongoData(data).flatMap(_ = > codeBlock(data)`

What do you think?

app/businessDetails/services/CreateBusinessDetailsService.scala
```diff
@@ -75,6 +75,8 @@ class CreateBusinessDetailsService @Inject()(val createIncomeSourceConnector: Cr
     Try {
       CreateBusinessIncomeSourceRequest(
         mtdbsa = user.mtditid,
+        idempotencyKey = viewModel.idempotencyKey,
+        addIncomeSource = if(viewModel.idempotencyKey.isDefined) Some(true) else None,
```
IoannisLazaridis-oc, line 79: also related to my previosu comment, if for any reason the idempotency key is missing then the `addIncomeSource` will return None without any compain. Not sure if should populate this based on the call result  but also  not sure if its safer to leave as-is

app/businessDetails/utils/JourneyCheckerManageBusinesses.scala
Mbrookesy: Yeah that makes sense, sure if that works though it may return a unit

app/businessDetails/services/CreateBusinessDetailsService.scala
Mbrookesy: The idea would be that if it contains nothing then it wouldn't need to be populated as they have to be together.

I imagine the other fix would solve that issue

IoannisLazaridis-oc (changes requested):

it/test/businessDetails/controllers/manageBusinesses/add/AddIncomeSourceStartDateControllerISpec.scala
```diff
@@ -114,7 +115,7 @@ class AddIncomeSourceStartDateControllerISpec extends ControllerISpecHelper {
         s"GET $path" when {
```
Mbrookesy, line 115: It's already there on line 71

IoannisLazaridis-oc (approved):

PR #3538, opened 2026-06-30
MIPR 2784

aashiqjenner-cloud:
Objective: Add the BusinessDetails feature switch so when enabled it redirects to the BusinessDetails service

Acceptance Criteria

When enabled the BusinessDetails to BusinessDetails service happen
When disabled stays in V&C frontend.

Steps to implement:{}

 add a new feature switch to list of feature switches in view-change-frontend and the stub
On the home page where links to the business details service are update so it is feature switched to go to business details when enabled

Useful documentation

Copy

TBC

Prototype

TBC

Testing strategy
Integration
Unit
Acceptance criteria

Known dependencies

TBC

Outstanding unknowns/work

TBC

DoR/DoD

### DIFF
--- a/app/hub/controllers/HomeController.scala
+++ b/app/hub/controllers/HomeController.scala
@@ -16,6 +16,7 @@
 
 package hub.controllers
 
+import common.models.admin.BusinessDetailsFrontend
 import businessDetails.auth.AuthActionsWithTriggeredMigrationCheck
 import common.auth.MtdItUser
 import common.config.*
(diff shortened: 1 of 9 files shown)


kSiadat (approved):

kSiadat (approved):

conf/application.conf
```diff
@@ -198,6 +198,11 @@ microservice {
         port = 9089
       }
 
+      income-tax-business-details-frontend {
+        host = localhost
+        port = 9088
+      }
+
```
Chrissystk, line 205: This should be done similar to how it is done for address-lookup etc

app/hub/views/partials/newHome/overview/IncomeSection.scala.html
```diff
@@ -16,18 +16,18 @@
 
 @import hub.views.html.partials.newHome.overview.OverviewCard
 @import businessDetails.controllers.manageBusinesses.routes.{ManageYourBusinessesController}
-
-
 @this(overviewCard: OverviewCard)
 
 
-@(isAgent: Boolean)(implicit messages: Messages)
+@(isAgent: Boolean, manageBusinessesUrl: Option[String] = None)(implicit messages: Messages)
```
Chrissystk, line 22: would have been better to pass the appConfig implicitly here then you don't need to pass an Option[String] around and instead call the config to get the url.

app/hub/views/helpers/injected/home/IncomeSourcesTile.scala.html
```diff
@@ -29,10 +29,12 @@
 @(yourBusinessesTileViewModel: YourBusinessesTileViewModel, isAgent: Boolean)(implicit messages: Messages)
```
Chrissystk, line 30: would have been better to pass the appConfig implicitly here then you don't need to pass an Option[String] around and instead call the config to get the url.

app/hub/controllers/HomeController.scala
```diff
@@ -104,6 +105,26 @@ class HomeController @Inject()(val homeView: hub.views.html.HomeView,
...
+      s"${appConfig.incomeTaxBusinessDetailsFrontendBaseUrl}$path"
+    } else {
+      if (isAgent) {
+        businessDetails.controllers.manageBusinesses.routes.ManageYourBusinessesController.showAgent().url
+      } else {
+        businessDetails.controllers.manageBusinesses.routes.ManageYourBusinessesController.show().url
+      }
+    }
+  }
+
```
Chrissystk, line 127: This would be better in AppConfig then it can be called in the view files rather than passing Option[String] around.

PR #3562, opened 2026-07-13
MISUV-11716 Add penalty content to interstitial payment page

andriikondiuk:


app/financials/models/ChargeItem.scala
```diff
@@ -187,6 +187,8 @@ case class ChargeItem(
 
   val isPenalty: Boolean = List(LateSubmissionPenalty, FirstLatePaymentPenalty, SecondLatePaymentPenalty).contains(this.transactionType)
 
+  val isLPPPenalty: Boolean = List(FirstLatePaymentPenalty, SecondLatePaymentPenalty).contains(this.transactionType)
```
tbg2003: Minor thing but LPP stands for "Late Payment Penalty" so LPPPenalty is a little confusing, perhaps rename val to just "isLPP"

```suggestion
  val isLPP: Boolean = List(FirstLatePaymentPenalty, SecondLatePaymentPenalty).contains(this.transactionType)
```

tbg2003 (changes requested):

app/financials/models/ChargeItem.scala
andriikondiuk: Nice one. Done, thanks

tbg2003 (approved):

PR #3585, opened 2026-07-23
MISUV-11776 Add new chargeClassification field to FE and add method to check for revenue amendment

andriikondiuk:


andriikondiuk (commented):
Ready for review. Important, as I understand from the requirements the **ChargeClassification** type maybe extended with new values.

nemanjatomanovic (dismissed):

dnyuar (approved):

nemanjatomanovic (approved):

PR #3587, opened 2026-07-27
adding returns frontend feature switch

adambanner1904:


Chrissystk (changes requested):
Although you have covered the ToDos you haven't updated the links on home pages to redirect to returns if the feature is enabled. If you a find of returns. in the hub folder you can see where you need to change.

Chrissystk (approved):

PR #3593, opened 2026-07-30
MIPR-2911: changed prod features button to read from stub

kSiadat:


app/testOnly/controllers/FeatureSwitchController.scala
```diff
@@ -122,15 +118,24 @@ class FeatureSwitchController @Inject()(featureSwitchView: FeatureSwitchView,
...
         for {
-          (fs, enableState) <- (getEnabledFeatureSwitches ++ getDisabledFeatureSwitches)
-        } yield featureSwitchService.set(fs, enableState)
-      )
-    } yield Redirect(testOnly.controllers.routes.FeatureSwitchController.show())
-
+          _ <- for {
+            _ <- featureSwitchService.resetToProd()
+            features <- featureSwitchService.getAll()
+          } yield features.map(fs => featureSwitchService.set(fs.name, fs.isEnabled))
```
Chrissystk: You shouldn't need to do this as the getAll will be called when it redirects to show() page again. You just need to call the resetToProd then redirect to the show() page

Chrissystk (changes requested):

Chrissystk (approved):

PR #3632, opened 2026-08-24
MISUV-11915 - CL289 | Tax refund or set off

nemanjatomanovic:


app/returns/models/liabilitycalculation/viewmodels/TaxDueSummaryViewModel.scala
```diff
@@ -74,6 +77,13 @@ case class TaxDueSummaryViewModel(
     }
   }
 
+  def getSATaxAmount: Option[BigDecimal] = {
```
dnyuar, line 80: Could you add a test for this function to show it's working correctly

dnyuar (changes requested):

dnyuar (approved):

dnyuar (approved):

